PR #135, opened 2018-11-14
Update server.xml.template

John-A-Davies:
Add 2 lines to disable update triggers and dropins.  This greatly reduces CPU and EXCP consumption.


Joe-Winchester:
Hi John.  Tks for doing this - you didn't create the DCO.  I also created the same fix overlapping with you (which is silly we didn't communicate sooner)

stevenhorsman:
Dup of #136

PR #147, opened 2018-11-22
Add zowe-pre-install

John-A-Davies:
Create new script /install/zowe-preinstall.sh that will perform pre-req checking and report problems.


Joe-Winchester:
This has been redone under a separate pull request so closing this pull request as it doesn't have DCO set

PR #150, opened 2018-11-22
Verify post install

John-A-Davies:
Add zowe-verify-post-install.sh and copy it to Zowe runtime directory as /scripts/zowe-verify.sh


John-A-Davies:
Superseded by #152

PR #155, opened 2018-11-27
Cleanup verify scripts

John-A-Davies:
Remove verbose messages to the user - just say OK if that check passes.
Relocate scripts to correct directories.
Remove false negative in ports-in-use section.
Remove check for available space.  
Remove file check for 'sample' directory.


John-A-Davies:
Abandoned due to missing DCO.

PR #157, opened 2018-11-27
pre and post changes from John

Joe-Winchester:


Joe-Winchester:
This should not be merged as it has been replaced by #159

PR #164, opened 2018-11-30
Generate certificates for zLUX using APIML CM

plavjanik:
This PR moves the APIML configuration to the beginning so it does generate certificates before other services are configured.

The zLUX configuration `zluxservice.json` is updated to use APIML-generated certificates.

Used to depend on https://github.com/zowe/zlux-example-server/pull/27 but the script has been updated to not cause problems if `certificateAuthorities` line is missing.

Resolves zowe/api-layer#92

It adds a `pytest` tests that help with unit testing of the script on UNIX-like system without doing it on z/OS from PAX.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,8 @@
 .history/
+tmp/
+.vscode/
+.pytest_cache/
+*.pyc
 
 # logs
 *.log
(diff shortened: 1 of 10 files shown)


plavjanik:
@stevenhorsman Please review

Issue #180, opened 2018-12-07
Can not pin or unpin apps to taskbar after start ZOWESVRG from sdsf


CindyZX:
The original problem is this one: https://github.com/zowe/zlux/issues/56
At first, I thought it was a problem of Zowe web desktop
But then I found that, it is much more tricky. 
Here are the steps to reproduce:
1.Download Zowe 0.9.4 official build from zowe.org
2.Install the build
3.After install, start zlux server by command zowe/0.9.4/zlux-example-server/bin/nodeServer.sh
In client side, login to Zowe web desktop, try to pin apps to taskbar, the app can be pinned to taskbar without problem
4.stop the node server
5.try to start ZOWESVRG in SDSF
ZOWESVRG can not be started, there is an error message in log: run-zowe.sh can not be found
6.from user directory, run command: chmod -R a+x zowe
7.start ZOWESVFG in SDSF again, ZOWESVRG can be started successfully
8.In client side, login to Zowe web desktop, and try to pin app to taskbar
Problem: the app can not be pinned or unpinned

CindyZX:
The workaround is:
go to zowe/0.9.4/zlux-example-server/deploy/instance, chmod -R 777 users

1000TurquoisePogs:
/zlux-example-server/deploy/instance is one of the https://github.com/zowe/zlux/wiki/Configuration-Dataservice folders.
Folders & files are created for individual users when a "user" scope is used. Pinning of apps is done in the user scope.
The server always needs to run as the same user, or at least in the same group, so that it can continue to use the files & folders it creates.
You shouldn't need 777, but 500 for the server user, or 550 should work.

jackjia-ibm:
@1000TurquoisePogs would you please check if this is a valid fix? https://github.com/zowe/zowe-install-packaging/pull/181

jackjia-ibm:
This is fixed by https://github.com/zowe/zowe-install-packaging/pull/181.

PR #181, opened 2018-12-07
issue #180 fix permission denied error

jackjia-ibm:


1000TurquoisePogs:
deploy/instance & deploy/site are intended to be writable. deploy/product also exists and is intended to be read-only.
The location of product, site, instance, group, and user folders are actually configuration-defined, but these are the default locations.

jackjia-ibm:
Thanks for the feedbacks. I agree and will keep those in mind. Once we have those configuration options available, we need to make sure this part is re-visited.

PR #187, opened 2018-12-13
Git issue 186

Joe-Winchester:


Joe-Winchester:
@nannanli and @John-A-Davies .  Agreed that troubleshoot is the wrong name for the script.  Perhaps just having it called "zowe-verify" would be better to go full circle ?

John-A-Davies:
So the script names need to be updated, and then I'll approve them.

PR #195, opened 2018-12-14
set tn3270 security type

### LINKED ISSUE #127: Allow specification of tn3270 connection type
The terminal has a capability to use TLS instead of plain telnet, and is controlled here: https://github.com/zowe/tn3270-ng2/blob/master/_defaultTN3270.json#L5 where "tls" is the other option.
This should be exposed in the installer so that users can choose their port, but also their security preference.

John-A-Davies:

fixes #127


John-A-Davies:
Yes we should.  Will fix now.

PR #214, opened 2018-12-20
Check free space

John-A-Davies:
Check that there will be enough disk space to install Zowe before the install shell script is run.  The YAML file defines this target directory, which may or may not exist.  The zowe-check-prereqs.sh script uses the specified target, and checks its parent directory if the nominated one does not yet exist.


John-A-Davies:
We also report which existing directory will be checked for space to the installing user.

PR #218, opened 2018-12-21
Configure stc2

John-A-Davies:
Add script `scripts/zowe-config-stc.sh` to configure the zowesvr STC to start zowe with the right userid and group.

And fix up `install/zowe-install.sh` to show the configured PROCLIB member name instead of the default.

And copy the log file to the runtime directory.


John-A-Davies:
Includes fixes from #216

PR #221, opened 2019-01-03
JES explorer node UI

jackjia-ibm:


jordanCain:
Requires tomcat jobs api to be configured as part of zowe-install-packaging before merging

Issue #222, opened 2019-01-03
zowe-install.sh unable to run deploy.sh on line 116


Joe-Winchester:
A number of users have reported problems with zowe-install.sh when it is unable to run deploy.sh on line 116.
This is because the line is
. deploy.sh
and changing it to be 
./deploy.sh
picks up the file correctly.

jackjia-ibm:
Fixed with PR https://github.com/zowe/zowe-install-packaging/pull/223

PR #223, opened 2019-01-03
Deploy.sh running with qualified slash

Joe-Winchester:


Joe-Winchester:
Fix for git issue #222

PR #240, opened 2019-01-11
Set zowe env vars without directly running .profile

John-A-Davies:


stevenhorsman:
$PROFILE should be removed, or better set to .zowe_profile - then some of the other changes aren’t required
Remove the section about touch ~/$PROFILE
If you created zowe_profile will it have execute permission?

John-A-Davies:
The 'source' command, i.e. the dot command, does not require its target to be executable.

John-A-Davies:
I removed the `touch` of PROFILE, so we don't create an unnecessary file.  PROFILE is set by `zowe-install-sh` anyway so I removed that too.

stevenhorsman:
Can you commit the zowe-install.sh change to this PR?
Also zowe-install-test:uninstall-zowe.sh reference to profile will need updating

stevenhorsman:
Convention also seems to be to use _ rather that - in hidden profile files

John-A-Davies:
I should have said  "PROFILE is _already_ set by the existing zowe-install-sh script."

Issue #217, opened 2018-12-21
Integrate new springboot based data-sets app into install 


stevenhorsman:
Similar to #206, but for data-sets
Depends on https://github.com/zowe/data-sets/issues/1
Depends on https://github.com/zowe/data-sets/issues/2

Pull down the assets from artifactory
Add a new port for the app into the yaml
share the api ml certificate with the app
Create a script to start the jar and call this from zowe-start?
What else?

jackjia-ibm:
Done by https://github.com/zowe/zowe-install-packaging/pull/242

PR #242, opened 2019-01-13
Issue #217: add data sets api into zowe build

jackjia-ibm:
The zowe build `libs-snapshot-local/com/project/zowe/0.9.6-ISSUE-217/zowe-0.9.6-20190112.090638-5.pax` has datasets api packaged and can be started.

### DIFF
--- a/scripts/zowe-explorer-api-configure.sh
+++ b/scripts/zowe-explorer-api-configure.sh
@@ -15,11 +15,9 @@
 # $ZOWE_JAVA_HOME
 # $ZOWE_ROOT_DIR
 # $ZOWE_EXPLORER_SERVER_JOBS_PORT
+# $ZOWE_EXPLORER_SERVER_DATASETS_PORT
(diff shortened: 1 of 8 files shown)


scripts/zowe-explorer-api-configure.sh
```diff
@@ -60,4 +58,41 @@ chmod 755 $ZOWE_ROOT_DIR/$EXPLORER_INSTALL_FOLDER/scripts
 # Configure explorer jobs api ended
 #############################################
 
+#############################################
```
stevenhorsman: Post 1.0 we need to clean up this file (and probably a bunch of others) to refactor and remove all the copy-pasting.

jackjia-ibm: New PR https://github.com/zowe/zowe-install-packaging/pull/248 created to optimize the install/configure scripts.

PR #241, opened 2019-01-11
Fix zlux paths upon server rename

1000TurquoisePogs:
Depends on zlux repo also being updated, so please don't merge if you approve... I'll coordinate merging.
Text replace of zlux-example-server to zlux-app-server, zlux-proxy-server to zlux-server-framework

Signed-off-by: 1000TurquoisePogs <sgrady@rocketsoftware.com>

PR #243, opened 2019-01-13
Includes APIML authentication plugin to Zowe zLUX

plavjanik:
This PR uses the new minor version of APIML `0.4.x` which includes the authentication service based on z/OSMF.

The authentication plugin is added to the zLUX during installation and it does the authentication using APIML.

The APIML token is stored in zLUX session and it is passed to the API catalog when it opened.

### DIFF
--- a/scripts/zowe-install-existing-plugin.sh
+++ b/scripts/zowe-install-existing-plugin.sh
@@ -1,0 +1,34 @@
+#!/bin/sh
+
+# Requires passed in:
+# ZOWE_ROOT_DIR - install root directory
+# PLUGIN_ID - id of the plugin - e.g. "org.zowe.zlux.auth.apiml"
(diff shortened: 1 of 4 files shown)


scripts/zowe-install-existing-plugin.sh
```diff
@@ -0,0 +1,34 @@
...
+if ! [ -d "$1" ]; then
+  echo "$1 not a directory, please provide the full path to the root installation directory of Zowe" >&2
+  exit 1
+fi
+if ! [ -d "$3" ]; then
+  echo "$3 not a directory, please provide the full path to the directory with the plugin" >&2
+  exit 1
+fi
+
+zluxserverdirectory='zlux-example-server'
```
stevenhorsman: This has changed to 'zlux-app-server' in https://github.com/zowe/zowe-install-packaging/pull/241

PR #248, opened 2019-01-14
Optimize explore api/ui install scripts

jackjia-ibm:


jackjia-ibm:
The installation script succeeded with test build https://wash.zowe.org:8443/job/zowe-install-test/job/master/211/console

Issue #258, opened 2019-01-15
Unable to run zowe-start.sh 


Joe-Winchester:
zowe-start.sh logs
You must run this script from the ZOWE_ROOT_DIR/scripts directory
You are running from .

jackjia-ibm:
PR https://github.com/zowe/zowe-install-packaging/pull/260

jackjia-ibm:
Verified with build 403.

PR #260, opened 2019-01-15
Remove explorer-server from zowe-start.sh

Joe-Winchester:


Joe-Winchester:
Fixes https://github.com/zowe/zowe-install-packaging/issues/258

Issue #308, opened 2019-01-27
.zowe_profile not found error during installation


jackjia-ibm:
With Zowe build 455:

If the system doesn't have ~/.profile and ~/.zowe_profile, the installation script will stop at an error:

```
/MV3B/tmp/jack/zowe-1.0.0/install:>zowe-install.sh
---------------------------------------------------------------------
.: /MV3B/tmp/jack/zowe-1.0.0/install/..//scripts/zowe-init.sh 54: .: zowe-install.sh 51: /u/jackjia/.zowe_profile: not found
```

PR #318, opened 2019-01-29
Ensure .zowe profile exists

John-A-Davies:
If the user's .profile did not exist, zowe-init.sh did not create the .zowe_profile, because it needed .profile to populate .zowe_profile.  However, the .zowe_profile is executed ("sourced") right afterwards, and is also referenced in 'persist( )'.

The fix is to always create .zowe_profile if it does not exist.  If the user wants new lines from their .profile to be used, they can delete the .zowe_profile file before install.


John-A-Davies:
The change to zowe-init.sh fixes #308.

PR #351, opened 2019-02-05
verify node.js version in post-install script

John-A-Davies:

The part of `zowe-verify.sh ` that checks node.js version did not detect if node was not found, and neither did it check the version was at least the level required.  

`zowe-check-prereqs.sh` does check the version was at least the level required, but users might not run that.


John-A-Davies:
OK.  Will fix in pre- and post-install verify scripts.

John-A-Davies:
Both scripts now check for minimum v6.14.4.

PR #356, opened 2019-02-06
Switch to nodeCluster instead of nodeServer

1000TurquoisePogs:
Use nodeCluster instead of nodeServer as it is crash-resistent for uncaught exceptions

### DIFF
--- a/scripts/run-zowe.sh
+++ b/scripts/run-zowe.sh
@@ -22,7 +22,7 @@ if [[ ! -f $NODE_HOME/"./bin/node" ]]
 then
 export NODE_HOME=$nodehome
 fi
-`dirname $0`/../../zlux-app-server/bin/nodeServer.sh --allowInvalidTLSProxy=true &
+`dirname $0`/../../zlux-app-server/bin/nodeCluster.sh --allowInvalidTLSProxy=true &
 `dirname $0`/../../api-mediation/scripts/api-mediation-start-discovery.sh
 `dirname $0`/../../api-mediation/scripts/api-mediation-start-catalog.sh
 `dirname $0`/../../api-mediation/scripts/api-mediation-start-gateway.sh


hogstrom:
I've tested this PR against my system.  The startup is a bit slower (expected because of multiple servers).  I dont see an issue with this fix.

MarkAckert:
I would prefer to switch to clustering post-1.0.0. A new deployment introduces some risk to the build, and I'm unclear on the strength of clustering within a single address space - anything which takes the address space out takes out the entire cluster as well. If an uncaught exception takes one zlux instance down within the address space, why would the same uncaught exception not take place on the clustered instance if an exact follow-up request is made? How common are uncaught exceptions taking out the zlux server instance?

1000TurquoisePogs:
Mark - node clustering logic is being used to spin up a process in the cluster if one dies, whereas when clustering is not used, your entire server is gone. 
Uncaught exceptions can happen due to bugs, especially asynchronous ones, as you can't throw a big try-catch around an async function.
We used to use a "handler of last resort" that nodejs has, but they recommend never using it because it means to leak memory: https://github.com/zowe/zlux-server-framework/blob/master/lib/process.js#L28

1000TurquoisePogs:
An alternative we could do is to bubble up the behavior of "die or leak" to the admin to decide on... here we're defaulting to "true=die" https://github.com/zowe/zlux-server-framework/blob/master/lib/index.js#L43

1000TurquoisePogs:
Yet another alternative is "just patch the hole for now": https://github.com/zowe/zlux-server-framework/pull/70

jackjia-ibm:
**After we have `staging` branch, this PR should be adjusted to merge to `staging` instead of `master`.**

1000TurquoisePogs:
I think this is ready for merging whenever. Originally we didn't want to do it when the staging branch didnt exist, but now it does.

John-A-Davies:
This change requires a change to `zowe-verify.sh`, to cater for the 2 extra STCs that arise from the use of `nodeCluster`.

jackjia-ibm:
This work has also been implemented and handled in https://github.com/zowe/zowe-install-packaging/pull/456, so it's duplicated now. This change will go into v1.2.0 release.

PR #382, opened 2019-02-20
replace csp frame-ancestors config

jackjia-ibm:


jackjia-ibm:
Smoke test passed on this build (https://wash.zowe.org:8443/job/zowe-install-test/job/explorer-cpf/2/) and CSP header is added into response header.

Issue #389, opened 2019-02-28
Add TSS Security Commands for ZSS Install Scripts


MarkAckert:
Add and validate TSS security commands for ZSS. Validation may be performed on private infrastructure. Documentation should be updated to reflect the commands.

hogstrom:
@MarkAckert is this something we should move to an active issue?

stevenhorsman:
With the updates to ZWESECUR ACF2 and TSS equivalents are in the JCL

PR #391, opened 2019-02-28
Add TSS commands to ZOWE install scripts

lvuono:


MarkAckert:
Fixes #389 . Will wait for PAX from this PR, and then test on internal system.

MarkAckert:
This PR is now ready for review; I manually tested the Top Secret with an already-complete security setup; partially-complete security setup, and a blank security setup. No issues encountered.

PR #394, opened 2019-03-01
Change api def to include unix file endpoint

jordanCain:
Signed-off-by: Jordan Cain <jordan.cain1@uk.ibm.com>

### DIFF
--- a/scripts/zowe-api-mediation-configure.sh
+++ b/scripts/zowe-api-mediation-configure.sh
@@ -152,20 +152,22 @@ iconv -f IBM-1047 -t IBM-850 $TEMP_DIR/zosmf.yml > $STATIC_DEF_CONFIG/zosmf.yml
 cat <<EOF >$TEMP_DIR/datasets.yml
 #
 services:
-  - serviceId: datasets
-    title: IBM z/OS Datasets
-    description: IBM z/OS Datasets REST API service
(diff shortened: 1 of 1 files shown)


scripts/zowe-api-mediation-configure.sh
```diff
@@ -152,20 +152,22 @@ iconv -f IBM-1047 -t IBM-850 $TEMP_DIR/zosmf.yml > $STATIC_DEF_CONFIG/zosmf.yml
 cat <<EOF >$TEMP_DIR/datasets.yml
 #
 services:
-  - serviceId: datasets
-    title: IBM z/OS Datasets
-    description: IBM z/OS Datasets REST API service
-    catalogUiTileId: datasets
+  - serviceId: datasetsAndUnixFiles
```
jackjia-ibm: If we rename this service id, will the URL be converted to https://<gateway>:7554/api/v1/datasetsAndUnixFiles?

jordanCain: Yes you're right I'm just validating serving under a different service ID

jackjia-ibm: To make less impact, we can define another unixfiles service. So the datasets endpoint may be left unchanged.

PR #410, opened 2019-03-13
Use individual zLUX artifacts & put zLUX release configuration

dmitrynikolaev:
That's a part of https://github.com/zowe/zlux/issues/110

Due to the zLUX capstone changes:
- packaging should use individual artifacts for each plugin and zLUX framework
- release configuration should be defined in the packaging repository

Signed-off-by: Dmitry Nikolaev <dnikolaev@rocketsoftware.com>

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -130,7 +130,7 @@ sed -e 's/{BUILD_BRANCH}/${env.BRANCH_NAME}/g' \
       echo downloadResult
       def downloadResultObject = readJSON(text: downloadResult)
       if (downloadResultObject['status'] != 'success' ||
-          downloadResultObject['totals']['success'] != 9 || downloadResultObject['totals']['failure'] != 0) {
+          downloadResultObject['totals']['success'] != 18 || downloadResultObject['totals']['failure'] != 0) {
         echo "status: ${downloadResultObject['status']}"
         echo "success: ${downloadResultObject['totals']['success']}"
         echo "failure: ${downloadResultObject['totals']['failure']}"
--- a/artifactory-download-spec.json.template
+++ b/artifactory-download-spec.json.template
@@ -1,7 +1,52 @@
 {
   "files": [{
-      "pattern": "rocket-closed-source-builds/1.0.1+20190228/ZLUX.pax",
-      "target": "pax-workspace/content/zowe-{ZOWE_VERSION}/files/",
+      "pattern": "libs-snapshot-local/org/zowe/zlux/zlux-core/1.0.1-STAGING/zlux-core-1.0.1-20190313.111628.pax",
+      "target": "pax-workspace/content/zowe-{ZOWE_VERSION}/files/zlux/zlux-core.pax",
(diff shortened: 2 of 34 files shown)


MarkAckert:
Hi @dmitrynikolaev, Are there corresponding pipeline / build changes in zLUX I can review as part of this? I would like to understand the complete end-to-end flow of zLUX.

I have some initial concerns around the configuration files living here as opposed to bundled with some of the 'pax' artifacts. Also curious about the decision to 'pax' artifacts like sample-angular-app as opposed to 'zip' which would be more portable. I need to review these in the context of the complete pipeline and intent behind each artifact's structure.

dmitrynikolaev:
Hi @MarkAckert ,

The pipelines (there're still under review):
https://wash.zowe.org:8443/job/zlux-core/
https://wash.zowe.org:8443/job/zlux-plugins/

As for release configuration files, our intention was to separate dev and release configs, and the repo
seemed to be appropriate place.
@1000TurquoisePogs , any comments on that?

As for zip / pax:
- we may want to keep file permission information for some plugins
- I don't really know if we can easily unzip files on mainframe and convert everything to EBCDIC on the fly. Can we? Packaging pre-converted files would break portability.
- Usage of `pax` is just consistent with other artifacts.

1000TurquoisePogs:
Hi Mark, a few things:

1. I'd also like to see zips for apps. I'm actually working on this via https://github.com/zowe/zlux-server-framework/pull/57 where an app archive model based on zip is being used. In the future, we'll then be able to then have apps installed by invoking an install script rather than needing to create those small JSON files.

2. the repo zlux-app-server is a collection of scripts and config defaults. The config defaults made assumptions about being included into the official zowe release, whereas if someone wanted to start from the source, the config defaults may be less trivial than desired (maybe people dont want 5 different Apps... maybe they want 0. Maybe they want to test the Desktop without installing ZSS). So, we want to move the release defaults out of that repo, and it seems like it is best to have it in **some** repo so that people can see what those will be, and when they change. Wouldn't this repo be the best place for that, or is there another?

dmitrynikolaev:
@MarkAckert , 
Are you OK with this PR? Can we merge it?

MarkAckert:
@dmitrynikolaev I am in the process of reviewing, however I think we should off at minimum until the pipeline reviews are complete and become the source of truth for zLUX builds - right now it looks like they're being built from your fork (which I assume has a corresponding PR with zowe/zlux* repos).

1000TurquoisePogs:
OK - the PRs were merged to the staging branches of the zlux repos. Is master here now going to be based on staging of those? Or should this be retargeted to staging/v1.1.x

dmitrynikolaev:
@MarkAckert ,
The pipelines are active now, all the artifacts have been re-built from zlux*.

dmitrynikolaev:
BTW, there're no "pinned" plugins in the build because of this deploy script issue:
https://github.com/zowe/zlux-build/issues/21

It will be fixed separately.

PR #439, opened 2019-04-15
use tls1.2 restricted version of explorer components

jackjia-ibm:

This is a demonstration build which has explorer TLS v1.2 changes. This build will fail more test cases, so we may need to revert the changes once the debugging session is done.


jackjia-ibm:
The change had been handled by other PR.

PR #448, opened 2019-04-25
Update zowe-install.yaml.template

patbossman:
Added a comment to clarify the directory is install of a zowe instance.
There has been some confusion between where zowe is unpax'd, which has been called the install directory, and where a zowe instance is installed.  Also wanted to specify the default directory goes into the home directory of the user running the script.  There are folks new to *nix, and this is meant as a gentle clue.


jackjia-ibm:
@patbossman, about to merge this change. Could you please fix the DCO sign-off error? You can follow instruction from https://github.com/zowe/zowe-install-packaging/pull/448/checks?check_run_id=135509070. Thanks.

PR #451, opened 2019-04-26
use Node cluster instead of node server

John-A-Davies:
Prevents AppScan from disabling node server by using node cluster instead.  The verify script is also updated to take account of this new configuration.  

The change to zowe-init.sh is not part of this PR.


jackjia-ibm:
We have a replacement https://github.com/zowe/zowe-install-packaging/pull/451 to merge into `staging`.

Issue #454, opened 2019-04-29
Verify script should check that the xmem server is up


John-A-Davies:
The ZSS Cross-Memory Server must be up and running, and offering its service with the right name to the other Zowe STCs.  

Typical errors include
 - STEPLIB DSN is not APF-authorised

stevenhorsman:
As of 1.8 documentation state that zowe-verify is not update and from 1.9 it will be removed, so closing this issue

PR #459, opened 2019-04-30
Add check of ZSS to zowe-verify script

John-A-Davies:
Check the following ZSS server items
 - Is program ZWESIS01 running?
 - Is program ZWESIS01 running under user IZUSVR?
 - Try to determine ZSS server job name
 - Is ZSS server job running as a STC?
 - Is the status of the ZSS server OK in the zssServer log in this ZOWE_ROOT_DIR?


John-A-Davies:
Fixes #454

PR #488, opened 2019-05-21
use derived IP if user enters blank

John-A-Davies:


Joe-Winchester:
Not needed as this is superceded by a later pull request that gets both the blank IP in as well as the fix for the trailing blank breaking the SED

John-A-Davies:
The superseding PR is #491

PR #496, opened 2019-05-23
include upgraded zlux components into staging

jackjia-ibm:
I saw zlux had been upgraded to 1.2.0, adjust to pick up new versions.

Please double check some components doesn't have 1.2.0-STAGING, but have 1.2.0-RC instead.

Thanks.


1000TurquoisePogs:
The naming is sort of misleading, since the versions of the apps do not correspond to the version of zowe as a whole. Several apps haven't been version bumped properly, and so we need a better way to do that.

dmitrynikolaev:
Staging of zlux is broken.
Jordan has made a fix:
https://github.com/zowe/zlux-server-framework/pull/101

Issue #494, opened 2019-05-23
Configuration scripts will fail if ZOWE environment variables contain blanks


John-A-Davies:
The configuration scripts use `sed` to edit the configuration files.  About 30 instances use a `sed` string quoted like this

    's/\*\*ZOSMF_IP\*\*/'$ZOWE_IPADDRESS'/g'

and if $ZOWE_IPADDRESS contains any blanks, the script will fail with

    sed: FSUM7296 newline or end of file found in pattern

Some ZOWE variables can be entered by the user at config time, and these should be checked when they are entered.  But for extra robustness, the `sed` strings should avoid making `sed`  fail by using a syntax like 

    "s/\*\*ZOSMF_IP\*\*/$ZOWE_IPADDRESS/g"

PR #506, opened 2019-05-30
Sed blanks

John-A-Davies:
If the user enters blanks in the input string for any of the env variables, some of the `sed` commands fail because the way the `sed` string is quoted induces a syntax error if the variable has embedded blanks.  The `sed` strings have been re-formed to eliminate the syntax errors.


John-A-Davies:
Fixes #494

PR #523, opened 2019-06-13
Apply jenkins library

jackjia-ibm:


jackjia-ibm:
Replaced with https://github.com/zowe/zowe-install-packaging/pull/524

PR #559, opened 2019-07-19
Xmc copy temp

stevenhorsman:


Joe-Winchester:
I managed to install the XMEM server from the runtime on 3b into my datasets OK - the only thing that failed is the RACF PERMIT because I didn't have permission.  The errors are the same as if I ran the script out of the install folder and these steps are "one time only" versus install a new server so merging this change to get it into a 1.4 build.  I also got it to work on ukzowe3 except for some permission problems that have nothing to do with the script

PR #568, opened 2019-07-23
use proper tag/branch for staging, add vscode extension into source d…

jackjia-ibm:
…ependencies, add destinations


Joe-Winchester:
*** this  PR is setting up for 1.5 by switching locked-in tags to `master` branches…so we should not merge ***..  i.e. don't include this in 1.4 or merge without checking with @MarkAckert.

PR #593, opened 2019-08-23
introduced zssServerName in zowe-install.yaml

vvvlc:
Signed-off-by: Vitek Vlcek <vitezslavvit.vlcek@ca.com>

* introduced a new item `zssServerName=ZWESIS_STD` in section `install` of `zowe-install-apf-server.yaml`
  to override a parameter value `NAME='ZWESIS_STD'` in `zss\samplib\zis\ZWESIS01`  STC PROC
* introduced a new item `zssServerName=ZWESIS_STD` in section `zlux-server` of `zowe-install.yaml` 
 to override `"privilegedServerName": "ZWESIS_STD"` in `zluxserver.json`


stevenhorsman:
@vvvlc - I think we need to be very careful with the naming here. There is a lot of confusion and people often call ZWESIS 'ZSS', but my understanding zss and cross memory server are different (see https://zowe.org/docs-site/latest/getting-started/zowe-architecture.html) so I think you might want to check this

stevenhorsman:
Ok - Sean approved whilst I was writing, so I assume it's just me that's confused!

ifakhrutdinov:
Technically, it's the name of a component and not ZSS. I would call it ```zssCrossMemoryServerName```, that way, if we need to identify each ZSS (non privileged HTTP server) with a name in the future, we could use ```zssServerName```.

vvvlc:
* thank you folks for the feedback, I will rename `zssServerName` to `zssCrossMemoryServerName` in both yaml files.
 * Initially I thought that user would have to set `zssCrossMemoryServerName` in both `zowe-install.yaml` and  `zowe-install-apf-server.yaml`, but it looks that `zowe-copy-xmem.sh` executed from  `zowe-install.sh` creates `xmem-server` folder in `ZOWE_RUNTIME` directory.
  Therefore I modify `zowe-copy-xmem.sh` to propagate value of `zssCrossMemoryServerName` from `zowe-install.yaml` to `zowe-install-apf-server.yaml` so user will have to do modification in one place.

ifakhrutdinov:
@vvvlc I think the ```ZOWE_ZSS_SERVER_NAME ``` variable should also have some indication that it's the cross memory server's name, maybe ```ZOWE_ZSS_XMEM_SERVER_NAME ```.

Issue #582, opened 2019-08-01
Implement new naming scheme for ZOWESVR started tasks


John-A-Davies:
Revise the STC naming scheme as follows;

In the YAML template file, you will now see

    install:prefix 
    install:instance

which will appear like this

    install:
      rootDir=~/zowe/{ZOWE_VERSION}
      prefix=ZOWE
      instance=1

Together, these tags define a prefix for Zowe address space STC names associated with 
USS processes. Each component has its own STC name according to the following format:

    pfxNss

where:
• pfx - prefix that contains up to four characters, for example, ZOWE.
• N   - Zowe instance number or letter

• ss - a component identifier that consists of 2 characters:

• AG - API ML Gateway
• AD - API ML Discovery Service
• AC - API ML Catalog
• EJ - Explorer API Jobs
• ED - Explorer API Data Sets
• UD - Explorer UI Data Sets
• UJ - Explorer UI Jobs
• UU - Explorer UI USS
• DT - Zowe Desktop Application Server
• SV - Master STC that spawns the others

Joe-Winchester:
I like this a lot as it reduces the current 3 letter prefix to classify address spaces into 2.  @vvvlc - are you happy that the instance number gets moved from the end to the middle, so ZOWEAAG1 and ZOWEADS1 now would become ZOWE1AG and ZOWE1AD ?   The reason is that it is easier in SDSF to do a PRE ZOWE1* (or PRE ZOWE2*).  Eventually we may move from ZOWE to ZWE which would give us another letter at the end for duplicate address spaces

John-A-Davies:
I agree.  

Also note that to keep all the STC names unique for one instance of ZOWE, the jobname in the START command in `zowe-start-template.sh` must be 7 or fewer characters.  This is because MVS appends a digit to the sibling job spawned from the master server job, for example ZOWE1SV, giving ZOWE1SV2.  The other job names are fine, they can have 8 characters.

vvvlc:
I think the instance number in @plavjanik original post (https://github.com/zowe/zowe-install-packaging/issues/385#issuecomment-489555612) was meant for additional instances from HA perspective.  Eg:

one gateway and discovery services in non HA mode would have these jobnames
* ZOWEAGW1
* ZOWEADS1

while in HA mode where we have many instances of same services would result in
* ZOWEAGW1
* ZOWEAGW2
* ZOWEADS1
* ZOWEADS2

[...]

PR #594, opened 2019-08-23
Stc names 582

John-A-Davies:
This updates the STC job names according to the naming scheme in issue #582.  The `zowe-verify.sh` script already caters for this new scheme, and is already delivered.


John-A-Davies:
Oops.  I want to commit this to staging, but I don't know how to change that.  Help, anybody?

Issue #427, opened 2019-04-09
z/OSMF running of different host than zowe core


vvvlc:
Hi

I think zowe does not support z/OSMF on other host in sysplex than zowe core is running.

I had to change 
 * explorer-data-sets-api/scripts/data-sets-api-server-start.sh
   *  `-Dzosmf.ipAddress=10.175.84.31`
 * explorer-jobs-api/scripts/jobs-api-server-start.sh
   *  `-Dzosmf.ipAddress=10.175.84.31`
 *  api-mediation/api-defs/zosmf.yml
  *   ```yaml
      instanceBaseUrls:
        - https://USILCA32.lvn.broadcom.net:1443/zosmf/```

configuration scripts needs to be adjusted.
 * `~/.zowe_profile` is missing `ZOWE_ZOSMF_HOST` variable
 * use `ZOWE_ZOSMF_HOST` instead of `ZOWE_IPADDRESS` in https://github.com/zowe/zowe-install-packaging/blob/adb2b31ef475955641ce450748cbadae271363eb/scripts/zowe-explorer-api-configure.sh#L63
* use `ZOWE_ZOSMF_HOST` insted of `ZOWE_EXPLORER_HOST~ in https://github.com/zowe/zowe-install-packaging/blob/adb2b31ef475955641ce450748cbadae271363eb/scripts/zowe-api-mediation-configure.sh#L133
  
Vitek

vvvlc:
Fixed in https://github.com/zowe/zowe-install-packaging/pull/598   this should be available in zowe 1.5

PR #598, opened 2019-08-28
Zosmf overide

vvvlc:
This is a fix for  https://github.com/zowe/zowe-install-packaging/issues/427, when z/OSMF runs on another host than whole zowe (APIML, Explorer APIs, ZAF), user performing installation has to carry out some post installation changes.

This PR introduces a new environment variable `ZOWE_ZOSMF_HOST` that is used like `ZOWE_ZOSMF_PORT`. 
To exploit it execute
```sh
export ZOWE_ZOSMF_HOST=myzosfm.customhostname.abc
export ZOWE_ZOSMF_PORT=1443
zowe-install.sh
```
if `ZOWE_ZOSMF_HOST` is not specified it defaults to `ZOWE_EXPLORER_HOST`.

`ZOWE_ZOSMF_HOST` is used in these places:
 *   `scripts/configure/zowe-configure-api-mediation.sh`
    from `https://$ZOWE_EXPLORER_HOST:$ZOWE_ZOSMF_PORT/zosmf/` 
    to `https://$ZOWE_ZOSMF_HOST:$ZOWE_ZOSMF_PORT/zosmf/`
 * `scripts/configure/zowe-configure-explorer-api.sh`
  from  `-e "s/\*\*ZOSMF_IP\*\*/$ZOWE_IPADDRESS/g" \` 
  to `-e "s/\*\*ZOSMF_IP\*\*/$ZOWE_ZOSMF_HOST/g" \` 

As we can see there are two variables `ZOWE_EXPLORER_HOST` and `ZOWE_IPADDRESS` used to point to z/OSMF, I decided that I would rather use hostname rather than ip address, hopefully this will not cause any issue because both hostname and ip address has been used until now.

NOTE: this PR is based on https://github.com/zowe/zowe-install-packaging/pull/593

### DIFF
--- a/scripts/configure/zowe-configure-api-mediation.sh
+++ b/scripts/configure/zowe-configure-api-mediation.sh
@@ -16,6 +16,8 @@
 # $ZOWE_ROOT_DIR
 # $ZOWE_EXPLORER_HOST
 # $ZOWE_IPADDRESS
+# $ZOWE_ZOSMF_HOST
+# $ZOWE_ZOSMF_PORT
 # $ZOWE_APIM_CATALOG_PORT
 # $ZOWE_APIM_DISCOVERY_PORT
 # $ZOWE_APIM_GATEWAY_PORT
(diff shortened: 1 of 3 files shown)


stevenhorsman:
From discussed with Vitek - "I would keep it as environment variable for 1.5 as an experimental then I think we can document it, not sure if these two (ZOSMF_HOST and PORT) should be available in zowe-install.yaml later"

PR #607, opened 2019-08-28
Stc names 582

John-A-Davies:
The replacement of zowe_instance was omitted, and this PR fixes that.

PR #608, opened 2019-08-29
redo #595 + #607

John-A-Davies:


scripts/configure/zowe-configure.sh
```diff
@@ -66,7 +66,7 @@ fi
 zluxserverdirectory='zlux-app-server'
 echo "Preparing folder permission for zLux plugins foder..." >> $LOG_FILE
 chmod -R u+w $ZOWE_ROOT_DIR/$zluxserverdirectory/plugins/
-chmod -R u+w $ZOWE_ROOT_DIR/$zluxserverdirectory/deploy/site
```
stevenhorsman: Why has this been removed?

scripts/configure/zowe-configure.sh
```diff
@@ -99,25 +117,6 @@ chmod a+x "${ZOWE_ROOT_DIR}/scripts/zowe-support.sh"
 echo "Attempting to setup Zowe Proclib ... "
 . $CONFIG_DIR/zowe-configure-proclib.sh
 
-# TODO LATER - same as the above - zss won't start with those permissions,
-sed -e "s#{{root_dir}}#${ZOWE_ROOT_DIR}#" \
-  -e "s#{{zosmf_admin_group}}#${ZOWE_ZOSMF_ADMIN_GROUP}#" \
```
stevenhorsman: Why has this code block been moved and bits been deleted?

John-A-Davies: BAU changes happened in between.  Fixing this now.

scripts/configure/zowe-configure.sh
John-A-Davies: It should be back in now.

scripts/zowe-parse-yaml.sh
```diff
@@ -196,93 +201,98 @@ parseConfiguationFile ./zowe-install.yaml
 if [[ $ZOWE_ROOT_DIR == "" ]] 
 then
     ZOWE_ROOT_DIR="~/zowe/$ZOWE_VERSION"
-    echo "  ZOWE_ROOT_DIR not specified:  Defaulting to ~/zowe/$ZOWE_VERSION"
+    echo "  ZOWE_ROOT_DIR not specified:  Defaulting to ~/zowe/$ZOWE_VERSION" >> $LOG_FILE
```
stevenhorsman: These errors should be written to the log, but should also come out on stdout so the user can see that their configuration is incorrect

scripts/zowe-support.template.sh
```diff
@@ -40,7 +40,10 @@ if [[ -z "${ZOWE_PREFIX}" ]];then
     echo "The ZOWE_PREFIX environment variable wasn't properly populated during install. Exiting."
     exit
 fi
-
+if [[ -z "${ZOWE_INSTANCE}" ]];then
```
stevenhorsman: ZOWE_INSTANCE doesn't seem to be defined in this file, so this if block will always trigger. I also don't see the need for it

scripts/zowe-parse-yaml.sh
John-A-Davies: OK, I will do that.

stevenhorsman:
A value of ZOWE_INSTANCE is still missing from the default yaml

stevenhorsman:
I'd like to see PR #611 merged in to clean up the code, but approving on the basis that Joe has assured me the community have approved the new component naming scheme implemented in this change

PR #611, opened 2019-08-29
Reduce duplicate code

stevenhorsman:


John-A-Davies:
But `tee` doesn't append unless you use -a, so

`echo "... message..." | tee -a $LOG_FILE`

Issue #314, opened 2019-01-29
Install scripts use literal default values instead of referring to YAML keys


GrahamHannington:
Some Zowe install scripts use literal default identifier values instead of referring to YAML keys.

Values include (there might be others) the following z/OSMF defaults:

* `IZUSVR` (already has the YAML key `zosmfUser`, but the scripts do not always use that key)
* `IZUADMIN`
* `IZUUSER`

For example, in `zowe-check-prereqs.sh` (which extracts other YAML key values):

    tsocmd lg izuadmin 2>/dev/null |grep IZUSVR >/dev/null

and in `zowe-runtime-authorize.sh`:

    chgrp -R IZUUSER ./zlux-app-server/deploy

If z/OSMF is already installed, and whoever installed it didn't use the default values for these identifiers, then you have to debug the Zowe installation scripts (identify which scripts use these values and edit them).

stevenhorsman:
Since 1.8 the yaml file and IZU user and groups are not used with the zowe server, so from 1.9 dev build onwards these changes have been removed

PR #736, opened 2019-09-19
IZUSVR read access to BPX.JOBNAME

John-A-Davies:
Add IZUSVR to IZUADMIN as the users to check for READ access to BPX.JOBNAME, and reflect this in the error message.


stevenhorsman:
We should be reading in the user name and group from the yaml file as per https://github.com/zowe/zowe-install-packaging/issues/314

John-A-Davies:
I'd like to keep the fix for #736 (which hits every user, and needs to be done anyway, because they will always see false errors if IZUSVR is not explictly able to read BPX.JOBNAME) separate from #314 (which hits only those few who don't use IZUSVR/IZUADMIN, and is a pervasive code change)

OnnoVdT:
Isn't it &ZOWEUSER who needs this permit? And if so, shouldn't we do this in ZWESUCUR?

stevenhorsman:
@OnnoVdT - I think this is just a very old PR relating to the verify script which is broken and removed from doc, so we can close it.

OnnoVdT:
@stevenhorsman 
While it's true this is old and parts of it no longer apply, the BPX.JOBNAME permit itself is still very valid. If you don';t have the permit, you cannot start tasks with a given name, and that's what &ZOWEUSER now does for all the child servers.

Not sure if we should reopen this one and adjust title, or open a new one.

https://www.ibm.com/support/knowledgecenter/en/SSLTBW_2.3.0/com.ibm.zos.v2r3.bpxb200/fclass.htm
BPX.JOBNAME

Controls which users are allowed to set their own job names by using the _BPX_JOBNAME environment variable or the inheritance structure on spawn. Users with READ or higher permissions to this profile can define their own job names.

stevenhorsman:
@OnnoVdT we have https://github.com/zowe/zowe-install-packaging/issues/1049 to cover adding BPX.JOBNAME to ZWESECUR - this item was about changing the messaging in a script that is no longer based on an id that we aren't using, which is why I closed it.

PR #768, opened 2019-10-04
Smpe.pax

OnnoVdT:
Updates to smpe.pax. No dependencies on other updates.

### DIFF
--- a/smpe/pax/MVS/ZWE1SMPE.jcl
+++ b/smpe/pax/MVS/ZWE1SMPE.jcl
@@ -23,7 +23,7 @@
 //* 1) Add the job parameters to meet your system requirements.
 //*
 //* 2) Change #csihlq to the high level qualifier for the CSI and
-//*    other SMP/E data sets. The maximum length is 35 characters.
+//*    other SMP/E data sets. The maximum length is 26 characters.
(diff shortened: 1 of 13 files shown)


stevenhorsman:
The SMPE build fails as the script has changed to accept -V, smpe-fmid.sh call `$here/$allocScript -L "$VOLSER" "$2" "$3" "$4" "$5" "$6"`
When you haves PRs with SMPE changes in future it is probably best to re-run with BUILD_SMPE parameter set on like I did so you can see any errors.

OnnoVdT:
I agree with your comment. Thinking about it now, I probably would have been able to figure out how to get a build started for the smpe.pax branch and see the error. 
The plan was to have all updates in, including changes in other areas to not share the allocate-dataset script of smpe.pax and thus cope with the -V change. As mentioned in slack, this didn't work out and I got stuck half-way. At that point I just gave up and started on new coding, waiting for you to be back online and help me through my issues.

I tried git revert to undo just the -V update, but that isn't working (probably doing it wrong), locally it's reverted, but github still says 2 pending commits. If there is an easy solution, can you tell me how to fix it? If not, jsut trash this and I'll start over again.

PR #769, opened 2019-10-05
Build script updates. No dependencies on other updates.

OnnoVdT:
Signed-off-by: OnnoVdT <49648041+OnnoVdT@users.noreply.github.com>

### DIFF
--- a/smpe/bld/allocate-dataset.sh
+++ b/smpe/bld/allocate-dataset.sh
@@ -1,0 +1,174 @@
+#!/bin/sh
+#######################################################################
+# This program and the accompanying materials are made available
+# under the terms of the Eclipse Public License v2.0 which
+# accompanies this distribution, and is available at
+# https://www.eclipse.org/legal/epl-v20.html
+#
+# SPDX-License-Identifier: EPL-2.0
+#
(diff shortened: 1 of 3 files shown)


stevenhorsman:
These scripts are currently in the /smpe/pax/scripts directory. We shouldn't have 2 copies in the repo, so should they be moved from there and the references to those versions be updated to point to the new ones?

OnnoVdT:
The SMP/E build, SMP/E install and convenience build all need to allocate a dataset, but are all separate entities. If they share code that is buried in one of them, it is not obvious that the code is shared, and any update to it will have a ripple effect outside the intended target. So either each ships their own copy that cans tart living its own life, or the shared is moved to a location for which it is more obvious it is shared.

stevenhorsman:
> the shared code is moved to a location for which it is more obvious it is shared.
I think this should be the solution, but creating multiple copies of something that is supposed to be the same doesn't work, it's just a maintenance nightmare.

OnnoVdT:
Going to move the scripts to a shared location. Must be coordinated with updates to smpe.sh scripts and scripts that create convenience pax and smpe.pax

PR #779, opened 2019-10-09
improved symlink processing, first part in PR 772

OnnoVdT:


OnnoVdT:
opened against master

PR #780, opened 2019-10-09
simplify split logic

OnnoVdT:


OnnoVdT:
comes **AFTER** PR 781

stevenhorsman:
Tested: https://wash.zowe.org:8443/job/zowe-install-test/job/staging/324/

stevenhorsman:
This is fine for now - I think as more items get componentised we might want to split the children of the components directory to avoid that getting too big.

PR #784, opened 2019-10-09
update zowe-install.sh

OnnoVdT:


stevenhorsman:
What is the purpose of this PR - there's a massive refactor of the convienience build which doesn't seem to line up with the existing strategy and current work items

OnnoVdT:
moving main update to backlog

PR #785, opened 2019-10-09
update zowe-install-api-mediation.sh

OnnoVdT:


OnnoVdT:
moving main update to backlog

PR #786, opened 2019-10-09
updating zowe-install-explorer-api.sh

OnnoVdT:


OnnoVdT:
moving main update to backlog

PR #787, opened 2019-10-09
update zowe-install-explorer-ui.sh

OnnoVdT:
update zowe-install-explorer-ui.sh


OnnoVdT:
moving main update to backlog

PR #788, opened 2019-10-09
update zowe-install-MVS.sh

OnnoVdT:


OnnoVdT:
moving main update to backlog

PR #800, opened 2019-10-11
Samplib

OnnoVdT:
add ZWESECUR.jcl
update zowe-install-MVS.sh to process ZWESECUR.jcl & fix issues
* faulty copyright
* do not delete whole $TEMP_DIR
* process all members in LOADLIB/SAMPLIB to avoid that ZSS folks need to update install script when adding a member
* grouped temp code that alters input to simplify cleanup
* made dataset size a variable to simplify updates


OnnoVdT:
1. Regarding pre-z/OS 2.1 support comments. I just copied the member I created a few months ago without review. I removed the comment.
2. Regarding the # rexx. TSO will interpret everything in SYSTSIN as a command to execute, so if you add comments inside SYSTSIN you will get a bunch of error messages, and possibly actual execution of a comment word as command. By creating a dummy REXX named # and ensuring TSO can find it, you can start all comment lines with this character and TSO will not complain as it sees the comment as startup argument for the REXX.
Since there is no way to be sure the customer already has such a rexx, we create it dynamically in a temp dataset that is trashed once the JCL finishes.
3. Regarding ZWES.IS. Good catch, either I misread back when, or the profile name changed. This is fixed.

OnnoVdT:
thanks, fixed

OnnoVdT:
done

OnnoVdT:
looks like the DATA, part didn't make it in my copy/paste. Now I use the correct DATA,DLM=

stevenhorsman:
SMPE tested in https://wash.zowe.org:8443/job/zowe-install-test/job/staging/357

PR #805, opened 2019-10-14
PTF prep

OnnoVdT:
* add random MLQ to avoid conflict during concurrent runs
* updated zowe.yaml customization to set HLQ
* updated .zowe_profile processing to avoid infinite loop
* moved temporary zowe.yaml & .zowe_profile processing out of smpe-install.sh to alter.sh
* updated zowe-install.sh to not fail on future startup args. This was smpe-install.sh can already pass them in and doesn't need to be updated once zowe-install.sh supports them
* put a toggle on -R processing to simplify future activation


stevenhorsman:
I think this PR depends on #800 before it can be merged?

OnnoVdT:
I used the samplib branch as base to keep my workspace up-to-date, but it's not absolutely necessary for this specific update. The samplib branch update is necessary for a (near) future PR where I actually package the datasets created. At that point the datasets must be complete, and the samplib branch adds the missing ZWESECUR JCL.

PR #824, opened 2019-10-21
drive GIMDTS to create PTF parts

OnnoVdT:
step 1 of PTF creation; update smpe-service.sh to analyze RELFILEs and create a job to convert members within to FB80 (using GIMDTS)


OnnoVdT:
Opened against master :(, will recreate against staging.

PR #826, opened 2019-10-21
Added ng2desktop settings to plugin list

sakeerthy:
Signed-off-by: Suneeth Keerthy <skeerthy@rocketsoftware.com>


1000TurquoisePogs:
I am merging this because I see no reason why this shouldn't be here. It seems like things would have been broken without it, so not sure how people got to the settings menu prior.

PR #830, opened 2019-10-26
update HLQ length comments

OnnoVdT:
The HLQ length differences in ZWE1SMPE and ZWE3ALLOC were questioned, so I revisited and added some comment on where the limits come from:
ZWE1SMPE: 
The longest DSN using CSIHLQ is <dsprefix>.<fmid>.F<xx>
an FMID is always 7 characters long and assuming we will never have more than 99 RELFILEs (2 digits), the fixed part uses maximum 12 characters (fmid, relfile counter, F and 2 periods), leaving 32 characters for the HLQ
ZWE3ALLOC:
Both target and distribution libraries have a fixed low level qualifier of maximum 8 characters, with the mandatory period the fixed part is maximum 9 characters, leaving 35 characters for the HLQ.


OnnoVdT:
adjusted comment in JCL

PR #858, opened 2019-11-18
Script to tag files by extension

1000TurquoisePogs:

#### PR type
- [ ] Bugfix 
- [x] Feature 

#### Relevant issues
Accommodates https://github.com/ibmruntimes/node/issues/142

This is a script that can be used in packaging pipelines to tag files. This script assumes the files came from a linux/windows machine in that it assumes files that are ASCII do not need to be converted to ASCII, just tagged. 

The exception is scripts in order to not break current use of shell invocation.
It's also recommended that JSON or YAML files that SED is used are set to 1047, so that the use of sed does not break. That is not covered in this script, as the files that sed is used on are not ALL JSON /YAML, but just a few.

- [ ] Yes
- [x] No

The script does not break anything, but use of this script should be tested thoroughly to make sure that the output is desired.

It is a script that should be run in a packaging pipeline to make sure that the files put on a z/os system are properly tagged, to avoid implicit character encoding behavior.

Not currently. I'm not yet sure that any doc changes are needed, but worst case just a mention of use of tagging & BPXK env vars.

### DIFF
--- a/scripts/tag-files.sh
+++ b/scripts/tag-files.sh
@@ -1,0 +1,218 @@
+#! /bin/sh
+# This program and the accompanying materials are
+# made available under the terms of the Eclipse Public License v2.0 which accompanies
+# this distribution, and is available at https://www.eclipse.org/legal/epl-v20.html
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+# Copyright Contributors to the Zowe Project.
+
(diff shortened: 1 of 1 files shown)


dmitrynikolaev:
Hi @1000TurquoisePogs,
Shouldn't it go to `staging`?

PR #865, opened 2019-11-20
Tagging files used by app server

1000TurquoisePogs:

We're moving towards having all of the app server content tagged so that nodejs can do the right thing.
James & Dmitry noticed that pinnedPlugins.json was causing bad behavior during our tagging efforts, so I did a sweep through this repo to tag other files that nodejs will be reading.
Seems everything is in 1047 already, so just tagging it as such.


dmitrynikolaev:
Hi @1000TurquoisePogs,
Shouldn't it go to `staging`?

stevenhorsman:
Tested: https://wash.zowe.org:8443/job/zowe-install-test/job/staging/819/

Issue #627, opened 2019-09-04
Componentise Zlux/Zss/zlux-app-server into cupids format


stevenhorsman:
**As a** Zowe user,  
**I want** to be able to start the zss and zowe desktop framework and apps with the configuration passed in at runtime and have a separate user_directory/workspace for my instance data
**so that** I can make changes to my zowe config without having to reinstall and lose my config.

*Details/notes*
- We'll follow the pattern set out in #433 for launching a component and apply it to Zlux app service
- [ ] During configure we'll need to inject new properties into run-zowe.sh
ZOWE_ZLUX_SERVER_HTTPS_PORT - rename this?
ZOWE_ZSS_SERVER_PORT - rename this?
ZOWE_ZSS_XMEM_SERVER_NAME - can we look like up/read from the xmem server rather than risk a customer listing twice differently?
- Create a new start.sh script that is created in <zowe_root_dir>/components/<component-name>/bin which runs `$DIR/../../zlux-app-server/bin && _BPX_JOBNAME=$ZOWE_DESKTOP ./nodeCluster.sh --allowInvalidTLSProxy=true &`?
- In <zowe_root_dir>/components/<component-name>/bin/configure.sh create a new configure.sh script that does all the configure set-up - sed injecting the properties into from shell variables (and others?) based on zowe-configure-zlux-certificates.sh (if required each time), zowe-configure-zlux-ports.sh. Check zowe-config.sh for anything else missing. It also need to inject the correct variables to put zlux behind the APIML gateway - if the gateway is in the launch config (maybe a separate task to document/track component dependencies?)
- [ ] Add the new component id to the LAUNCH_COMPONENTS list and remove the old code

- We'll create a follow up story about doing component validation for zlux app server similar to #445 after it is refactored to share logic

**Questions?**
- Does zowe-configure-zlux-certificates.sh need to be done every time, or just once?
- What is the component id for zlux?
- Sean says that some components have read-only defaults, should we keep these in runtime, or copy a default to the user-dir

#### Acceptance Criteria

*Given* an run-zowe.sh forked script with defining all the properties
    *When* I run start ZOWE to drive the script
    *Then* ZOWE should start successfully
    *And* when I look at the zowe job address spaces through display active, I can see the zss and desktop services active
    *And* The desktop should be available on port <ZOWE_ZLUX_SERVER_HTTPS_PORT> and zss work on <ZOWE_ZSS_SERVER_PORT>

[...]

PR #911, opened 2019-12-06
Component-ize zlux & zss

1000TurquoisePogs:
See https://github.com/zowe/zlux-app-server/pull/87 for more.
Requires https://github.com/zowe/zlux-app-server/pull/87 and https://github.com/zowe/zlux-build/pull/42

#### PR type
- [ ] Bugfix 
- [x] Feature 

#### Relevant issues
Fixes https://github.com/zowe/zowe-install-packaging/issues/627

- [x] Yes
- [ ] No

Many paths & default paths have changed. The behavior of the servers remains the same, but anything relying on specific paths may fail. This is not a drop-in upgrade over a prior release, although manual steps can be taken to do so.

See https://github.com/zowe/zowe-install-packaging/issues/627


manifest.json.template
```diff
@@ -12,68 +12,79 @@
   },
   "binaryDependencies": {
     "org.zowe.zlux.zlux-core": {
-      "version": "~1.7.0-STAGING",
+      "version": "~1.8.0-STAGING",
```
stevenhorsman: It looks like you've not reverted the manifest values to point to the correct staging branches

scripts/configure/zowe-configure-iframe-plugins.sh
```diff
@@ -16,29 +15,24 @@
 
 if [ -z "$NODE_BIN" ]
 then
-  NODE_BIN=${NODE_HOME}/bin/node
+  if [ -n "$NODE_HOME" ]
+  then
+    NODE_BIN=${NODE_HOME}/bin/node
+  else
+    NODE_BIN=node
+  fi
```
stevenhorsman: After configure-node.sh NODE_HOME should always be set, so this shouldn't be needed

bin/internal/run-zowe.sh
```diff
@@ -78,6 +78,17 @@ then
   LAUNCH_COMPONENTS=${LAUNCH_COMPONENTS},files-api,jobs-api,api-mediation,explorer-jes,explorer-mvs,explorer-uss #TODO this is WIP - component ids not finalised at the moment
 fi
 
+#Explorers may be present, but have a prereq on gateway, not desktop
+if [[ $LAUNCH_COMPONENT_GROUPS == *"DESKTOP"* ]]
+then
+  LAUNCH_COMPONENTS=${LAUNCH_COMPONENTS},app-server,zss
```
stevenhorsman: Where is components/zss created as it doesn't appear on my, or the Marist install?

stevenhorsman:
When the test are run on Marist the desktop is not accessible and trying to test without zss in the launch component list the following error is in the logs: Encountered parse exception while reading /ZOWE/tmp/.zowe/workspace/app-server/serverConfig/server.json

bin/internal/run-zowe.sh
1000TurquoisePogs: Turns out I wrote this down but didn't implement zss as a distinct component. It is within app-server. Will remove this and add a note instead.

stevenhorsman:
I've tried testing the latest build and we are still hitting `2020-01-02 15:04:10.893 <ZWED:33620878> IZUSVR WARN (_zsf.json,jsonUtils.js:82) Encountered parse exception while reading /ZOWE/tmp/.zowe/workspace/app-server/serverConfig/server.json`

You can see the rest of the job log https://wash.zowe.org:8443/job/zowe-install-test/job/staging/607/consoleFull

1000TurquoisePogs:
Hi Steve. This appears to be a bug in the nodejs you are using: 8.16.1
Its similar to https://github.com/ibmruntimes/node/issues/142 but not quite the same...
That issue deals with what to do with untagged files. However, on my system, server.json IS tagged, as iso8850-1. However, it appears that the suggested untagged fix of setting __UNTAGGED_READ_MODE=**V6** somehow also effects **tagged** files, as setting __UNTAGGED_READ_MODE=**DEFAULT** in a test made it work again. However, that breaks **untagged** file behavior for us.

The fix appears to be to tag all of our files, and work was done on that but not merged into 1.8. I propose that for 1.8 we do not use 8.16.1

Joe-Winchester:
This was tested by https://wash.zowe.org:8443/job/zowe-install-test/job/staging/610/.

PR #921, opened 2019-12-12
Zip 869

### LINKED ISSUE #869: Enable extenders to add static definitions
**As a** Zowe extender,  
**I want** to easily add my static definition files to the zowe configuration  
**so that** my product shows up on the API Catalog and is API gateway enabled without me having to modify my code to include the dynamic registration.

*Details/notes*  
If a service is using the static definition then they either need to add it to the <instance-dir>/workspace/api-mediation/api-defs/ directory, or we need a way to register their config to find their copy of the static definition - using an EXTERNAL_APIML_STATIC_DEF_DIRECTORIES list in the instance.env which we append to the /<zowe-root-dir>/api-mediation/scripts/api-mediation-start-discovery.sh -Dapiml.discovery.staticApiDefinitionsDirectories property with a semi-colon separator

Note: we could create a utility script for this that would import the static def and then refresh the service using #746, but that is currently outside the scope of this work item

#### Acceptance Criteria

**Scenario 1:* Add static definition to zowe's directory*
*Given* An extender locates the instance directory of a zowe instance <instance> and a valid yaml static definition
*When* The yaml file is added to <instance-dir>/workspace/api-mediation/api-defs/ and Zowe is recycled, (or the static definitions are refresh with a POST to https://:/discovery/api/v1/staticApi with username:password eureka:password)
*Then* My extender service appears in the API catalog and I am able to route to it via the gateway

**Scenario 2:* Add external static definition*
*Given* An extender with a valid yaml static definition in <extender_path>/api-defs/ directory
*When* <extender_path>/api-defs/ is appended to EXTERNAL_APIML_STATIC_DEF_DIRECTORIES (semi-colon separated if it's not the only item) and Zowe is recycled
*Then* My extender service appears in the API catalog and I am able to route to it via the gateway

stevenhorsman:
Resolves #869 
Requires https://github.com/zowe/api-layer/pull/457


jackjia-ibm:
@stevenhorsman I didn't merge staging into your PR regarding to migrate artifactory:

- I saw a conflict on scripts/instance.template.env
- APIML team had approved https://github.com/zowe/api-layer/pull/456. Please make sure after you sync with staging branch (which will pick components from new artifactory), PR 456 had been merged, so new APIML components are also published to new artifactory.

stevenhorsman:
@jackjia-ibm - thanks Jack - I've merged form staging and resolved the conflict, so I think we should be good now

PR #968, opened 2020-01-10
Zip 845

stevenhorsman:
Move zowe-support out of zowe-configure.sh


stevenhorsman:
Tested by https://wash.zowe.org:8443/job/zowe-install-test/job/staging/645/

PR #1005, opened 2020-01-22
Rename node and java home

### LINKED ISSUE #1004: Update node and java home property names
We've previously had the java and node home property names called:
ZOWE_JAVA_HOME and ZOWE_NODE_HOME to avoid clashing with the 'base' version, but on the Cupids call the consensus seemed to be that this was confusing and we should revert back to the base.

In order to not break the components we'll need to shadow these properties in run-zowe.sh

stevenhorsman:
Resolves #1004


stevenhorsman:
Tested in https://wash.zowe.org:8443/job/zowe-install-test/job/staging/686/

PR #1042, opened 2020-02-03
Readme convert

stevenhorsman:
Create initial markdown version of SMP/E readme worddoc


stevenhorsman:
I've managed to replace the markdown code formating with HTML and depending on what interpretation of GFM they are using some of the tags work, so I've gone for a belt and braces approach. Take a look and see what you think

OnnoVdT:
I like it. Using bold does not stand out as much as highlight, but you gotta work with the tools you have.  If we can get the names of the .md and .docx sync'd up then I think we're good.

PR #1058, opened 2020-02-05
Zip 1022 - external component lifecycle

### LINKED ISSUE #1022: Add support for extender lifecycling with Zowe
**As a** Zowe extender,  
**I want** to lifecycle my extension with Zowe
**so that** my product starts up with Zowe and (optionally) integrates with the API gateway and zowe desktop.

*Details/notes*
At least 1 extender is far along enough with Zowe that they want to fully embrace the lifecycling and start up with the rest of the Zowe components.
In order to allow this we need to add a way for them to get hooked into the component lifecycle. I think the simplest way would be to add an EXTERNAL_COMPONENTS concatenation path to instance.env where they could add the full path to their component folder. This would then be appended to the validation, configuration and start lifecycle in run-zowe.sh
In return they need to conform to the component's (validate.sh (optional), configure.sh, start.sh) design in their component folder

**Question**: should we formalise the component lifecycling pattern via conformance? 

As noted in the parent epic we should update https://github.com/zowe/sample-node-api to work with this framework as a sample

#### Acceptance Criteria

**Scenario 1:* Add an iframe component*
*Given* An extender with a directory </usr/lpp/extender> containing start.sh which starts their ui component and configure.sh which sets up the iframe similar to https://github.com/zowe/explorer-jes/blob/master/scripts/explorer-jes-configure.sh
and a value EXTERNAL_COMPONENTS=</usr/lpp/extender> in their instance.env
*When* Zowe starts up
*Then* The extender's application is also started and the iframe plugin is available on the desktop

**Scenario 2:* Add api to the mediation layer*
*Given* An extender with a directory </usr/lpp/extender> containing start.sh which starts their API component and configure.sh which sets up the iframe similar to https://github.com/zowe/jobs/blob/master/jobs-zowe-server-package/src/main/resources/scripts/jobs-api-configure.sh
and a value EXTERNAL_COMPONENTS=</usr/lpp/extender> in their instance.env
*When* Zowe starts up
*Then* The extender's application is also started and the API is accessible through the gateway

stevenhorsman:
Fixes #1022


stevenhorsman:
Tested: https://wash.zowe.org:8443/job/zowe-install-test/job/staging/763/

PR #1060, opened 2020-02-05
New build of ZSS with fix

struga0258:


Joe-Winchester:
@struga0258 , @ifakhrutdinov , @1000TurquoisePogs - what is the issue that is fixed with this newer ZSS server ?  Does it need to be in 1.8 and create an rc3, or can it wait until we begin 1.9 (i.e. is it a stop ship defect or a feature that just missed the 1.8 release train).

PR #1062, opened 2020-02-06
Re-order zowe-install echos

### LINKED ISSUE #1061: echo at the end of zowe-install.sh has unhelpful step order
The 1-time only steps in the echo are at the end, but are best to be done first.

stevenhorsman:
Fixes #1061


stevenhorsman:
You don't have to create the keystore directory before creating the instance directory, but would need to circle back to the instance directory to update it with the instance, so I'd advice doing it first (as per the doc)

PR #1096, opened 2020-02-13
Remove debug/logs echoing to stdout

### LINKED ISSUE #1086: Clean up echo statements on zowe-install-proc.sh
**Describe the bug**
zowe-install-proc.sh writes information to the console that looks like debug info.

Screenshot 2020-02-11 at 15 15 27

**Expected behavior**
Do not have the <zowe-install and <zowe-copy statements in the console.  Write these to the log file but not to the end user.
Do not write the Check name and Copied statement that has the temporary file name (zowetemp).  
Do not write exit 0 lines to the console (two are in the screen shot above)

Just have the status of what happened and is useful to the end user.  
If the verbose echo are needed for running in the pipeline then maybe add a -verbose argument that brings them back, but for an end user running the script leave it tight and tidy

stevenhorsman:
Resolves #1086 

Reduces the noise that echos to stdout

### DIFF
--- a/scripts/utils/zowe-install-proc.sh
+++ b/scripts/utils/zowe-install-proc.sh
@@ -12,8 +12,8 @@
 # Function: Copy Zowe server PROC from datasetPrefx.SZWESAMP(ZWESVSTC) to JES concatenation
 # Needs ./zowe-copy-to-JES.sh
 script_exit(){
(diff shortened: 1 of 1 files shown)


stevenhorsman:
Tested: https://wash.zowe.org:8443/job/zowe-install-test/job/staging/801/

PR #1099, opened 2020-02-14
Move log into temp and relocate if r/o

### LINKED ISSUE #752: zowe-install.sh requires installation ZFS to be mounted R/W
Running 'zowe-install.sh -I' from a Zowe installation ZFS in R/O mode shows:

mkdir: FSUM6404 directory "/SYSTEM/var/zowe/lpp/zowe-1.4.0-install/temp_2019-09-11": EDC5141I Read-only file system

where the Zowe ZFS is mounted on /SYSTEM/var/zowe/lpp/. 
Usually, ZFSes containing software are to be mounted in R/O mode in order to prevent corruption as much as possible. We would like Zowe installation to adhere to rule that as well.  

In order to temporarily bypass this, I modified the setting of the TEMP_DIR variable value in the script to:
export TEMP_DIR=/tmp/zowe_"`date +%Y-%m-%d`"_$(whoami)  
and remounted the ZFS in R/O mode. That works fine for us.
The user id of the invoker is added as a precaution to the directory name in order not to block some other user from invoking the install script.

We are running Zowe SMP/E Alpha, v1.4.0

stevenhorsman:
Fixes #752 

Move the install_log file into the temp directory and if the install_directory is not writeable relocate our temp file to ${TMPDIR:-/tmp}

### DIFF
--- a/install/zowe-install.sh
+++ b/install/zowe-install.sh
@@ -56,11 +56,18 @@ separator() {
 }
 separator
 
-# Create a log file with the year and time.log in a log folder 
-# that scripts can echo to and can be written to by scripts to diagnose any install 
-# problems.  
+# Create a temp directory to be a working directory for sed replacements and logs, if install_dir is read-only then put it in ${TMPDIR}/'/tmp\'
+if [[ -w "${$INSTALL_DIR}" ]]
+then
+  export TEMP_DIR=$INSTALL_DIR/temp_"`date +%Y-%m-%d`"
+else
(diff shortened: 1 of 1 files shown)


stevenhorsman:
Test where install_dir couldn't be written to: https://wash.zowe.org:8443/job/zowe-install-test/job/staging/804
Test on normal flow: https://wash.zowe.org:8443/job/zowe-install-test/job/staging/805/

PR #1106, opened 2020-02-18
Change variable to LAUNCH_COMPONENT

### LINKED ISSUE #1101: Add LAUNCH_COMPONENT helper variable in run-zowe.sh loop
As a user, if i can get path of my custom component config.sh, start.sh and validate.sh script, it can save user from re-inventing different mechanisms to fetch this information like $0, $1, $BASH_SOURCE,  all of which might be more error prone.

Reliable path information can help user call their own custom scripts and navigate their own component structure if needed too.

As zowe, we already know path of external components from 
EXTERNAL_COMPONENT variable, which we loop over in run-zowe.sh, 
so lets set additional variable like LAUNCH_COMPONENT or any other appropriate name, 
which can be used in components lifecycle scripts to get path of executing script reliably

stevenhorsman:
Fixes #1101

### DIFF
--- a/bin/internal/run-zowe.sh
+++ b/bin/internal/run-zowe.sh
@@ -114,9 +114,10 @@ LAUNCH_COMPONENTS=${INTERNAL_COMPONENTS}",${EXTERNAL_COMPONENTS}"
 
 # Validate component properties if script exists
 ERRORS_FOUND=0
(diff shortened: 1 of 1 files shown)


stevenhorsman:
Tested: https://wash.zowe.org:8443/job/zowe-install-test/job/staging/817/

PR #1113, opened 2020-02-19
disable node check for zowe-setup-certificates.sh

js665999:
Signed-off-by: js665999 <js665999@broadcom.com>

- [ ] ~~Tests for the changes have been added (for bug fixes / features)~~
- [ ] ~~Necessary documentation (if appropriate) have been added / updated~~

#### PR type
- [x] Bugfix 
- [ ] Feature 

#### Relevant issues

Fixes 

- Disable node check in zowe-setup-certificates.sh

- [ ] Yes
- [x] No

### DIFF
--- a/bin/zowe-init.sh
+++ b/bin/zowe-init.sh
@@ -18,6 +18,7 @@
 # ZOWE_EXPLORER_HOST points to the current host name
 # ZOWE_IP_ADDRESS is the external IP address of the host ZOWE_EXPLORER_HOST where Zowe is installed 
 # NODE_HOME points to the node directory
+# NONODE if true skips node check
 
 # This script checks to see whether they are set, and if not tries to locate them, 
 # and if they can't be found prompt for them before setting them
@@ -167,7 +168,9 @@ then
 else
     echo "  NODE_HOME variable value="${NODE_HOME} >> $LOG_FILE
 fi
-promptNodeHome ${NODE_HOME}
+if ! ${NONODE}; then
+    promptNodeHome ${NODE_HOME}
+fi
 
 ###identify ping
 getPing_bin
--- a/bin/zowe-setup-certificates.sh
+++ b/bin/zowe-setup-certificates.sh
@@ -69,6 +69,7 @@ fi
 
 ZOWE_EXPLORER_HOST=${HOSTNAME}
 ZOWE_IP_ADDRESS=${IPADDRESS}
+NONODE=true #skip node check
 . ${ZOWE_ROOT_DIR}/bin/zowe-init.sh
 . ${ZOWE_ROOT_DIR}/scripts/utils/configure-java.sh
 


stevenhorsman:
I think using a getopts flag (like -v option in https://github.com/zowe/zowe-install-packaging/blob/master/bin/internal/run-zowe.sh) might be a cleaner way of doing it rather than relying on a shell argument being set?

js665999:
> I think using a getopts flag (like -v option in https://github.com/zowe/zowe-install-packaging/blob/master/bin/internal/run-zowe.sh) might be a cleaner way of doing it rather than relying on a shell argument being set?

I wanted to do that at first but the script is run sourced so it does not work well with arguments
`. ${ZOWE_ROOT_DIR}/bin/zowe-init.sh`

stevenhorsman:
See PR #1115 - I just tried it on my local mainframe and it worked. Do you have problems with that?

js665999:
It does not work for me, options are being ignored. I will try it from your build in case i did something differently

js665999:
I just confirmed that this does not work for me. Node is still being prompted. Maybe you have NODE_HOME variable already set so it does not ask for it?

stevenhorsman:
I modified my code to 
```
if [[ ! -z ${SKIP_NODE} ]]
then
    promptNodeHome ${NODE_HOME}
else
	echo "skipping node"
fi
``` 
and my output was
```
/u/stevenh/staging/runtime/bin/zowe-setup-certificates.sh -p /u/stevenh/staging/runtime/bin/zowe-setup-certificates.env
Loading default variables from /u/stevenh/staging/runtime/bin/zowe-setup-certificates.env file.
Loading /u/stevenh/staging/runtime/bin/zowe-setup-certificates.env file and overriding default variables.
skipping node
```
So mine is definitely skipping. Maybe we have a different shell environment? I've noticed that your script had `shift "$(($OPTIND-1))"` after the getopts block, so tried adding that to mine

js665999:
I'm still being prompted in your latest build. Should I try to find a working solution with options or could we keep my original change? You mentioned in scrum update that zowe-init.sh will be refactored anyway in the near future.

stevenhorsman:
I'm mixed - I don't want to cause un-due work, but I'd really like to understand what is different in your environment in case it's the same in other customers and we have other scripts that don't work.
The re-work that I was thinking would use getopts to toggle whether node is checked or not anyway, so I don't think that will help

stevenhorsman:
Can you add some debug/echos to work out what specifically is going please. ie print all the args at the start, check if it is the getopts s) that is never hit, or my if logic that is broken?

js665999:
Switching if/else made it work for me. Not really sure how that could behave differently on our machines. 

from
`if [[ ! -z ${SKIP_NODE} ]]` 
to
`if [[ -z ${SKIP_NODE} ]]` 

Can you test this on your end?

stevenhorsman:
Ok - it sounds to me like I've got some errors in my code and it just happened to be working for me yesterday (who says 2 wrongs don't make a right!). I'll dig in deeper and debug mine and get back to you

stevenhorsman:
Ok - many apologies I completely messed it up! I did some investigation and found that when calling getopts from source and a scripts that has getopts'd already (like zowe-setup-certificate.sh) the OPTIND parameters has been updated, so needs resetting. After I fixed that and updated the if logic it seems to be working now for me and hopefully you too.

js665999:
Yes this works for me too. I guess your branch could be PRd to staging directly. Thanks for the help and sorry to get you mixed up in this small issue.

stevenhorsman:
There is no need for you to apologise - I'm the one that muddied the water with incorrect code! Glad we can get the update in though :)

PR #1145, opened 2020-02-25
Bugfix/read instance fix

1000TurquoisePogs:
https://github.com/zowe/zowe-install-packaging/pull/1144 but for staging instead of RC


1000TurquoisePogs:
No wait this isnt quite right, let me fix it.

PR #1149, opened 2020-02-26
WIP: zip 173

### LINKED ISSUE #173: Put versions in zowe-install.yaml file
We have a Jenkins script that handles installation of Zowe.  Since we need to customize ports that Zowe uses, we take a copy of zowe-install.yaml file and update the ports. Because we are taking a copy of the yaml file, we run a risk of having yaml file being outdated (when Zowe driver updates the yaml file). If the zowe-install.yaml file version can be put in the header of the yaml file, our script can just easily compare the versions to know the customized yaml file is outdated or not.

stevenhorsman:
Add version to instance.env
Fixes #173


stevenhorsman:
After discussing with Joe, decided this wasn't a good change

PR #1152, opened 2020-02-26
Zowe convenience workflow

abastyr:
#### PR type
- Feature

#### Relevant issues
N/A

- Workflow has been added that can be used to install the Zowe runtime from the convenience build.

No.

Allows easier installation of the Zowe runtime from the convenience build.

No.


files/workflows/ZWEWRF04.xml
```diff
@@ -0,0 +1,177 @@
...
+        <canMarkAsFailed>false</canMarkAsFailed>
+        <template>
+            <inlineTemplate substitution="true">mkdir -p "${instance-zowe_install_path}"</inlineTemplate>
+            <submitAs maxRc="0">shell-JCL</submitAs>
+            <maxLrecl>1024</maxLrecl>
+            <!-- zosmfOutput -->
+        </template>
+    </step>
+    <step name="mount_zfs" optional="false">
+        <title>Mount ZFS</title>
```
OnnoVdT: In comments, show sample update to BPXPRMxx to mount on next IPL
```
MOUNT FILESYSTEM('#dsn')                  
 MOUNTPOINT('#mountpoint')                  
 MODE(RDWR)        /* can be MODE(READ) */           
 TYPE(ZFS) PARM('AGGRGROW') /* zFS, with extents */  
```

files/workflows/ZWEWRF04.xml
```diff
@@ -0,0 +1,177 @@
...
+        <category>Build Location</category>
+        <string valueMustBeChoice="false" multiLine="false">
+            <regularExpression>~?(\/[^\/]+)+</regularExpression>
+            <errorMessage>Please specify proper full USS path ending without a slash.</errorMessage>
+            <!-- Specify choices here -->
+            <default>/u/users/ibmusers/zowe-build/zowe-1.8.0</default>
+        </string>
+    </variable>
+    <variable name="zowe_install_path" scope="instance" visibility="public">
+        <label>USS Install Path</label>
```
OnnoVdT: add optional variable to define TMPDIR. This is used by zowe-install.sh when $INSTALL_DIR is read-only

files/workflows/ZWEWRF04.xml
```diff
@@ -0,0 +1,177 @@
...
+        <!-- condition -->
+        <instructions substitution="true">Run this step to run the Zowe install script.</instructions>
+        <weight>1</weight>
+        <skills>None</skills>
+        <autoEnable>true</autoEnable>
+        <canMarkAsFailed>false</canMarkAsFailed>
+        <template>
+            <inlineTemplate substitution="true">set -x
+cd "${instance-zowe_build}/install"
+sh ./zowe-install.sh -i ${instance-zowe_install_path} -h ${instance-zowe_dsn_prefix}</inlineTemplate>
```
OnnoVdT: sh TMPDIR=${"${our_tmpdir}":-/tmp}; .zowe-install.sh ...

files/workflows/ZWEWRF04.properties
```diff
@@ -0,0 +1,33 @@
+# zowe_build
+# Label: Zowe Build
+# Abstract: Path to the extracted convenience build
+# Category: Build Location
+# Description:
+#  Path to the folder from the extracted convenience build PAX file.
+zowe_build=/u/users/ibmusers/zowe-build/zowe-1.8.0
```
OnnoVdT: the path is more likely going to be `~/zowe-1.9.0`
`/u/users/ibmusers/zowe-build` sounds like something very specific to your environment

files/workflows/ZWEWRF04.xml
abastyr: I have made just a simple TMPDIR assign, because inside the `zowe-install.sh` script there already is:

```sh
export TEMP_DIR=${TMPDIR:-/tmp}/zowe_"`date +%Y-%m-%d`"
```

files/workflows/ZWEWRF04.xml
```diff
@@ -168,6 +189,8 @@
         <template>
             <inlineTemplate substitution="true">set -x
 cd "${instance-zowe_build}/install"
+TMPDIR="$!{instance-zowe_user_temp}"
+export TMPDIR="$TMPDIR"
```
OnnoVdT: zowe-user-temp is set as optional, so provide a default here
TMPDIR="$!{instance-zowe_user_temp}"
export TMPDIR="${TMPDIR:-/tmp}"

OnnoVdT:
Just thought of another request, which me might move to a follow-up issue. We have the Zowe version hardcoded in the script, which means it must be updated on each RC build. We should update the build process to insert to correct version ($ZOWE_VERSION) as part of the build.

files/workflows/ZWEWRF04.xml
abastyr: There is already default in `zowe-install.sh` on line 55:

```sh
export TEMP_DIR=${TMPDIR:-/tmp}/zowe_"`date +%Y-%m-%d`"
```

Putting a default on the `TMPDIR` would overwrite the aforementioned default.

PR #1159, opened 2020-02-27
WIP: Zip 1154

### LINKED ISSUE #1154: Add -l parameter to install/config scripts to overwrite the log directory
We currently have a few different scripts that each log during running:

install, certificate setup, proclib copying...
these each write logs in a slightly different place (or different files in the home directory), making management annoying.

For the pipeline (at least) adding a -l (lowercase L) parameter specifying the log directory for each step might be the best solution.

Needs to update:
- [x] zowe-install.sh - Note this file is moved into the runtime and deleted at the end
- ~zowe-configure-instance.sh~ this defaults to the instance directory so we can access these in service
- [x] zowe-setup-certificates.sh

Blocked by #1156 
- ~[ ] zowe-copy-to-JES.sh~
- ~[ ] zowe-install-proc.sh~
- ~[ ] zowe-install-xmem.sh~

**Question**
Given that the current behaviour of zowe-install.sh is, on completion to move the log into the runtime install (so that we can pick it up there for service purposes), how do we want to handle -l?
Do we:
1. Create the temp log location in -l then move to runtime
2. Create the log location in -l then copy to runtime, so it is in both
3. Create the log location in -l and don't move at the end (this might cause issues with the service process)

stevenhorsman:
### PR type
- [X] Feature

#### Relevant issues
Fixes #1154 

- Update zowe-install and zowe-setup-certificates.sh scripts to add new -l parameter to overwrite the log directory

- [ ] Yes
- [X] No

-l (lowercase L) option added to install scripts to allow installer to overwrite default log directory location

Doc issue/PR number: https://github.com/zowe/docs-site/issues/1054


stevenhorsman:
Tested in https://wash.zowe.org:8443/job/zowe-install-test/job/staging/908/

PR #1170, opened 2020-03-02
add wf to generate certificates

uzuko01:
**PR type**
Feature
**Relevant issues**
N/A

**Changes proposed in this PR**
Workflow has been added that can be used to create zowe certificates with custom values.
**Does this PR introduce a breaking change?**
No.

**Does this PR do something the person installing Zowe should know about?**
Allows easier creation of certificates.

**Is there a related doc issue or Pull Request?**
No.


files/workflows/ZWEWRF05.xml
```diff
@@ -0,0 +1,215 @@
...
+ZOSMF_CERTIFICATE='${ZOSMF_CERTIFICATE}'
+KEYSTORE_PASSWORD='${KEYSTORE_PASSWORD}'
+KEYSTORE_DIRECTORY='${KEYSTORE_DIRECTORY}'
+KEYSTORE_ALIAS='${KEYSTORE_ALIAS}'
+ZOWE_USER_ID='${ZOWE_USER_ID}'
+ZOWE_GROUP_ID='${ZOWE_GROUP_ID}'
+
+sed -e "s#HOSTNAME=#HOSTNAME=$HOSTNAME#g" -e "s#IPADDRESS=#IPADDRESS=$IPADDRESS#g" -e "s#VERIFY_CERTIFICATES=true#VERIFY_CERTIFICATES=$VERIFY_CERTIFICATES#g" -e "s#EXTERNAL_CERTIFICATE=#EXTERNAL_CERTIFICATE=$EXTERNAL_CERTIFICATE#g" -e "s#EXTERNAL_CERTIFICATE_ALIAS=#EXTERNAL_CERTIFICATE_ALIAS=$EXTERNAL_CERTIFICATE_ALIAS#g" -e "s#EXTERNAL_CERTIFICATE_AUTHORITIES=#EXTERNAL_CERTIFICATE_AUTHORITIES=$EXTERNAL_CERTIFICATE_AUTHORITIES#g" -e "s#ZOSMF_CERTIFICATE=#ZOSMF_CERTIFICATE=$ZOSMF_CERTIFICATE#g" -e "s#KEYSTORE_PASSWORD=password#KEYSTORE_PASSWORD=$KEYSTORE_PASSWORD#g" -e "s#KEYSTORE_DIRECTORY=/global/zowe/keystore#KEYSTORE_DIRECTORY=$KEYSTORE_DIRECTORY#g" -e "s#KEYSTORE_ALIAS=localhost#KEYSTORE_ALIAS=$KEYSTORE_ALIAS#g" -e "s#ZOWE_USER_ID=ZWESVUSR#ZOWE_USER_ID=$ZOWE_USER_ID#g" -e "s#ZOWE_GROUP_ID=ZWEADMIN#ZOWE_GROUP_ID=$ZOWE_GROUP_ID#g" ${zowe_install_path}/bin/zowe-setup-certificates.env > ${USSDIR}/zowe-setup-certificates.env
```
OnnoVdT: I'm not sure if xml has issues with it, but I'd prefer to see this sed command spread over multiple lines, with one substitution per line (followed be \ to indicate continuation) to simplify readability and maintenance.

uzuko01: Hi @OnnoVdT ,
I'd prefer that, too, but it is not working on mainframe. The only way I could make this work is like that. :\

files/workflows/ZWEWRF05.xml
```diff
@@ -0,0 +1,215 @@
...
+        <skills>z/OS System Programmer</skills>
+        <autoEnable>true</autoEnable>
+        <canMarkAsFailed>true</canMarkAsFailed>
+        <template>
+            <inlineTemplate substitution="true">
+if test ! -d ${USSDIR}; then mkdir -p -m 755 ${USSDIR}; fi
+cp ${zowe_install_path}/bin/zowe-setup-certificates.env ${USSDIR}/zowe-setup-certificates.env
+chmod 755 ${USSDIR}/zowe-setup-certificates.env
+
+HOSTNAME='${HOSTNAME}'
```
OnnoVdT: Since we are not generating the workflow and zowe-setup-certificates.env from a single source, we should add a comment to zowe-setup-certificates.env indicating that development must keep it in sync with this workflow, and vice versa. Other option of course is to rework the setup in Github so we can single source the two files that end up being shipped.

uzuko01: @OnnoVdT I missed that comment on top, sorry.
I don't think anyone will want to change the workflow without editing the zowe-setup-certificates.env file, since the wf is covering the variables in the zowe-setup-certificates.env. Maybe for now we can just add instructions in the zowe-setup-certificates.env file and in the future think of a way to generate them both from a single source?

OnnoVdT: I agree, hence that being my first proposition. A note in the .env should be sufficient as any changes to the process will likely be tested without the workflow in play, and thus the comment will be noticed.

PR #1187, opened 2020-03-10
Zowe instance workflow

abastyr:
Simple z/OSMF workflow that allows the user to create, configure and start a new Zowe instance.

#### PR type
- [ ] Bugfix 
- [x] Feature 

#### Relevant issues

No relevant issue

This PR contains the z/OSMF workflow to be used to create, configure and start the new Zowe instance.

- [ ] Yes
- [x] No

---

---

---

---
_multi-line description_

No


workflows/ZWEWRF0n.xml
```diff
@@ -302,4 +302,82 @@ sh ./zowe-configure-instance.sh -c "${instance-zowe_instance_path}"</inlineTempl
...
+        <instructions substitution="false">Run this step to change the instance configuration.</instructions>
+        <weight>1</weight>
+        <skills>None</skills>
+        <autoEnable>true</autoEnable>
+        <canMarkAsFailed>false</canMarkAsFailed>
+        <template>
+            <inlineTemplate substitution="true">//REPLFILE EXEC PGM=IEBGENER
+//SYSPRINT DD SYSOUT=*
+//SYSIN    DD DUMMY
+//SYSUT2   DD PATH='${instance-zowe_instance_path}/instance.env',
```
OnnoVdT: By coding the path directly in the DD the length is limited to 36 characters (total line must be 71 or less). 
If we want we can allow a longer path (57 chars) by using a SET statement
// SET INST='${instance-zowe_instance_path}'
// SET QUOTE=''''
...
//SYSUT2   DD PATH=&QUOTE.&INST/instance.env',

Either way, we should document the maximum path length where customers can specify the path. This likely also applies to other paths like zowe_install_path.

workflows/ZWEWRF0n.xml
```diff
@@ -433,5 +433,37 @@ exit $?</inlineTemplate>
...
+        <!--variableValues-->
+        <!-- condition -->
+        <instructions substitution="true">Run this step to start the Zowe instance.
+        
+If you want to start the instance manually (e.g. after IPL), you can do it with the following command:
+&lt;code&gt;sh "${instance-zowe_instance_path}/bin/zowe-start.sh"&lt;/code&gt;
+
+You can also start the instance using the following MVS command:
+
+&lt;code&gt;/S ZWESVSTC INSTANCE='${instance-zowe_instance_path}',JOBNAME=${instance-ZOWE_PREFIX}${instance-ZOWE_INSTANCE}SV&lt;/code&gt;
```
OnnoVdT: the leading / in the command is specific for SDSF to route a command to the console. It should not be part of the sample, but we can add a comment saying 
Note: Operator commands can be issued from the SDSF command line if you are authorized to do so. In SDSF, operator commands are prefixed with the forward slash (/) character.

workflows/ZWEWRF0n.xml
```diff
@@ -433,5 +433,37 @@ exit $?</inlineTemplate>
...
+        <prereqStep name="create_instance"/>
+        <prereqStep name="change_instance_configuration"/>
+        <!--variableValues-->
+        <!-- condition -->
+        <instructions substitution="true">Run this step to start the Zowe instance.
+        
+If you want to start the instance manually (e.g. after IPL), you can do it with the following command:
+&lt;code&gt;sh "${instance-zowe_instance_path}/bin/zowe-start.sh"&lt;/code&gt;
+
+You can also start the instance using the following MVS command:
```
OnnoVdT: ... the following operator command

workflows/ZWEWRF0n.xml
abastyr: I have added the SDSF related command and removed slash from the command example

workflows/ZWEWRF0n.xml
abastyr: Fixed

workflows/ZWEWRF0n.xml
abastyr: z/OSMF is a little bit different. The max record length is set to 1024, which means each individual line of the job submitted from that step can be up to 1024 characters. It is a little specialty that z/OSMF has, otherwise, variables would require a lot of additional checks.

OnnoVdT: Didn't know that. Thanks for the education.

PR #1189, opened 2020-03-11
Smpe vtl fix

abastyr:

#### PR type
- [x] Bugfix 
- [ ] Feature 

#### Relevant issues

No related issue

- Removed pointless `sysmodsCheck` variable
- Fixed some conditions for `ibmTemplate` to make sure conditions work properly even if the variable default is not initialized during the workflow execution
- Added extra conditional steps for SMP/E Apply and Accept that will be executed if checks work well.

- [ ] Yes
- [x] No

---

---

---

---
_multi-line description_

No

Doc issue/PR number:


smpe/ZOSMF/vtls/ZWE7APLY.vtl
```diff
@@ -55,7 +55,7 @@
 //*
 //********************************************************************
 //*
-//APPLY    EXEC PGM=GIMSMP,REGION=0M
+//APPLY    EXEC PGM=GIMSMP,REGION=0M,COND=(4,LT)
```
OnnoVdT: as FYI, the point of the pointless condition is simplifying merging all SMPE jobs into a single one and halt on error.

OnnoVdT:
You changed the FMID APPLY/ACCEPT vtls to do check and then conditionally do it again without check. However, the `#if($ibmTemplate == 'NO' || !$!ibmTemplate)` check is not present for the PTF APPLY/ACCEPT vtls.

smpe/ZOSMF/vtls/ZWE7APLY.vtl
abastyr: I figured it was like that, however, the condition is "pointless" in the workflow :)

abastyr:
I made the change for the workflow, so the APPLY/ACCEPT will always run with and without the check. We do not use the PTF vtls in the workflow :)

Issue #1268, opened 2020-04-16
Add log parameter to install/configure scripts


stevenhorsman:
Update the following scripts to implement -l parameter with the logic specified in #956:
- [ ] zowe-install.sh - Note this file is moved into the runtime and deleted at the end
- ~zowe-configure-instance.sh~ this defaults to the instance directory so we can access these in service
- [ ] zowe-setup-certificates.sh
- [ ] zowe-copy-to-JES.sh
- [ ] zowe-install-proc.sh
- [ ] zowe-install-xmem.sh
```
If a -l parameter is supplied to the script:
  LOG_DIR = -l parm
else:
  if /global/zowe/logs is writable:
     LOG_DIR = /global/zowe/logs
  else:
       LOG_DIR = ~/zowe/logs
if LOG_DIR not writable (ie -l parameter, or ~/zowe/logs doesn't have permission):
   echo an error explaining and directing the user to supply -l to a writable directory
   exit with non-zero code
```
- In all the scripts we should ensure that we echo out the log file created so that users can find it and any pipeline automation can as well

**Scenario 1:* Valid -l parameter supplied
*Given* a writable directory <log_dir>
*When* I execute any of zowe-install.sh/zowe-setup-certificates.sh/zowe-copy-to-jes.sh/zowe-install-proc.sh/zowe-install-xmem.sh with -l <log_dir>
*Then* the script should create it's date-stamped logs in <log-dir>/
    *And* the script should echo out the name of the log file created

**Scenario 2:* invalid -l parameter supplied
*Given* a non-writable directory <log_dir>
*When* I execute any of zowe-install.sh/zowe-setup-certificates.sh/zowe-copy-to-jes.sh/zowe-install-proc.sh/zowe-install-xmem.sh with -l <log_dir>
*Then* the script errors explaining that <log_dir> is not writable and no actions are taken

**Scenario 3:* /global/zowe/logs is writable
*Given* /global/zowe/logs is writable
*When* I execute any of zowe-install.sh/zowe-setup-certificates.sh/zowe-copy-to-jes.sh/zowe-install-proc.sh/zowe-install-xmem.sh with no -l parameter
*Then* the script should create it's date-stamped logs in /global/zowe/logs
    *And* the script should echo out the name of the log file created

**Scenario 4:* /global/zowe/logs is not writable, ~/zowe/logs is
*Given* /global/zowe/logs is not existing/writable, but ~/zowe/logs is
[...]

PR #1272, opened 2020-04-17
Add new log directory setting

stevenhorsman:

#### PR type
- [X] Feature

#### Relevant issues
Implements #1268

Create a unified place for all Zowe install/configure logs

Various zowe install and configuration scripts now have the option to add `-l <log_dir>` to give a specific log directory output

Doc issue: https://github.com/zowe/docs-site/issues/1133


stevenhorsman:
@jackjia-ibm - are you able to take a lot at this PR please?
PTF test: https://wash.zowe.org:8443/job/zowe-install-test/view/change-requests/job/PR-1272/20/
Convenience test: https://wash.zowe.org:8443/job/zowe-install-test/view/change-requests/job/PR-1272/19/

jackjia-ibm:
Thanks @stevenhorsman for the fix. The build and test all look good.

Extra thoughts,

- I saw many similar code of `source setup-log-dir.sh` to `echo "Log file created: ${LOG_FILE}"`, not sure if it makes better we define a function `ensure_log_file` in `setup-log-dir.sh` and call it like `ensure_log_file "zowe-install"` to create the log file target directory.
- Will anyone assign values for `zowe_root_dir`, `zowe_instance_dir`, `zowe_keystore_dir` and `zowe_logs_dir` with space in directory name? For example, `zowe_logs_dir: /u/myuser/zowe 1.10.0/logs`. This should be a really rare use case.

Anyway those don't really big issues since everything work as expected already.

stevenhorsman:
Thanks for the suggestion JacK. I've updated based on that (and fixed the addcontext linting problem).
Builds https://wash.zowe.org:8443/job/zowe-install-test/view/change-requests/job/PR-1272/23/ and https://wash.zowe.org:8443/job/zowe-install-test/view/change-requests/job/PR-1272/24/
WRT point 2 - I think it's worth us testing directories with a space - probably manually to start with and can worry about whether we automate it later?

jackjia-ibm:
Thanks Steve for the fix, including the test case (it was a quick fix for a security concern). Yes I agree point 2 is a little off topic and is not the purpose of this PR. We can test it out later.

Issue #1282, opened 2020-04-21
Quick test to check if extra instance properties are preserved during upgrade


stevenhorsman:
As extenders add their own properties to instance.env - we want to ensure that these aren't wiped when an instance.

The test will take the form:
- During fmid configure add a new property to instance.env
- Do the ptf install and update the instance-dir
- Check the new property is still in instance.env

PR #1283, opened 2020-04-21
Users/stevenh/test instance property preserved

stevenhorsman:

#### PR type
- [X] Other - new test case

#### Relevant issues
Implements #1282 

- Quick test to check if extra instance properties are preserved during upgrade


playbooks/roles/configfmid/defaults/main.yml
```diff
@@ -68,3 +68,6 @@ zowe_external_domain_name:
 # optional, no default value
 # required for zD&T (external IP is different from internal IP)
 zowe_external_ip_address:
+
+# property added to instance.env for extender test
+zowe_extender_test_property: EXAMPLE_EXTENDER_PROPERTY=TEST123
```
jackjia-ibm, line 73: I'm thinking this makes FMID installation a little different comparing to convenience build. So if it's better to add some comments for this variable to not confuse people, like "This variable is for PTF installation verification purpose, not used by any extensions. You are safe to ignore." blah blah.

stevenhorsman: Thanks for the suggestions, added the comment.

PR #1293, opened 2020-04-23
Add sso auth plugin by default

1000TurquoisePogs:
sso-auth plugin is a combination of apiml-auth and zss-auth, making them redundant. apiml also handles zosmf auth due to their sso integration. This PR also includes the changes from the last release, which was done in the rc branch and staging had not had those changes back-ported, where the zosmf auth & proxy were removed since apiml-auth had made them redundant. So, this is an auth plugin consolidation PR.

### DIFF
--- a/files/zlux/config/allowedPlugins.json
+++ b/files/zlux/config/allowedPlugins.json
@@ -37,6 +37,10 @@
       "versions": ["*"]
     },
     {
+      "identifier": "org.zowe.zlux.auth.safsso",
+      "versions": ["*"]
+    },
+    {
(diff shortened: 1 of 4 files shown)


1000TurquoisePogs:
Note: this does not delete those plugins referenced, it just changes what is used by default.

1000TurquoisePogs:
This should not go into the release for tomorrow, we need more time to make sure the explorers & apiml are working well with this change. We'll do it for 1.12

1000TurquoisePogs:
Put this code into https://github.com/zowe/zowe-install-packaging/pull/1347

PR #1296, opened 2020-04-23
Update manifest with zss zlux 1.11 files

1000TurquoisePogs:


jackjia-ibm:
The sourceDependencies need to be `rc` for now so we can start source scan. The `v1.11.0` tags on zlux are not created yet. Will merge and fix them later. @1000TurquoisePogs @struga0258

Issue #801, opened 2019-10-13
zowe-install.sh -f <logfile>


OnnoVdT:
*Customer content*

**Is your feature request related to a problem? Please describe.**
zowe-install.sh currently always decides how to name the log file which by default is LOG_FILE=$INSTALL_DIR/log/$(date +%Y-%m-%d-%H-%M-%S).log. This is troublesome for automated install to do the SMP/E packaging of Zowe, as now the install-log of the SMP/E-specific parts cannot be added, nor can the log be saved as part of the build output for future reference.

**Describe the solution you'd like**
I propose that zowe-install.sh accepts a new startup argument, -f , that overrides the default LOG_FILE definition. This way, automated install can simply specify the desired location as part of the zowe-install.sh invocation.

**Describe alternatives you've considered**
none

**Additional context**
none

---
*Engineering team info (eg https://github.com/zowe/explorer-jes/issues/4)*

**As a** [type of user],  
**I want** [some goal]  
**so that** [some reason].

*Details/notes*  
[Detail - implementation notes]

#### Acceptance Criteria

**Scenario 1:* [Title]  
*Given* [context]  
   *And* [some more context]...  
*When* [event]  
*Then* [outcome]  
    *And* [another outcome]...

OnnoVdT:
related to issue https://github.com/zowe/zowe-install-packaging/issues/752

stevenhorsman:
Covered by #1154 which adds log directory parameter for all install/config scripts

OnnoVdT:
Not exactly covered, as #1154 handles the log directory, not the log file name. I'll alter the smpe script to do a rename to a known name (and location).

stevenhorsman:
Yep - sorry Onno, I should have been more specific - what I meant to say is that with -l for the directory being rolled out across all our scripts I'm not sure that -f specifying the file as well on zowe-install.sh makes much sense. It is worth noting that the zowe-install.sh actually has always moved it's log into the runtime installed environment at the end of the script, so it is accessible from the root directory, so an option is to grab it from there as needed?

OnnoVdT:
Copying it to runtime is what we do NOT want for SMP/E, as it our install log, not SMP/E's. The smpe script trashes any log dirs it finds after install in the target staging area. The build logs are kept in a separate location that also goes to artifactory in case we need it for problem solving.
with -f smpe can tell zowe-install.sh which file to write logs too and be done with it. Even if the install fails we have the log up to that point tucked away for review.

PR #1300, opened 2020-04-24
Zip 1280 - Update scripts to support -f to go direct to log

### LINKED ISSUE #1280: zowe-copy-to-jes.sh creates a separate log
If you run zowe-install-proc.sh, or zowe-install-xmem.sh then it will log some stuff, then call zowe-copy-to-jes which creates a separate log file. This can make it tricky to stitch them together and debug each step.

stevenhorsman:

#### PR type
- [X] Bugfix

#### Relevant issues
Fixes #1280 #801 

- Update logging so that subscripts called append/embed into the main logging script


jackjia-ibm:
@stevenhorsman I later realize the variable `zowe_logs_dir` defined in playbook could be misleading. The user may think it's all zowe logs, including runtime logs, but actually it's just install/confgure logs. So maybe `zowe_install_logs_dir` is more accurate?

stevenhorsman:
Tested in convenience: https://wash.zowe.org:8443/job/zowe-install-test/view/change-requests/job/PR-1300/3/ and ptf: https://wash.zowe.org:8443/job/zowe-install-test/view/change-requests/job/PR-1300/4/

install/zowe-install.sh
```diff
@@ -11,16 +11,17 @@
...
 fi
 
-while getopts "h:i:l:d" opt; do
+while getopts "f:h:i:l:d" opt; do
   case $opt in
     d) # enable debug mode
       # future use, accept parm to stabilize SMPE packaging
       #debug="-d"
       ;;
+    f) LOG_FILE=$OPTARG;;
```
jackjia-ibm, line 24: I found not like other configuration scripts, `zowe-install-proc.sh` etc, this script is added a `-f` option. I know why we need `-f` for `zowe-copy-to-JES.sh`, but didn't quite get the purpose of here. But I can see it's not harmful at all.

stevenhorsman: Hey Jack - this is to meet the requirement #801 that Onno had during the smpe packaging, so I've not doc'd it as I consider it internal use. Apologies for the lack of clarity, I've add a comment to explain

PR #1323, opened 2020-04-29
default bogus hostname to IP

### LINKED ISSUE #1279: When ZOWE_HOST_NAME cannot be resolved default to ZOWE_IP_ADDRESS in instance.env
**Describe the bug**
On some systems the ZOWE_HOST_NAME cannot be resolved externally.  It is used in the configure.sh scripts for the desktop exolorers, and this causes 404 errors

**Expected behavior**
When the ZOWE_HOST_NAME cannot be resolved if none is specified as an override then default to the ZOWE_IP_ADDRESS so that the instance.env file ZOWE_HOST_NAME is the ZOWE_IP_ADDRESS

This logic will be zowe-init.sh.  

**Screenshots**
When parsing the results of the dig command on the system in error is shown below

image (12)

The result of running zowe-init.sh on the system in error is shown belos
image (13)

**Additional context**
This problem will resolve https://github.com/zowe/explorer-jes/pull/169

John-A-Davies:

#### PR type

- [x] Bugfix 
- [ ] Feature 

#### Relevant issues

Fixes #1279 

- Set the ZOWE_EXPLORER_HOST to be the ZOWE_IP_ADDRESS if dig and ping don't confirm that they match.

- [ ] Yes
- [x] No

If the install script cannot determine the external IP address of the system being configured, the installer will be asked to enter it as the ZOWE_IP_ADDRESS.  Be aware that this IP address will be used for the ZOWE_EXPLORER_HOST also, if the script cannot resolve the given ZOWE_EXPLORER_HOST with the ZOWE_IP_ADDRESS.  
  

---
Configuration after install.  
---

---

---
_multi-line description_

Doc issue/PR number: 

#### Other information

The commands `dig` and `ping` were used before, but a new case was found (#1279), where `dig` failed to resolve *any* external address, because the system was not visible externally.


John-A-Davies:
Tested on ukzowe3 and winmvs3b, as well as on Nayer's system

PR #1330, opened 2020-05-01
Run initInstance on instance creation, to prevent ownership issues later

1000TurquoisePogs:
- [not yet] Necessary documentation (if appropriate) have been added / updated

#### PR type
- [x] Bugfix 
- [ ] Feature 

I've been seeing issues since 1.8 where people need to edit the app server's server.json, which is writable only by the owner, but cannot because the owner is an stc because it gets initialized on first server run, and is run by an stc. This makes some edge-case configuration that existed prior to 1.8 impossible. 
However, if the files were initialized at instance creation in addition to sanity-checked at startup as we do now, then ownership would be on whoever ran the create command, not what runs the server. This could cause issues for those who do not have the server runner and server installer in the same group, but otherwise should cause no issues. Is that already guaranteed, or must we doc it and consider the consequences?

- [ ] Yes
- [ ] No
- [x] Maybe

Output before:
```
> ./zowe-configure-instance.sh -c ~/zowe/inst/1
Creating zowe instance in /u/me/zowe/inst/1/
Error: ping myhostname.here command failed to find hostname
    Please enter the ZOWE_EXPLORER_HOST of this system
myhostname.here
Error: ping myhostname.here command failed to find IP
error : ZOWE_EXPLORER_HOST or ZOWE_IP_ADDRESS is an empty string
    Please enter the ZOWE_IP_ADDRESS of this system
192.168.1.100
Info: User entered ZOWE_IP_ADDRESS=192.168.1.100
warning : ping myhostname.here did not match stated IP address 192.168.1.100
Configure instance completed. Please now review the properties in /u/me/zowe/inst/1//instance.env to check they are correct.
To start Zowe run the script /u/me/zowe/inst/1//bin/zowe-start.sh
   (or in SDSF directly issue the command /S ZWESVSTC,INSTANCE='/u/me/zowe/inst/1/')
To stop Zowe run the script /u/me/zowe/inst/1//bin/zowe-stop.sh
  (or in SDSF directly the command /C ZWESVSTC)
```

Output after:
```
> ./zowe-configure-instance.sh -c ~/zowe/inst/1
Creating zowe instance in /u/me/zowe/inst/1/
Error: ping myhostname.here command failed to find hostname
    Please enter the ZOWE_EXPLORER_HOST of this system
myhostname.here
Error: ping myhostname.here command failed to find IP
error : ZOWE_EXPLORER_HOST or ZOWE_IP_ADDRESS is an empty string
    Please enter the ZOWE_IP_ADDRESS of this system
192.168.1.100
Info: User entered ZOWE_IP_ADDRESS=192.168.1.100
warning : ping myhostname.here did not match stated IP address 192.168.1.100
ZWED5010I - Creating new config= {
  node: {
    https: {
      ipAddresses: [Array],
      port: 8544,
      keys: [Array],
      certificates: [Array],
      certificateAuthorities: [Array]
    },
    mediationLayer: { server: [Object], enabled: false },
    childProcesses: [ [Object] ]
  },
  agent: { host: 'localhost', http: { ipAddresses: [Array], port: 8542 } },
  privilegedServerName: 'ZWESIS_STD',
  productDir: '/u/me/zowe/root/init-me/components/app-server/share/zlux-app-server/defaults',
  dataserviceAuthentication: {
    defaultAuthentication: 'zss',
    rbac: false,
    implementationDefaults: { zss: [Object] }
  },
  siteDir: '/u/me/zowe/inst/1/workspace/app-server/site',
  instanceDir: '/u/me/zowe/inst/1/workspace/app-server',
  groupsDir: '/u/me/zowe/inst/1/workspace/app-server/groups',
  usersDir: '/u/me/zowe/inst/1/workspace/app-server/users',
  pluginsDir: '/u/me/zowe/inst/1/workspace/app-server/plugins'
}
ZWED5011I - Generating default plugin references
ZWED5012I - Copying default plugin preferences into instance
Configure instance completed. Please now review the properties in /u/me/zowe/inst/1//instance.env to check they are correct.
To start Zowe run the script /u/me/zowe/inst/1//bin/zowe-start.sh
   (or in SDSF directly issue the command /S ZWESVSTC,INSTANCE='/u/me/zowe/inst/1/')
To stop Zowe run the script /u/me/zowe/inst/1//bin/zowe-stop.sh
  (or in SDSF directly the command /C ZWESVSTC)
```


Joe-Winchester:
@1000TurquoisePogs - what is the failure that caused this PR ?  It looks as though you are checking for node 12 and doing some exports.  The generic script `zowe-set-env.sh` should set the vars already, and that is used by all of the setup shell scripts (including `zowe-configure-instance.sh itself`).  What does the init-instance.sh script do ?

1000TurquoisePogs:
> @1000TurquoisePogs - what is the failure that caused this PR ? It looks as though you are checking for node 12 and doing some exports. The generic script `zowe-set-env.sh` should set the vars already, and that is used by all of the setup shell scripts (including `zowe-configure-instance.sh itself`). What does the init-instance.sh script do ?

This isn't about node 12 or exports, it's about initialization of `$INSTANCE_DIR/workspace` files

initInstance.js is the logic of the app-server's configure.sh. It has a bit to do, so I didn't want to do it all in one shell file. configure.sh invokes it.

The problem is that initInstance.js creates files, and currently does so only because of configure.sh, which means only at runtime.
People have been running zowe as an stc with an stc user that they never log in as, so the files become owned by a user that isn't a real user. This becomes a problem later with permissions.
Therefore, I want to run initInstance.js at creation time such that permissions are set up to the user who created the instance.

Joe-Winchester:
@1000TurquoisePogs - the design is that `ZWESVUSR` runs the `ZWESVSTC`.  If it creates file in the instance directory then they will be owned by itself.   The instance directory (as you say) is created by whoever ran the `zowe-configure-instance.sh` script, which is likely to just be Fred Smith or whoever is logged on.   The idea of the `ZWEADMIN` group is that the Zowe started task user of `ZWESVUSR` and Fred Smith are both members and it's the group permissions that allow both access to the instance directory.

Reading the doc it doesn't make this clear enough but this is the design point.  Copying @stevenhorsman  for his input, but I am reticent to have stuff moved from config for this reason, as only at config.sh do you know things what root dir is, the IP and host, and other stuff that isn't safe to burn in at install time.

1000TurquoisePogs:
I'm going to put this on the backburner, I think there's a less overkill way to do fix the issues I've been seeing with permissions. I think Joe is right that when ZWEADMIN is used properly, and files are group-writable, that it's not an issue.

Joe-Winchester:
@1000TurquoisePogs - do you want to close this PR ?  I saw on Tuesday May 19th that you'd asked on install-packaging slack for a review ?

stevenhorsman:
@1000TurquoisePogs - can we close this PR given that it's been sat in the queue for 5 months?

1000TurquoisePogs:
I think there's still issues but nobody has brought it up recently so I guess people have found solutions to the problem in the field.

PR #1334, opened 2020-05-04
zip 1310 - validate log directory not in runtime + refactoring and tests

stevenhorsman:
#### PR type
- [X] Feature

#### Relevant issues
Depends on #1322 
Fixes #1310

#### Changes proposed in this PR
- Add validation to check that the log directory supplied isn't inside the runtime directory
- Refactor code to reduce duplication and add a new file-utils script
- Add test for the new file-utils script
- Fix SMP/E packaging build so it can accept branch names with '/' in

Issue #1239, opened 2020-04-02
Create a verify only mode of run-zowe.sh


stevenhorsman:
**As a** Zowe System Admin,  
**I want** to just just the validate mode of Zowe  
**so that** I can get an idea if there are problems without starting all the the component address spaces.

*Details/notes*  
- I think creating an optional VALIDATE_ONLY=true flag in the instance.env and then post-validation checking for that and exiting with a message is the simplest way to do this
- I think we should clean up the -v option at the same time as it was very doc'd, and is difficult for customers to use - in future we could re-introduce that behaviour with a FAIL_ON_VALIDATE_ERROR flag or similar if required.

#### Acceptance Criteria

**Scenario 1:* No flag in instance.env
*Given* An instance.env which doesn't have a VALIDATE_ONLY property
*When* I start Zowe
*Then* All the components are started as previously

**Scenario 1:* VALIDATE_ONLY=true
*Given* An instance.env which has VALIDATE_ONLY=true
*When* I start Zowe
*Then* The validation runs
*And* There is a message indicating that we were in validation only mode
*And run-zowe.sh exits

**Scenario 2:* VALIDATE_ONLY not set to true
*Given* An instance.env which has VALIDATE_ONLY=false
*When* I start Zowe
*Then* All the components are started as previously

stevenhorsman:
@kathysychan - Hi Kathy, IIRC you asked for a feature like this. If so would this meet your requirements?

stevenhorsman:
> Yes. That would work. Just want to clarify, does that mean not having the VALIDATE_ONLY flag is the same as setting it to false?
The plan is that not having VALIDATE_ONLY is the same as it being set false (or set to anything apart from 'true')

> Also, please clarifies that validation will confirm if everything required for running Zowe is there and Zowe can start. Where will the validation output be?
This mode will run the validation scripts as they currently exist and the output will go into the STDERR of the job.

PR #1335, opened 2020-05-04
ZIP 1239 - Add validate only mode

stevenhorsman:
## PR type
- [X] Feature

#### Relevant issues
Depends on #1334 
Implements #1239 

- Add validation mode of Zowe.
If `VALIDATE_ONLY=true` is set in instance.env then launching Zowe will exit after the validation stage. This lets a Zowe administration run a check of whether all the Zowe install's component validation checks work without having to actually start any of the components.

Doc issue: https://github.com/zowe/docs-site/issues/1180


stevenhorsman:
Convenience tests: https://wash.zowe.org:8443/job/zowe-install-test/view/change-requests/job/PR-1335/2/
smp/e ptf tests: https://wash.zowe.org:8443/job/zowe-install-test/view/change-requests/job/PR-1335/3/

stevenhorsman:
Manual testing:
```
/u/stevenh/1335/instance-dir:>echo "VALIDATE_ONLY=true" >> instance.env
/u/stevenh/1335/instance-dir:>cat instance.env
#!/bin/sh

# Note this script is executed like a shell script, so can have embedded variables
...
VALIDATE_ONLY=true
```

STDOUT:
```
Appending JAVA_HOME/bin to the PATH...
Appending NODE_HOME/bin to the PATH...
OK: Node is working
Error 0: NODE Version 8.16.1 is not compatible with Zowe. Please use a different version. See https://docs.zowe.org/stable/troubleshoot/app-framework/app-known-issues.html#desktop-apps-fail-to-load for more details
1 errors were found during validatation, please check the message, correct any properties required in /u/stevenh/1335/instance-dir/instance.env and re-launch Zowe
1 errors were found during validatation, please check the message, correct any properties required in /u/stevenh/1335/instance-dir/instance.env and re-launch Zowe
Validation complete - VALIDATE_ONLY mode set to true, so Zowe will not start.
```

Joe-Winchester:
@stevenhorsman - there are some conflicts that I can't see how I can resolve easily, apologies

stevenhorsman:
@Joe-Winchester - the dependencies are due to PR #1334 which this requires not being merged.
Sorry, I'm having trouble understanding:
> Looks great - one thing I found doing some testing is that when I added lines to `validate.sh` scripts like "echo "JRW This is an error" it stopped Zowe but possibly now how I had thought it would, e.g.
> 
> ```
> Error 0: JRW
> Error: /u/winchj/zowe-1.11.0/runtime/bin/internal/run-zowe.sh 127: FSUM7351 not found
> ```
> 
> Is there an API way to do something in an error script that flags an error as occuring or is that waiting on the APIness of the error.sh script ?
That not found error doesn't look great. Can you let me know the changes you made that triggered it?

Issue #1336, opened 2020-05-05
Make utils scripts into SPI


stevenhorsman:
Sanitise and move (copy initially until the components are migrated over) the 'good' utils scripts into /bin/utils and add proper tests for these.
They can then be used for extenders.
We have a fair few candidates:
- [x] configure-java.sh
- [x] configure-node.sh
- [x] error.sh
- [x] validate-apiml-variables.sh
- [x]  validate-directory-is-accessible.sh
- [x]  validate-directory-is-writable.sh
- [x]  validate-java.sh
- [ ]  validate-keystore-directory.sh
- [x]  validate-node.sh
- [x]  validate-port-available.sh
- [x]  validate-static-definition-directory.sh
- [x]  validate-zosmf-host-and-port.sh
- [x]  validate-zowe-prefix.sh
- [x]  zosmfHttpRequest.js

Some of these are generic and some specific to components, so we might want some sub-foldering or something going on.

stevenhorsman:
I've finished creating the new scripts that cover these functions. I hope these will be removed as part of the HA/install-components work if versioning allows.

PR #1369, opened 2020-05-13
Users/stevenh/node utils

stevenhorsman:
#### PR type
- [X] Other... Add more component and extender properly tested SPI scripts

#### Relevant issues
Part of #1336 

- Add node utils function and refactor utils error reporting

### DIFF
--- a/bin/utils/common.sh
+++ b/bin/utils/common.sh
@@ -20,7 +20,9 @@ print_error_message() {
     ERRORS_FOUND=0
   fi
 
-  # echo error to standard out and err
+  # echo error to standard out and err - this was requested so that the errors go into STDOUT of the job
+  # and save people going into STDERR (and make it inline with the rest of the logs), but it does result
+  # in double outputting when called from shell environment, so maybe we should reconsider?
   echo "Error ${ERRORS_FOUND}: ${message}"
   echo "Error ${ERRORS_FOUND}: ${message}" 1>&2
   let "ERRORS_FOUND=${ERRORS_FOUND}+1"
--- a/bin/utils/common.sh
+++ b/bin/utils/common.sh
@@ -34,8 +34,8 @@ print_message() {
   echo "${message}"
 }
 
-# If LOG_FILE is an writable file then output to it, otherwise echo instead?
-log_message_if_applicable() {
+# If LOG_FILE is an writable file then output to it, otherwise echo instead
+log_message() {
   message=$1
   if [[ -n "${LOG_FILE}" ]] && [[ -w "${LOG_FILE}" ]];
   then
--- a/bin/utils/common.sh
+++ b/bin/utils/common.sh
@@ -10,12 +10,14 @@
 # Copyright IBM Corporation 2020
 ################################################################################
 
+#TODO NOW - test this file
+
 #TODO LATER - do we want to provide a ENV_VAR flag that toggles whether errors are printed or not?
 
 print_error_message() {
   message=$1
   #output an error and add to the count
-  if [ -z "${ERRORS_FOUND}" ];
+  if [[ -z "${ERRORS_FOUND}" ]];
   then
     ERRORS_FOUND=0
   fi
@@ -26,4 +28,15 @@ print_error_message() {
   echo "Error ${ERRORS_FOUND}: ${message}"
   echo "Error ${ERRORS_FOUND}: ${message}" 1>&2
   let "ERRORS_FOUND=${ERRORS_FOUND}+1"
+}
+
+# If LOG_FILE is an writable file then output to it, otherwise echo instead?
+log_message_if_applicable() {
+  message=$1
+  if [[ -n "${LOG_FILE}" ]] && [[ -w "${LOG_FILE}" ]];
+  then
+    echo ${message} >> $LOG_FILE 
+  else
+    echo ${message}
+  fi
 }
--- a/bin/utils/common.sh
+++ b/bin/utils/common.sh
@@ -10,8 +10,6 @@
 # Copyright IBM Corporation 2020
 ################################################################################
 
-#TODO NOW - test this file
-
 #TODO LATER - do we want to provide a ENV_VAR flag that toggles whether errors are printed or not?
 
 print_error_message() {
@@ -25,9 +23,15 @@ print_error_message() {
   # echo error to standard out and err - this was requested so that the errors go into STDOUT of the job
   # and save people going into STDERR (and make it inline with the rest of the logs), but it does result
   # in double outputting when called from shell environment, so maybe we should reconsider?
-  echo "Error ${ERRORS_FOUND}: ${message}"
   echo "Error ${ERRORS_FOUND}: ${message}" 1>&2
+  print_message "Error ${ERRORS_FOUND}: ${message}"
   let "ERRORS_FOUND=${ERRORS_FOUND}+1"
+}
+
+# In future we can add timestamps/message ids here
+print_message() {
+  message=$1
+  echo "${message}"
 }
 
 # If LOG_FILE is an writable file then output to it, otherwise echo instead?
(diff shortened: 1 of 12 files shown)


stevenhorsman:
@jackjia-ibm - if you get a chance to look this is the next set of utils changes with some of the refactoring from the previous PR done, but probably some more stuff that you can think of ways to improve

bin/utils/file-utils.sh
```diff
@@ -10,7 +10,27 @@
 # Copyright IBM Corporation 2020
 ################################################################################
 
-#TODO LATER - provide flag that toggles all functions to error if they exit non-zero?
+# TODO LATER - anyway to do this better?
```
jackjia-ibm, line 13: Not sure a better way, one thing we may be able to try to consolidate the variable names `INSTALL_DIR`, `ZOWE_ROOT_DIR`, and `ROOT_DIR`. But seems not the scope of this PR.

stevenhorsman:
Passing convenience tests: https://wash.zowe.org:8443/job/zowe-install-test/view/change-requests/job/PR-1369/6/ and SMP/E PTF test: https://wash.zowe.org:8443/job/zowe-install-test/view/change-requests/job/PR-1369/7/

bin/utils/node-utils.sh
```diff
@@ -0,0 +1,144 @@
...
+  fi
+
+  ls ${NODE_HOME}/bin | grep node$ > /dev/null
+  if [[ $? -ne 0 ]];
+  then
+    print_error_message "NODE_HOME: ${NODE_HOME}/bin does not point to a valid install of Node"
+    return 1
+  fi
+
+  node_ok=`${NODE_HOME}/bin/node -e "console.log('ok')" 2>&1`
```
jackjia-ibm, line 88: This line has a chance to hang in a rare case. So maybe we echo something before this line to indicate we are validating if node bin is functional. So when we checking the log, we are sure the flow hangs on this line.

Also seems better this section moved below `node -v` check.

bin/utils/node-utils.sh
```diff
@@ -0,0 +1,144 @@
...
+  echo "Could not work out the path to the utils directory. Please 'export ZOWE_ROOT_DIR=<zowe-install-directory' before running." 1>&2
+  exit 1
+fi
+
+# Source common util functions
+. ${utils_dir}/common.sh
+
+# TODO - how to test well given interaction and guess?
+# Interactive function that checks if the current NODE_HOME is valid and if not requests a user enters the node home path via command line
+prompt_node_home() {
```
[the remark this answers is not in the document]
stevenhorsman: I guess there are a couple of problems:
1. These scripts are called a fair way down the stack, so passing a flag through all is not great. We could make a opt on zowe-configure-instance.sh/zowe-setup-certificates.sh to make it non interactive and export a shell variable based on that? I'm not sure if `-i` would be confusing for non-interactive though!
2. What behaviour do we want for non-interactive? We obviously don't want to exit the script, so just print an error and continue?

bin/utils/common.sh
```diff
@@ -0,0 +1,42 @@
...
+# SPDX-License-Identifier: EPL-2.0
+#
+# Copyright IBM Corporation 2020
+################################################################################
+
+#TODO NOW - test this file
+
+#TODO LATER - do we want to provide a ENV_VAR flag that toggles whether errors are printed or not?
+
+print_error_message() {
```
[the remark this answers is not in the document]
stevenhorsman: Add print_message

bin/utils/node-utils.sh
```diff
@@ -0,0 +1,144 @@
...
+  node_ok=`${NODE_HOME}/bin/node -e "console.log('ok')" 2>&1`
+  if [[ ${node_ok} == "ok" ]]
+  then
+    echo "OK: Node is working"
+  else
+    print_error_message "NODE_HOME: ${NODE_HOME}/bin/node is not functioning correctly: ${node_ok}"
+    return 1
+  fi
+
+  node_version=`${NODE_HOME}/bin/node --version`
```
jackjia-ibm, line 97: `node -v` has chance to display an error rather than exact version number. When this happens, the flow will run in a dead loop kept complaining the version is not supported. We should detect and notify the user if this command hits an error and let the user to exit the flow to fix it offline.

FYI, this also happens on `java -version`.

bin/utils/node-utils.sh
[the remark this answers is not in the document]
jackjia-ibm: `NODE_HOME` is mandatory in this case and for future steps to start Zowe, so I think if it's non-interactive mode, it's safe we prompt the error and exit the script.

bin/utils/node-utils.sh
stevenhorsman: I've added the logs and moved this check below the version, which required some refactoring in order to still test

bin/utils/node-utils.sh
stevenhorsman: Added a new check and test for if we get an error when calling version

bin/utils/node-utils.sh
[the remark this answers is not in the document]
stevenhorsman: My concern is that not all of our scripts are idempotent, so I don't know if we will hit issues when the user fixes the issue then re-runs. Do you mind if I create a new issue for this rather than have to solve it right now?

bin/utils/file-utils.sh
stevenhorsman: Agreed - created https://github.com/zowe/zowe-install-packaging/issues/1373 to cover

bin/utils/node-utils.sh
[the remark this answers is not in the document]
jackjia-ibm: Yes that sounds good to me. Thanks!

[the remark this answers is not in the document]
stevenhorsman: Captured in https://github.com/zowe/zowe-install-packaging/issues/1374

stevenhorsman:
Tests: https://wash.zowe.org:8443/job/zowe-install-test/view/change-requests/job/PR-1369/10/ & https://wash.zowe.org:8443/job/zowe-install-test/view/change-requests/job/PR-1369/9

bin/utils/node-utils.sh
```diff
@@ -0,0 +1,159 @@
...
+      if [ "$rep" != "Y" ] && [ "$rep" != "y" ]
+      then
+        echo "Please enter a path to where node is installed.  This is the a directory that contains /bin/node "
+        read NODE_HOME
+        loop=1
+      fi
+    fi
+  done
+  export NODE_HOME=$NODE_HOME
+  echo "  NODE_HOME variable value="${NODE_HOME} >> $LOG_FILE
```
jackjia-ibm, line 62: Sorry didn't catch this early, this line should use `log_message` function.

bin/utils/node-utils.sh
```diff
@@ -0,0 +1,159 @@
...
+}
+
+ensure_node_is_on_path() {
+  if [[ ":$PATH:" != *":$NODE_HOME/bin:"* ]]
+  then
+    echo "Appending NODE_HOME/bin to the PATH..."
```
jackjia-ibm, line 68: Same reason, this line should use `print_message`. This function will be executed in `run-zowe.sh`.

stevenhorsman:
New tests:
- https://wash.zowe.org:8443/job/zowe-install-test/view/change-requests/job/PR-1369/11/
-https://wash.zowe.org:8443/view/Zowe/job/zowe-install-test/view/change-requests/job/PR-1369/15/
Fixed issues and refactored pre-reqs to remove duplicated code:
```Checking Node version...
Node version v8.17.0 is supported
Validating if node bin is functional...
Node bin is functional
Script zowe-check-prereqs.sh ended
```

bin/utils/node-utils.sh
stevenhorsman: Fixed

bin/utils/node-utils.sh
stevenhorsman: Fixed

PR #1376, opened 2020-05-15
Revert "Update default plugins to use ROOT_DIR relative path"

1000TurquoisePogs:
Reverts zowe/zowe-install-packaging#1362


1000TurquoisePogs:
no longer required.

PR #1380, opened 2020-05-19
Users/stevenh/java utils

stevenhorsman:
#### PR type
- [X] Other... Add more component and extender properly tested SPI scripts

#### Relevant issues
Depends on #1369 
Part of #1336 

- Add java utils function

### DIFF
--- a/bin/zowe-init.sh
+++ b/bin/zowe-init.sh
@@ -135,13 +135,8 @@ else
     echo "  ZOWE_ZOSMF_PORT variable value="$ZOWE_ZOSMF_PORT >> $LOG_FILE
 fi
 
-if [[ -z ${JAVA_HOME} ]]
-then
-    JAVA_HOME=/usr/lpp/java/J8.0_64
-else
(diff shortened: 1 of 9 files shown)


bin/zowe-init.sh
```diff
@@ -135,13 +135,8 @@ else
     echo "  ZOWE_ZOSMF_PORT variable value="$ZOWE_ZOSMF_PORT >> $LOG_FILE
 fi
 
-if [[ -z ${JAVA_HOME} ]]
-then
-    JAVA_HOME=/usr/lpp/java/J8.0_64
-else
-    echo "  JAVA_HOME variable value="${JAVA_HOME} >> $LOG_FILE
-fi
-locateJavaHome ${JAVA_HOME}
```
jackjia-ibm, line 144: Do we still need `locateJavaHome` and `getJavaVersion` function in this file?

stevenhorsman: Thanks Jack - I missed that from the clean up!

PR #1395, opened 2020-05-25
Permit AUX user to access the cross-memory server

### LINKED ISSUE #1361: ZWESECUR should PERMIT READ to AUXUSER to ZWES.IS
The manual https://docs.zowe.org/stable/user-guide/configure-xmem-server.html#step-4-configure-saf states that 

> "It is recommended that you also grant READ access to the STC user under which the ZWESASTC started task runs which is ZWESIUSR."

This is actually a requirement, so the security job ZWESECUR be updated to permit READ to AUXUSER:

```
000266 /* permit Zowe main server to use XMEM cross memory server       
000267   RLIST   FACILITY ZWES.IS ALL                                   
000268   RDEFINE FACILITY ZWES.IS UACC(NONE)                            
000269   PERMIT ZWES.IS CLASS(FACILITY) ACCESS(READ) ID(&ZOWEUSER.)     
000270   PERMIT ZWES.IS CLASS(FACILITY) ACCESS(READ) ID(&AUXUSER.)
```

ifakhrutdinov:
**WARNING: please make sure to look at the `Other information` section**

#### PR type
- [ ] Bugfix 
- [x] Feature 

#### Relevant issues

Fixes #1361

- Ensure that the AUX STC user has access to the cross-memory server by giving it READ access to `ZWES.IS`

- [ ] Yes
- [x] No

---
* Affected function: ESM rule assignment and removal in the ZWESECUR.vtl and ZWENOSEC.jcl files

---
* Description: `&AUXUSER` will be given READ access to `ZWES.IS`

---
* Part: ZWESECUR.vtl and ZWENOSEC.jcl

---

Doc issue/PR number: none

#### Other information
* Only the RACF part has been tested due to lack of environment and expertise. I'd appreciate if someone who has TSS/ACF2 environments and can confirm that the changes work.
* ACF2 has been using `&STCGRP` (why a group name?) as the role for `&ZOWEUSER` to access the cross-memory profile (`ZWES.IS`) and BPX profiles. `&AUXUSER` has been added to that role to gain access to `ZWES.IS`. Can the authors of the ACF2 instructions confirm this is acceptable? Would we want to use different roles?
* Will that be possible to run full regression testing on a system where we already have the expected behavior (perhaps river)?


ifakhrutdinov:
Hi @Joe-Winchester , who could review this?

Joe-Winchester:
@John-A-Davies - are you able to look at Irek's requests ?  They look good to me but it'd be good to get another set of eyes on them.

John-A-Davies:
I'm happy to review this.

workflows/templates/ZWESECUR.vtl
```diff
@@ -482,20 +491,25 @@ GROUP(&STCGRP.) +
 STCID(&AUXSTC.)
 F ACF2,REFRESH(STC)
 *
-* DEFINE ZOWE SERVER PERMISIONS ...................................
+* DEFINE XMEM SECURITY RESOURCES ..................................
```
John-A-Davies, line 494: I've checked the RACF statements.  None of the changes mention XMEMUSER.  This may be OK.   Elsewhere XMEMUSER and AUXUSER are set to the same value.  I think what you're saying with these changes is that ZOWEUSER and now also AUXUSER can both access the XMEM server (ZWES.IS).

ifakhrutdinov: Correct. I grant AUXUSER access to ZWES.IS, since that STC needs it. And I don't make the assumption that AUXUSER = XMEMUSER, since they may be different on some customer sites.

ifakhrutdinov:
@Joe-Winchester who could have a look at TSS and ACF2? I think it'd be best to ask the authors of the original instructions to take a look.

Joe-Winchester:
@MarkAckert - are you able to help @ifakhrutdinov 's request to have ACF/2 and TSS commands added for the changes above ?

MarkAckert:
@vit-tomica I think you contributed to the ACF2/TSS install JCL -  are you able to review?

workflows/templates/ZWESECUR.vtl
```diff
@@ -482,20 +491,25 @@ GROUP(&STCGRP.) +
...
 *
-* define a role holding the permissions and add &ZOWEUSER to it
+* define a role holding the permissions, add &AUXUSER and
+* &ZOWEUSER to it
 *
 SET X(ROL)
-INSERT &STCGRP. INCLUDE(&ZOWEUSER.) ROLE
+INSERT &STCGRP. INCLUDE(&AUXUSER.,&ZOWEUSER.) ROLE
```
vit-tomica, line 500: Here, if AUXUSER and ZOWEUSER variables are set to the same acid then the command fails with the `ACF00204 DUPLICATE DATA NOT ALLOWED IN FIELD INCLUDE` message and nothing is defined. I would suggest to split it into two commands, so the sequence would be:
```
SET X(ROL)
INSERT &STCGRP. INCLUDE(&ZOWEUSER.) ROLE
F ACF2,NEWXREF,TYPE(ROL)
CHANGE &STCGRP. INCLUDE(&AUXUSER.) ADD
```
In this case, if the variables are set to the same then only the `CHANGE` command fails with the "duplicate" message which is okay.

ifakhrutdinov: @vit-tomica thank you, will do

stevenhorsman:
@John-A-Davies - am I correct in thinking that this change requires HOLDDATA as the user will need to re-run ZWESECUR.jcl. Onno said that you are pulling this together.

John-A-Davies:
@stevenhorsman makes a good point.  HOLDDATA will be required because some actions in ZWESECUR.jcl will be different.  The HOLDDATA  instruction to existing installations might be to run ZWENOSEC.jcl and then re-run ZWESECUR.jcl.  Alternatively they could just run the extra commands manually.

I will create HOLDDATA if this is agreed.

John-A-Davies:
And ZWENOSEC.jcl will also need a corresponding un-do of the new actions that this PR creates.

ifakhrutdinov:
@John-A-Davies I thought I added that. Please see `ZWENOSEC.jcl`.

workflows/templates/ZWESECUR.vtl
ifakhrutdinov: Please see my updates. I've also fixed some missing `&` in the comments.

vit-tomica: Thanks. It looks OK now.

ifakhrutdinov:
```
. (substitute) (length 75) /* replace AUTOGID with GID(&amp;ADMINGID.) if AUTOGID is not enabled    */

. (substitute) (length 74) /* replace AUTOGID with GID(&amp;ADMINGID.) if AUTOGID is not enabled   */

. (substitute) (length 73) /* replace AUTOGID with GID(&amp;ADMINGID.) if AUTOGID is not enabled  */

. (substitute) (length 72) /* replace AUTOGID with GID(&amp;ADMINGID.) if AUTOGID is not enabled */

** ERROR build-workflow.rex cannot trim line after substitution

(length 72) /* replace AUTOGID with GID(&amp;ADMINGID.) if AUTOGID is not enabled */

[Pax.pack][ERROR] failed on pre hook

Connection to zzow01.zowe.marist.cloud closed.
```
Is this the reason some line missed `&`?

vit-tomica:
I think the line is too long now (with the added `&`) so that not even build-workflow.rex can squeeze it under 72 characters. Either the comment has to be shorter or split into more lines.

ifakhrutdinov:
Okay, I've reverted the change since the PR wasn't really about those ampersands. Now it builds.

ifakhrutdinov:
FYI, I don't have permissions to merge this.

OnnoVdT:
I updated the HOLD data. I noticed that it only has the 3 header lines, not the body.

PR #1419, opened 2020-06-02
Import local CA into p12 file to fix zosmf jwt SSO issue - target to staging

vit-tomica:
The zosmf JWT is inserted into fake certificate that is signed by Zowe local CA. The zowe local CA has to imported to p12 file as well before the p12 is imported to pkcs11 token using gskkyman.


vit-tomica:
@jackjia-ibm Jack, can you approve and merge this RC fix, please?

PR #1451, opened 2020-06-22
Docker compatibility with sso p12 key handling

timgerstel:
Signed-off-by: Timothy Gerstel <tim.gerstel@gmail.com>

#### PR type
- [X] Bugfix 
- [ ] Feature 

#### Relevant issues

Fixes 

- When setting up Zowe on Linux, the P12 public file will need to be transferred to the ZSS host in order to enable SSO with the Desktop.

- [ ] Yes
- [X] No

---

---

---

---
_multi-line description_

Doc issue/PR number:


bin/utils/java-utils.sh
```diff
@@ -63,7 +63,7 @@ validate_java_home() {
   fi
 
   # As we know the java -version command works then strip out the line we need
-  java_version_output=$(${JAVA_HOME}/bin/java -version 2>&1 | grep ^"java version")
+  java_version_output=$(${JAVA_HOME}/bin/java -version 2>&1 | grep ^"openjdk version")
```
1000TurquoisePogs, line 66: This will NOT work on z/os, even if it does work on linux.
What you really need is to test conditional to the OS that is being used, or try both if one fails

bin/utils/setup-log-dir.sh
```diff
@@ -44,7 +44,7 @@ get_install_log_directory() {
 
   INSTALL_LOG_DIR=$(get_full_path "${INSTALL_LOG_DIR}")
 
-  if { [[ ! -d "${INSTALL_LOG_DIR}" ]] || [[ ! -r "${INSTALL_LOG_DIR}" ]] }
+  if [[ ! -d "${INSTALL_LOG_DIR}" ]] || [[ ! -r "${INSTALL_LOG_DIR}" ]]
```
1000TurquoisePogs, line 47: Have to be careful about this, as this may be a bash vs sh incompatibility. I havent tried it, but if it proves to be an issue then it may need to be specified differently.

install/zowe-install.sh
```diff
@@ -10,7 +10,7 @@
 # Copyright IBM Corporation 2018, 2020
 ################################################################################
 
-if [ $# -lt 4 ]; then
+if [ $# -lt 2 ]; then
```
1000TurquoisePogs, line 13: This sacrifices z/os for linux. A conditional is needed instead.

bin/internal/run-zowe.sh
```diff
@@ -95,7 +95,7 @@ fi
 #Explorers may be present, but have a prereq on gateway, not desktop
 if [[ $LAUNCH_COMPONENT_GROUPS == *"DESKTOP"* ]]
 then
-  LAUNCH_COMPONENTS=zss,app-server,${LAUNCH_COMPONENTS} #Make app-server the first component, so any extender plugins can use its config
```
1000TurquoisePogs, line 98: This sacrifices z/os for linux. A conditional is needed somewhere isntead. A reasonable solution to me would be to have the conditional within zss' start.sh, such that it starts if on z/os, and issues an info or warning message when not, to tell people that it's being skipped and that if they need it, hopefully they have one set up on zos already.

bin/utils/java-utils.sh
```diff
@@ -63,8 +63,10 @@ validate_java_home() {
   fi
 
   # As we know the java -version command works then strip out the line we need
-  java_version_output=$(${JAVA_HOME}/bin/java -version 2>&1 | grep ^"java version")
-
+  java_version_output=$(${JAVA_HOME}/bin/java -version 2>&1 | grep ^"openjdk version")
+  if [[ $java_version_output == "" ]]; then
+    java_version_output=$(${JAVA_HOME}/bin/java -version 2>&1 | grep ^"java version")
+  fi
```
sakeerthy, line 69: doing the || instead of this works for me

sakeerthy: might be it works on linux not z/os?

sakeerthy:
It looks like you didn't change both java_versions in the pr https://github.com/zowe/zowe-install-packaging/pull/1451#pullrequestreview-441035372 when you tried using the || that's probably why it didn't work for you

bin/utils/java-utils.sh
timgerstel: this syntax is necessary for z/OS and docker compatibility

1000TurquoisePogs:
test shows that servers start, which is the area of the code this PR effects, so I think this does not break anything.

PR #1507, opened 2020-07-13
Build amd64 docker image

1000TurquoisePogs:
This change allows for docker to be built as-needed. The docker code is within https://github.com/zowe/zowe-dockerfiles/tree/master/dockerfiles/zowe-release/amd64/zowe-v1-lts and has a readme to explain how it is built, but in summary the ordinary build is downloaded, converted from ebcdic to ascii, installed, and then when the docker image is ran, environment variables and volume mounts to docker are translated over to be used in run-zowe.sh and instance.env, such that the code behaves the same except for docker-isms around configuration, but at runtime our code does not know the difference so this is just another way of delivering the same code. The exception is that zss & zis are not usable in docker, so the preexisting download mechanisms exist in order to get those onto z/os.


jackjia-ibm (approved):
Looks good to me. Thanks Sean!

Jenkinsfile
```diff
@@ -131,7 +135,8 @@ sed -e 's#{BUILD_BRANCH}#${env.BRANCH_NAME}#g' \
             'KEEP_TEMP_FOLDER': (params.KEEP_TEMP_FOLDER ? 'yes' : '')
           ],
           extraFiles          : (params.BUILD_SMPE ? 'zowe-smpe.zip,fmid.zip,pd.htm,smpe-promote.tar,smpe-build-logs.pax.Z,rename-back.sh' : ''),
-          keepTempFolder      : params.KEEP_TEMP_FOLDER
+          keepTempFolder      : params.KEEP_TEMP_FOLDER,
+          buildDocker      : (params.BUILD_SMPE ? 'yes' : '')
```
jackjia-ibm, line 139: Not sure if this is variable is used or not, but seems wrong parameter. should be `params.BUILD_DOCKER` instead of `params.BUILD_SMPE`.

jackjia-ibm (changes requested):

Jenkinsfile
1000TurquoisePogs: you're right, bad merge from the code.

jackjia-ibm (approved):
All good to me. Thanks Sean!

PR #1529, opened 2020-07-22
Add keyring utilities source dependency

stevenhorsman:
I'm been informed that a bunch of our components are pulling in the keyring-utilites repo, so I figured they are supposed to be in the source scans

### DIFF
--- a/manifest.json.template
+++ b/manifest.json.template
@@ -335,6 +335,14 @@
       }]
     },
     {
+      "componentGroup": "Keyring Utilities",
+      "entries": [{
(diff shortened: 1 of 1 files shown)


stevenhorsman:
@MarkAckert - is my guess that this is required correct? I've picked up the latest source tag, as there doesn't seem to be a staging branch

MarkAckert:
Yes, this is required - missed it on the first pass. Thanks for adding it in....  If you want the latest version, you can use `master` instead of `1.0.3` for tag.

PR #1545, opened 2020-07-28
Users/stevenh/ui explorer test

### LINKED ISSUE #1536: Create install for new explorer-ui-server component
After https://github.com/zowe/explorer-ui-server/issues/36 explorer-ui-server with be a top-level component, so needs integrating into the pax and build.

stevenhorsman:
#### Relevant issues
Fixes #1536 

Depends on https://github.com/zowe/explorer-ui-server/pull/37
zowe/explorer-jes#207

- pull in explorer-ui component and add install script for it.

### DIFF
--- a/manifest.json.template
+++ b/manifest.json.template
@@ -64,7 +64,9 @@
       "explode": "true"
     },
     "org.zowe.explorer-jes": {
-      "version": "~1.0.3-PR-207"
+      "version": "1.0.3-PR-207",
+      "repository": "libs-snapshot-local",
+      "artifact": "explorer-jes-1.0.3-pr-207-5-20200728140926.pax"
     },
     "org.zowe.explorer-mvs": {
       "version": "1.0.2"
--- a/manifest.json.template
+++ b/manifest.json.template
@@ -64,9 +64,7 @@
       "explode": "true"
     },
     "org.zowe.explorer-jes": {
-      "version": "1.0.3-PR-207",
-      "repository": "libs-snapshot-local",
-      "artifact": "explorer-jes-1.0.3-pr-207-5-20200728140926.pax"
+      "version": "~1.0.3-PR-207"
     },
     "org.zowe.explorer-mvs": {
       "version": "1.0.2"
--- a/manifest.json.template
+++ b/manifest.json.template
@@ -64,7 +64,7 @@
       "explode": "true"
     },
     "org.zowe.explorer-jes": {
-      "version": "1.0.2"
+      "version": "~1.0.3-PR-207"
     },
     "org.zowe.explorer-mvs": {
       "version": "1.0.2"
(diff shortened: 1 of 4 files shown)


stevenhorsman:
Converting to draft whilst we decide if we want to continue with the explorer-ui-server separation next release

NakulManchanda:
From @1000TurquoisePogs 
> I had a concern about filesystem structure:
- this seems like a shared library rather than a component in itself
- it currently looks that way because it doesnt have its own start.sh script
- today everything in /components does have a start.sh
so, today scripting can take for granted /components are all the same type of thing, which could be beneficial for some future management tools.
i was thinking it would be good to move this new not-quite-component code into a new folder such as /lib or /shared

NakulManchanda:
yes, its not used on its own, rather it get called by three explorers
so, we can have some conventions
shared -> looks better to me - as its still an executable
lib -> can be just for non executable things

NakulManchanda:
@stevenhorsman @jackjia-ibm 
these are ready for review and if all ok, ready to merge
All tests are passing, jack verified the folder structure as well.

**old strucutre**
component
  jes > server app
 mvs > server app
 uss > server app

Moved structure to new shared folder,
We layout one copy of server instead of three, 
We still start three instances of server though.

**new structure**
component
  jes > app
  mvs > app
  uss > app
shared
  explorer-ui-server

As per me,
No documentation required, as for end user, there wont be any difference

Related PR's
https://github.com/zowe/explorer-ui-server/pull/37
https://github.com/zowe/explorer-jes/pull/207
https://github.com/zowe/zowe-install-packaging/pull/1545

manifest.json.template
```diff
@@ -64,13 +64,16 @@
...
     "org.zowe.explorer-mvs": {
-      "version": "~1.0.0-STAGING"
+      "version": "1.0.2"
     },
     "org.zowe.explorer-uss": {
-      "version": "~1.0.0-STAGING"
+      "version": "1.0.3"
+    },
+    "org.zowe.explorer-ui-server": {
+      "version": "~0.2.13-PR-37"
```
jackjia-ibm, line 76: We need to revert version without PR-37.

manifest.json.template
```diff
@@ -64,13 +64,16 @@
       "explode": "true"
     },
     "org.zowe.explorer-jes": {
-      "version": "~1.0.0-STAGING"
+      "version": "~1.0.3-PR-207"
```
jackjia-ibm, line 67: Need to revert version without PR-207 before merging.

jackjia-ibm (approved):
Merge the other PRs first and modify versions in binaryDependencies.

manifest.json.template
NakulManchanda: reverted

manifest.json.template
NakulManchanda: reverted

PR #1577, opened 2020-08-05
fix for RC2

struga0258:


vit-tomica:
Tested. Works fine.

stevenhorsman:
Hi James, apologies I was out yesterday afternoon, so couldn't comment then, but I've been told the ZLC decided that all keyring related work has been pushed back to 1.15, so we don't need a new RC for this. Sorry for the miscommunication.

PR #1586, opened 2020-08-11
Update zlux source tags to released version

Roavill2:
Release has been picked but this is pre-emptive. Waiting on Sean Grady to create/run release for tags to be set and picked up.


Roavill2:
Sean has now run his build - checking myself to see if the new tags do exist in the expected places.

PR #1592, opened 2020-08-12
Add OS check for ExportPrivateKey

sakeerthy:
Signed-off-by: Suneeth Keerthy <skeerthy@rocketsoftware.com>

#### PR type
- [ ] Bugfix 
- [ ] Feature 

#### Relevant issues

Fixes 

- 

- [ ] Yes
- [ ] No

---

---

---

---
_multi-line description_

Doc issue/PR number:


bin/apiml_cm.sh
```diff
@@ -282,7 +282,9 @@ function export_service_certificate {
 function export_service_private_key {
     echo "Exporting service private key"
     echo "TEMP_DIR=$TEMP_DIR"
-    cat <<EOF >$TEMP_DIR/ExportPrivateKey.java
+
+    if [ `uname` = "OS/390" ]; then
+        cat <<EOF >$TEMP_DIR/ExportPrivateKey.
```
1000TurquoisePogs, line 287: missing .java at the end

bin/apiml_cm.sh
```diff
@@ -336,11 +338,65 @@ public class ExportPrivateKey {
     }
 }
 EOF
+    else
+        cat <<EOF >$TEMP_DIR/ExportPrivateKey.java
```
1000TurquoisePogs, line 342: you should add a code comment so that this code is maintained over time:
why is there a nearly duplicate java function? because the import needed for z/os keyring support will fail in docker or in development environments. this section is for docker and does not support keyrings.

stevenhorsman: Rather than the copy-paste of code, I wonder if it would be better to have a z/OS specific subclass of private key given that the differences seem to be that the z/os version has a little bit of extra logic?

1000TurquoisePogs: The problem I saw was that you can't do a conditional import statement, and I think the zos-only import was an issue leading to a crash. We can improve upon this if that's not the case.

Issue #1619, opened 2020-08-27
Update keyring automation to undo keyring changes for RACF


stevenhorsman:
I'm not sure if we can use ZWENOKEY.jcl to uninstall to cleanup keyring changes.
The uninstall doesn't (necessarily) have access to the installed code, including the JCL and if we decide to change any parameters we inject into ZWEKRING then we'd need to customise and run things multiple times, so we might just need to use commands to clean it up, like we do for racf settings with ZWESECUR

PR #1633, opened 2020-09-03
And RACF undo for keyring in testing

stevenhorsman:
#### Relevant issues
Implements #1619

- Add RACF commands that undo the keyring changes during uninstall phase of installation testing


stevenhorsman:
Convenience tests: https://wash.zowe.org:8443/view/Zowe/job/zowe-install-test/view/change-requests/job/PR-1633/2/
keyring tests: https://wash.zowe.org:8443/view/Zowe/job/zowe-install-test/view/change-requests/job/PR-1633/3/
smp/e ptf tests: https://wash.zowe.org:8443/view/Zowe/job/zowe-install-test/view/change-requests/job/PR-1633/5/

vit-tomica:
It looks OK to me overall. However, I'm not sure whether this will work with ACF2 because ACF2 doesn't have any TSO command that could be used like RACDCERT or TSS. Instead, ACF2 has its own command processor that is invoked and then ACF2 commands are interactively issued within this command processor.

vit-tomica (approved):

stevenhorsman:
> It looks OK to me overall. However, I'm not sure whether this will work with ACF2 because ACF2 doesn't have any TSO command that could be used like RACDCERT or TSS. Instead, ACF2 has its own command processor that is invoked and then ACF2 commands are interactively issued within this command processor.

Ok, thanks for the info - we might need to embed part of the ZWENOKEY for ACF2, which is a bit of duplication, but not too bad. We can chat more next week.

Issue #1199, opened 2020-03-17
Create plugin installation test based on Ansible playbook


jackjia-ibm:
**Is your feature request related to a problem? Please describe.**

We need to create a new installation test case which will automatically install a plugin and verify it.

**Describe the solution you'd like**

- Create new Ansible playbook to install plugin.
- Create new installation test case which will install and verify plugin.
- The test case should call the Ansible playbook to perform plugin installation and validate with the method defined by the plugin.

**For this stage, we will install and test the sample api/plugin.**

#### Acceptance Criteria

- There should be a new Ansible playbook which allows others to consume and install any plugin.
- There should be a new option show in install test Jenkins job to allow the user to run  a plugin installation test.
- The test should generate similar test report like others installation test.

jackjia-ibm:
**Stage 1**: create `scripts/zowe-install-extension.sh` which supports these parameters:

- `-e|--extension`: file path and name of the extension package
- `-r|--runtime-dir`: path to zowe runtime directory
- `-i|--instance-dir`: path to zowe instance directory
- `-d|--target-dir`: optional, default value is `<zowe-runtime-dir>/extensions`

Below parameters will be removed in future stages:

- `-t|--type`: value of `component` or `desktop-plugin`.

Example:
```
./zowe-install-extension.sh \
  -e /path/to/sample-node-api.pax \
  -t /path/to/my/extensions-runtime-dir \
  -r /var/zowe/runtime \
  -d /var/zowe/instance-dir \
  -t component
```

`type=component` can be used to install `sample-node-api`
`type=desktop-plugin` can be used to install `sample-trial-app`.

The behaviors for component installation:

- unpax the package and put into target folder
- validate the package if it has `/bin` folder and lifecycle scripts. Currently `start.sh` is required.
- modify instance-dir `instance.env` to `EXTERNAL_COMPONENTS`

The behaviors for desktop plugin installation:

- unpax the package and put into target folder
- validate the package if it has `pluginDefinition.json`.
- call `${INSTANCE_DIR}/bin/install-app.sh ${COMPONENT_DIR}` to install plugin.

**Stage 2**: modify sample repositories to define and include `manifest.json` into final package.
https://github.com/zowe/sample-node-api
https://github.com/zowe/sample-trial-app

TBD.

jackjia-ibm:
[...]

PR #1671, opened 2020-09-23
Implemented a script to automate extension/plugin installation

JadinLuong:
Signed-off-by: Jadin Luong <Jadin.Luong@ibm.com>

#### PR type
- [ ] Bugfix 
- [x] Feature 
- [X] Other... Please describe:
Enhancement

#### Relevant issues

Fixes Stage 1 of issue #1199 

This PR introduces a single script "zowe-install-extension.sh" which assists in installing plugins and extensions on to zowe by performing the necessary steps to do so.

- [ ] Yes
- [X] No


scripts/zowe-install-extension.sh
```diff
@@ -0,0 +1,148 @@
...
+        -t|--type) #Represents the type of extension whether its a desktop plugin or web component
+        shift
+        if [[ $1 == "component" ]] || [[ $1 == "desktop-plugin" ]]; then
+            INSTALL_TYPE=$1
+        else
+            error_handler "-t|--type: Invalid value, possible values are 'component' and 'desktop-plugin'"
+        fi
+        shift
+        ;;
+        *)
```
jackjia-ibm, line 124: In addition to the original design, would like to add one more set of `-l|--logs-dir` to write installation logs to it. The value is optional, default will be `<instance-dir>/logs`. The log file should look like `install-extension-2020-09-24-13-00-26.log`.

In the log file, it should shows all the command line parameters, and all the steps including failures. We can follow other log as example.

Try to use `bin/utils/setup-log-dir.sh` and follow other scripts how the logs are written.

scripts/zowe-install-extension.sh
```diff
@@ -0,0 +1,148 @@
...
+
+SCRIPT_NAME="$(basename $0)"
+EXTENSION_PATH=""
+INSTANCE_DIR=""
+TARGET_DIR=""
+EXTENSION_NAME=""
+INSTALL_TYPE=""
+
+error_handler(){
+    echo "$SCRIPT_NAME: $1" 1>&2
```
jackjia-ibm, line 11: we can use functions defined in `bin/utils/common.sh`.

scripts/zowe-install-extension.sh
```diff
@@ -0,0 +1,148 @@
...
+    exit 1
+}
+
+install_component(){
+    cd $TARGET_DIR
+
+    if [ ! -d "$EXTENSION_NAME" ]; then
```
jackjia-ibm, line 18: Check `bin/utils/file-utils.sh` for functions if we can use.

jackjia-ibm (commented):
@stevenhorsman after some discussion, i think `/bin/zowe-install-extension.sh` should be a better location. What do you think? I compared the pax and runtime directory, I think it makes sense to keep this script into runtime directory after installation, so the user can use it again to install other extensions.

scripts/zowe-install-extension.sh
```diff
@@ -0,0 +1,148 @@
...
+    $INSTANCE_DIR/bin/install-app.sh $TARGET_DIR/$EXTENSION_NAME
+
+    echo "Zowe desktop plugin has been installed."
+}
+
+while [ $# -gt 0 ]; do #Checks for parameters
+    arg="$1"
+    case $arg in
+        -e|--extension) #Represents the path pointed to the extension's pax file
+        shift
```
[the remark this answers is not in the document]
stevenhorsman: yes /bin is the correct place for this not /scripts.

scripts/zowe-install-extension.sh
```diff
@@ -0,0 +1,148 @@
+#!/bin/sh
+
+SCRIPT_NAME="$(basename $0)"
+EXTENSION_PATH=""
+INSTANCE_DIR=""
+TARGET_DIR=""
+EXTENSION_NAME=""
+INSTALL_TYPE=""
```
stevenhorsman, line 8: I'm not sure we need to initialise these variables here

scripts/zowe-install-extension.sh
```diff
@@ -0,0 +1,148 @@
...
+        mkdir $EXTENSION_NAME
+    else
+        # Clean up the directory and make a new one
+        rm -rf $EXTENSION_NAME
+        mkdir $EXTENSION_NAME
+    fi
+
+    cd $EXTENSION_NAME
+
+    pax -ppx -rf "$EXTENSION_PATH"
```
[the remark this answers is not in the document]
stevenhorsman: `file-utils.sh` has code that will convert everything to a full path

scripts/zowe-install-extension.sh
```diff
@@ -0,0 +1,148 @@
...
+    echo "Zowe desktop plugin has been installed."
+}
+
+while [ $# -gt 0 ]; do #Checks for parameters
+    arg="$1"
+    case $arg in
+        -e|--extension) #Represents the path pointed to the extension's pax file
+        shift
+        if [[ "$1" = *.pax ]]; then
+            EXTENSION_PATH=$1
```
stevenhorsman, line 90: I think that renaming the variable to reflect that it should point to the pax would add some clarity.

scripts/zowe-install-extension.sh
```diff
@@ -0,0 +1,148 @@
...
+    if [ ! -e "pluginDefinition.json" ]; then
+        error_handler "$EXTENSION_PATH: Invalid package"
+    fi
+
+    # Uses install-app.sh in zowe-instance-dir to automatically set up the extension onto zowe
+    $INSTANCE_DIR/bin/install-app.sh $TARGET_DIR/$EXTENSION_NAME
+
+    echo "Zowe desktop plugin has been installed."
+}
+
```
stevenhorsman, line 83: please use /bin/internal/zowe-set-env.sh to ensure that common zowe script environment properties are set

scripts/zowe-install-extension.sh
```diff
@@ -0,0 +1,148 @@
...
+done
+
+if [ -z $EXTENSION_PATH ] || [ -z $INSTANCE_DIR ] || [ -z $EXTENSION_NAME ] || [ -z $INSTALL_TYPE ]; then
+    #Ensures that the required parameters are entered, otherwise exit the program
+    error_handler "Missing parameters, try: zowe-install-extensions.sh -e {PATH_TO_EXTENSION} -i {ZOWE_INSTANCE_DIR} -n {NAME_OF_EXTENSION} -t component|desktop-plugin"
+fi
+
+if [ -z $TARGET_DIR ]; then
+    #Assigns TARGET_DIR to the default directory since it was not set to a specific directory
+    TARGET_DIR="$INSTANCE_DIR/extensions"
```
stevenhorsman, line 136: I don't think it is good practice to be installing extension runtimes into the instance directory. That is supposed to hold configuration and workspace details for the instance, not binaries

scripts/zowe-install-extension.sh
```diff
@@ -0,0 +1,148 @@
...
+        else
+            error_handler "-t|--type: Invalid value, possible values are 'component' and 'desktop-plugin'"
+        fi
+        shift
+        ;;
+        *)
+        error_handler "$1 is an invalid flag\ntry: zowe-install-extensions.sh -e {PATH_TO_EXTENSION} -i {ZOWE_INSTANCE_DIR} -n {NAME_OF_EXTENSION} -t component|desktop-plugin"
+    esac
+done
+
```
stevenhorsman, line 128: We need some validation to ensure that the target directory is not inside the zowe runtime

stevenhorsman (commented):
I have concerns that this piece of work is being rushed in without a proper design consultation and without us having finalised the package design. It seems to ignore some features we have discussed, so I'm worried that it's going to leave us with code that doesn't do what we want, but we are not able to change.

scripts/zowe-install-extension.sh
```diff
@@ -0,0 +1,148 @@
...
+        mkdir $EXTENSION_NAME
+    else
+        # Clean up the directory and make a new one
+        rm -rf $EXTENSION_NAME
+        mkdir $EXTENSION_NAME
+    fi
+
+    cd $EXTENSION_NAME
+
+    pax -ppx -rf "$EXTENSION_PATH"
```
1000TurquoisePogs, line 71: Assuming the extension is in a pax archive will prevent every package manager from working, as none of them I'm aware of function pax files. Zowe tooling should probably be post-extraction exclusively in order to be compatible with package managers.

scripts/zowe-install-extension.sh
```diff
@@ -0,0 +1,148 @@
...
+
+    # Validates the pax file is correctly configured when unpaxed
+    if [ ! -e "pluginDefinition.json" ]; then
```
1000TurquoisePogs, line 74: This check is already done within install-app.sh

scripts/zowe-install-extension.sh
```diff
@@ -0,0 +1,148 @@
...
+if [ ! -e $TARGET_DIR ]; then
+    echo "Creating extensions folder at $TARGET_DIR"
+    mkdir $TARGET_DIR
+fi
+
+if [[ $INSTALL_TYPE == "component" ]]; then
+    install_component
+else
+    install_desktop_plugin
+fi
```
1000TurquoisePogs, line 148: What is the intended use case of this script such that this is an if-else? Is it intended to be run by users directly or is it intended to be contained within product install scripts?
A product may contain zero or more components and zero or more plugins. Another possible direction is that components contain plugins, so that installing a component installs the plugins.

scripts/zowe-install-extension.sh
```diff
@@ -0,0 +1,148 @@
...
+
+if [ ! -e $TARGET_DIR ]; then
+    echo "Creating extensions folder at $TARGET_DIR"
+    mkdir $TARGET_DIR
+fi
+
+if [[ $INSTALL_TYPE == "component" ]]; then
+    install_component
+else
```
1000TurquoisePogs, line 146: What about APIML static definitions?
What about iframe apps?

jes-explorer is both an app, a component, and registers a static definition and creates an iframe object. Should these tasks be moved out of the jes-explorer code and into this script?

1000TurquoisePogs (changes requested):
As a solution to #1199 this is fine, but while creating an install script it is worthwhile to make this useful to end users since it's nearly the same task.

I second @stevenhorsman 's concerns about this needing general agreement so that we are unlikely to end up with dead or contradictory code.

The extension products I am aware of come with their own scripts for installation because they are composed of more than one extension (a component + a plugin, or multiple plugins with no component)
For those products, it's beneficial to have well-known Zowe scripts that handle validation of extension & registration into zowe. install-app.sh is one such script and this uses it which is good., but there are other sections of the script I'd like to get a group discussion about so that we are comfortable with a direction of scripting & what it means to end users.

jackjia-ibm:
Thanks a lot @1000TurquoisePogs . I also agree with both of you and understand this affect and tangle with many other discussions. This PR is most likely a PoC for me to better understand the situation. And I talked with Steve offline to assure him there will be thorough discussion and agreement to move forward before we merge anything.

I had several long discussions with Jadin recently to help on clear up my head. I think almost all concerns you listed here we had noticed and have a solution proposed. But I couldn't make it this week. Maybe next week Architecture call I'm good to explain the proposal so we can have detail discussion on what's good/bad and missing.

1000TurquoisePogs:
By the way, I brought this up at the architecture call this week but I think many people including me forgot that ZIS also has plugins, and they work different from everything else, so we should not only have tests about it but it would be great to have a unified installer also handle them too. Currently, the install script is hiding (not seen in $INSTANCE_DIR accidentally) within https://github.com/zowe/zss/blob/staging/bin/zis-plugin-install.sh

JadinLuong:
@jackjia-ibm, @stevenhorsman, @1000TurquoisePogs. Sorry for the delay, I've pushed a newer version of zowe-install-extension.sh script with the requested changes. For the most part, I've tried my best to address all of your requests but there were a few things I did not follow through with because Jack and I are still figuring things out but we are definitely going to satisfy those requests in the future. Feel free to take a look at the script and once again, I'm open to all the feedback and concerns you guys may have. Thanks!

JadinLuong:
Hey @jackjia-ibm, I've made the modifications to the extension installation script for stage 2 of the issue https://github.com/zowe/zowe-install-packaging/issues/1199. Open to all feedback you may have!

bin/zowe-configure-component.sh
```diff
@@ -0,0 +1,159 @@
...
+
+    cd $COMPONENT_PATH
+
+    if [[ -f "$COMPONENT_PATH/manifest.yaml" ]]; then
+        MANIFEST_PATH="$COMPONENT_PATH/manifest.yaml"
+    elif [[ -f "$COMPONENT_PATH/manifest.yml" ]]; then
+        MANIFEST_PATH="$COMPONENT_PATH/manifest.yml"
+    elif [[ -f "$COMPONENT_PATH/manifest.json" ]]; then
+        MANIFEST_PATH="$COMPONENT_PATH/manifest.json"
+    fi
```
jackjia-ibm, line 60: What happens if `MANIFEST_PATH` is empty? The user could point to a wrong folder, or a legacy extension package without manifest.

bin/zowe-install-component.sh
```diff
@@ -0,0 +1,182 @@
...
+if [ -z $LOG_DIRECTORY ]; then
+    LOG_DIRECTORY="$INSTANCE_DIR/logs"
+fi
+
+. ${ZOWE_ROOT_DIR}/bin/utils/setup-log-dir.sh
+set_install_log_directory "${LOG_DIRECTORY}"
+validate_log_file_not_in_root_dir "${LOG_DIRECTORY}" "${ZOWE_ROOT_DIR}"
+set_install_log_file "configure-component"
+
+log_message "Zowe Root Directory: $ZOWE_ROOT_DIR"
```
jackjia-ibm, line 167: We don't have a lint standard for shell scripts. But generally I think there are few rules we agreed on:
- capital case variables are for global purpose, it may be exported as environment variables. For local variables, we can use lower case variable names.
- `"${LOG_DIRECTORY}"` is way better in several ways:
  * curly brackets make sure the variable name will be matched exactly without any ambiguous.
  * wrap the value with `""` will make sure it works ok if there are any spaces in the value. For example, if `TARGET_DIR ="/var/my folder with space"`, the line `${ZOWE_ROOT_DIR}/bin/zowe-configure-component.sh -o $COMPONENT_NAME -d $TARGET_DIR -i $INSTANCE_DIR` will have parsing error.

Could you convert the variable name, adding proper quotations, etc.

bin/zowe-install-component.sh
```diff
@@ -0,0 +1,182 @@
...
+    fi
+
+    # Locate manifest file
+    if [[ -f "manifest.yaml" ]]; then
+        MANIFEST_FILE="manifest.yaml"
+    elif [[ -f "manifest.yml" ]]; then
+        MANIFEST_FILE="manifest.yml"
+    elif [[ -f "manifest.json" ]]; then
+        MANIFEST_FILE="manifest.json"
+    fi
```
jackjia-ibm, line 57: Seems we need to move some functions into utils scripts. I started working on new utils scripts here https://github.com/zowe/zowe-install-packaging/pull/1822/files#diff-bb1b34e0a3e7bc5990caded4ab189e5aa353bbdf24ae7f16bdd2c9fe848206f2R13. Could be merged soon.

bin/zowe-configure-component.sh
```diff
@@ -0,0 +1,159 @@
...
+    print_error_message "$1"
+    exit 1
+}
+
+install_component(){
+
+    # Represents the current value of EXTERNAL_COMPONENT variable in instance.env
+    EXT_COMP_VAL=$(eval "grep '^EXTERNAL_COMPONENTS=' $INSTANCE_DIR/instance.env | cut -f2 -d= | cut -f1 -d#")
+
+    if [ "$EXT_COMP_VAL" = " " ]; then
```
jackjia-ibm, line 25: Instead of adding full component lifescycle script path to `EXTERNAL_COMPONENTS `, we can add component name as value, and also enforce `ZWE_EXTENSION_DIR` to be target_dir parameter value.

bin/zowe-configure-component.sh
```diff
@@ -0,0 +1,159 @@
...
+        ;;
+        *)
+            error_handler "$1 is an invalid flag\ntry: zowe-configure-component.sh -o {COMPONENT_NAME} -i {ZOWE_INSTANCE_DIR}"
+            shift
+    esac
+done
+
+if [ -z $TARGET_DIR ]; then
+    TARGET_DIR=$DEFAULT_TARGET_DIR
+fi
```
jackjia-ibm, line 132: We need to validate TARGET_DIR matches ZWE_EXTENSION_DIR defined in instance.env or not. If they don't match, then we need to reject the configure step.

bin/zowe-install-component.sh
```diff
@@ -0,0 +1,182 @@
...
+        -d|--target_dir) # Represents the path to the desired target directory to place the extensions (optional)
+            shift
+            TARGET_DIR=$(get_full_path "$1")
+            shift
+        ;;
+        -l|--logs-dir) # Represents the path to the installation logs
+            shift
+            LOG_DIRECTORY=$1
+            shift
+        ;;
```
jackjia-ibm, line 123: let's also add --native option here.

bin/zowe-install-component.sh
```diff
@@ -0,0 +1,182 @@
...
+    if [ ! -z $INSTANCE_DIR ]; then #instance_dir exists
+        ZWE_EXTENSION_DIR=$(eval "grep '^ZWE_EXTENSION_DIR=' $INSTANCE_DIR/instance.env | cut -f2 -d=")
+    fi
+    if [ -z $ZWE_EXTENSION_DIR ]; then
+        #Assigns TARGET_DIR to the default directory since it was not set to a specific directory
+        TARGET_DIR=$DEFAULT_TARGET_DIR
+    else
+        TARGET_DIR=$ZWE_EXTENSION_DIR
+    fi
+fi
```
jackjia-ibm, line 145: This check only applies to extension, which native is false.

bin/zowe-install-component.sh
```diff
@@ -0,0 +1,182 @@
...
+
+# Checks to see if target directory is inside zowe runtime
+validate_file_not_in_directory "$TARGET_DIR" "$ZOWE_ROOT_DIR"
```
jackjia-ibm, line 148: Only call this check for extensions.

bin/zowe-configure-component.sh
```diff
@@ -0,0 +1,218 @@
...
+        # we only enable if extension has commands.start defined in manifest
+        log_message "- enable ${COMPONENT_NAME} for instance ${INSTANCE_DIR}"
+        # append to EXTERNAL_COMPONENTS
+        update_zowe_instance_variable "EXTERNAL_COMPONENTS" "${COMPONENT_NAME}" "true"
+    fi
+}
+
+install_desktop_plugin(){
+    desktop_plugin_path=$(read_component_manifest "${component_path}" ".desktopPlugin[].path" 2>/dev/null)
+    if [ "${desktop_plugin_path}" != "null" ] && [ -n "${desktop_plugin_path}" ]; then
```
jackjia-ibm, line 83: @JadinLuong found a potential problem here. `desktop_plugin_path` could be an array which contains multiple lines. We need to iterate all lines and call `install-app.sh` on each of them. Examples of how to handle this scenario can be found in `bin/utils/component-utils.sh` `process_component_desktop_iframe_plugin` or `process_component_apiml_static_definitions` functions. One if iterate through lines, the other one is iterate for maximum 20 items to avoid parsing another small portion of json.

scripts/zowe-install-extension.sh
jackjia-ibm: The script has been updated to check manifest definition and call `install-app.sh` directly.

scripts/zowe-install-extension.sh
jackjia-ibm: The script has been upgraded to read manifest. If manifest defines multiple plugins, it will handle them all.

scripts/zowe-install-extension.sh
jackjia-ibm: APIML definitions and IFrame apps won't make differences in `install` and `configureInstance` stages. Those information can also be defined in manifest, but will be handled at `configure` stage, which is part of runtime, because they requires valid value of `ZOWE_EXPLORER_HOST` and `STATIC_DEF_CONFIG_DIR`.

jackjia-ibm:
Jadin Luong  8:43 AM
Builds look good :+1::skin-tone-3::
Convenience build: https://wash.zowe.org:8443/view/Zowe/job/zowe-install-test/view/change-requests/job/PR-1671/7/
SMP/e build: https://wash.zowe.org:8443/view/Zowe/job/zowe-install-test/view/change-requests/job/PR-1671/8/

@stevenhorsman @1000TurquoisePogs I think the PR is ready for another review. This PR includes install/configure scripts as well as necessary changes to use those new scripts to install build-in Zowe components.

There was discussion to add `autoEncoding` to manifest to tell whether we need the installer to tag the files. This is not added since the installer needs this information before reading the manifest file. There is `detect_component_manifest_encoding` function added to `bin/utils/component-utils.sh` to detect common encodings. The user can set option `-e|--auto-encoding` to `yes|no|auto` to control how tagging will be applied.

Related docs update is https://github.com/zowe/docs-site/pull/1526.

bin/utils/common.sh
```diff
@@ -44,6 +44,15 @@ log_message() {
   fi
 }
 
+# print message to stdout and also write message to log
+print_and_log_message() {
+  message=$1
+  echo "${message}"
+  if [[ -n "${LOG_FILE}" ]] && [[ -w "${LOG_FILE}" ]]; then
+    echo "${message}" >> $LOG_FILE
+  fi
```
stevenhorsman, line 53: This code looks like it can be refactored with log_message

bin/zowe-configure-component.sh
```diff
@@ -0,0 +1,221 @@
...
+# Constants
+DEFAULT_TARGET_DIR=/global/zowe/extensions
+
+#######################################################################
+# Prepare shell environment
+if [ -z "${ZOWE_ROOT_DIR}" ]; then
+  export ZOWE_ROOT_DIR=$(cd $(dirname $0)/../;pwd)
+fi
+
+. ${ZOWE_ROOT_DIR}/bin/utils/utils.sh
```
stevenhorsman, line 46: I wonder if we should call {RUNTIME_DIR}/bin/internal/zowe-set-env.sh if this might be called by an extension in order to ensure out settings are consistent?

bin/zowe-install-component.sh
```diff
@@ -0,0 +1,333 @@
...
+#######################################################################
+# Constants
+DEFAULT_TARGET_DIR=/global/zowe/extensions
+
+#######################################################################
+# Prepare shell environment
+if [ -z "${ZOWE_ROOT_DIR}" ]; then
+  export ZOWE_ROOT_DIR=$(cd $(dirname $0)/../;pwd)
+fi
+
```
stevenhorsman, line 56: I wonder if we should call {RUNTIME_DIR}/bin/internal/zowe-set-env.sh if this might be called by an extension in order to ensure out settings are consistent?

bin/zowe-install-component.sh
```diff
@@ -0,0 +1,333 @@
...
+    if [ -n "${INSTANCE_DIR}" ]; then
+        # CALL CONFIGURE COMPONENT SCRIPT
+        cmd="${ZOWE_ROOT_DIR}/bin/zowe-configure-component.sh"
+        cmd="${cmd} -c \"${component_name}\""
+        cmd="${cmd} -d \"${TARGET_DIR}\""
+        cmd="${cmd} -i \"${INSTANCE_DIR}\""
+        # we should always have LOG_FILE at this time
+        cmd="${cmd} -f \"${LOG_FILE}\""
+        if [ "${IS_ZOWE_CORE}" = false ]; then
+          cmd="${cmd} -n"
```
stevenhorsman, line 169: I think this should be -k now

stevenhorsman (changes requested):
A couple of minor comments/questions, but generally I think it's great. Really love the restructuring of zowe-install.sh

bin/zowe-install-component.sh
jackjia-ibm: good catch. thanks!

stevenhorsman (approved):

bin/utils/common.sh
jackjia-ibm: The `log_message` has chance to write to stdout instead of log file. So if we call `print_message` then `log_message`, we have chance to write twice to stdout. If we can change log_message to never write to stdout, we should be ok. Just I don't know what consequences could be. Do you think it's ok to change `log_message` by removing this:
```
  else
    echo ${message}
```

stevenhorsman: I see, without changing the log_message to take an extra paramter maybe it is best to keep as is then

jackjia-ibm:
Another round of test on build 39 has been done with most recent changes:
Convenience build: https://wash.zowe.org:8443/view/Zowe/job/zowe-install-test/view/change-requests/job/PR-1671/24/
PTF: https://wash.zowe.org:8443/view/Zowe/job/zowe-install-test/view/change-requests/job/PR-1671/25/

bin/utils/zowe-variable-utils.sh
```diff
@@ -76,3 +76,45 @@ validate_zowe_prefix() {
...
+    is_append=false # default value is false
+  fi
+
+  variable_name_exists=$(grep "^ *${variable_name}=" ${INSTANCE_DIR}/instance.env)
+
+  if [ -z "${variable_name_exists}" ]; then
+    echo "${variable_name}=${variable_value}" >> ${INSTANCE_DIR}/instance.env
+  else
+    curr_variable_value=$(read_zowe_instance_variable "${variable_name}")
+    # FIXME: we have risks if value has "#" character
```
1000TurquoisePogs, line 103: FYI I'm confirming I don't know of any app-server or zss parameters that would normally have "#", so it limits this exposure to unusual end-user input.

bin/zowe-configure-component.sh
```diff
@@ -0,0 +1,224 @@
...
+}
+
+install_desktop_plugin(){
+    iterator_index=0
+    desktop_plugin_path=$(read_component_manifest "${component_path}" ".desktopPlugins[${iterator_index}].path" 2>/dev/null)
+    while [ "${desktop_plugin_path}" != "null" ] && [ -n "${desktop_plugin_path}" ]; do
+        log_message "- install Zowe desktop plugin"
+        cd "${component_path}"
+        # Uses install-app.sh in zowe-instance-dir to automatically set up the component onto zowe
+        ${INSTANCE_DIR}/bin/install-app.sh "$(get_full_path ${desktop_plugin_path})"
```
1000TurquoisePogs, line 90: should work fine.
note that this should work regardless of if it is for app-server or zss.

bin/zowe-configure-component.sh
```diff
@@ -0,0 +1,224 @@
...
+    commands_start=$(read_component_manifest "${component_path}" ".commands.start" 2>/dev/null)
+    if [ "${commands_start}" != "null" ] && [ -n "${commands_start}" ] && [ "${IS_ZOWE_CORE}" = "false" ]; then
+        # we only enable if extension has commands.start defined in manifest
+        log_message "- enable ${COMPONENT_NAME} for instance ${INSTANCE_DIR}"
+        # append to EXTERNAL_COMPONENTS
+        update_zowe_instance_variable "EXTERNAL_COMPONENTS" "${COMPONENT_NAME}" "true"
+    fi
+}
+
+install_desktop_plugin(){
```
1000TurquoisePogs, line 83: all of the terminology around "desktop" is misleading since app framework plugins can be ZSS plugins that have no visibility in the desktop, if the developer chooses. It could merely be a rest api plugin, with no visuals.

bin/zowe-configure-component.sh
```diff
@@ -0,0 +1,224 @@
...
+    iterator_index=0
+    desktop_plugin_path=$(read_component_manifest "${component_path}" ".desktopPlugins[${iterator_index}].path" 2>/dev/null)
```
1000TurquoisePogs, line 85: Application Framework is the full name of what the plugins belong to. You'll see it in the conformance and docs. Perhaps `.appfw` or `.appfwplugins` would be short but still correct. `.desktopplugins` is not correct.

bin/zowe-configure-instance.sh
```diff
@@ -256,6 +256,18 @@ fi
 chmod -R 755 ${INSTANCE}
 chmod -R 755 ${INSTANCE_DIR}/bin
 
+# Go through Zowe build-in components and see they need to be configured for current instance
+component_list="jobs-api files-api api-catalog discovery gateway caching-service apiml-common-lib explorer-ui-server explorer-jes explorer-mvs explorer-uss"
```
1000TurquoisePogs, line 260: what about `app-server` and `zss`?

bin/zowe-install-component.sh
```diff
@@ -0,0 +1,339 @@
...
+    # automatically tag files
+    if [ "${RUN_ON_ZOS}" = "true" ]; then
+        manifest_encoding=
+        if [ "${AUTO_ENCODING}" = "auto" ]; then
+            manifest_encoding=$(detect_component_manifest_encoding "${TARGET_DIR}/temp-ext-dir")
+        fi
+        if [ "${AUTO_ENCODING}" = "yes" -o "${manifest_encoding}" = "ISO8859-1" ]; then
+            # automatically tag files
+            log_message "- automatically tag files"
+            $ZOWE_ROOT_DIR/scripts/utils/tag-files.sh "${TARGET_DIR}/temp-ext-dir" 1>&2 >> $LOG_FILE
```
1000TurquoisePogs, line 117: i hope our documentation will explain that `tag-files.sh` is opinionated about which extensions should be in which encoding. There is more than one right answer, but `tag-files.sh` chooses one.

install/zowe-install.sh
```diff
@@ -10,199 +10,277 @@
...
-if [[ ${root_dir_existing_children} -gt 0 ]]; then
-    if [[ -f "${ZOWE_ROOT_DIR}/manifest.json" ]]
-    then
-        OLD_VERSION=$(cat ${ZOWE_ROOT_DIR}/manifest.json | grep version | head -1 | awk -F: '{ print $2 }' | sed 's/[",]//g' | tr -d '[[:space:]]')
-        NEW_INSTALL="false"
-        echo "  $ZOWE_ROOT_DIR contains version ${OLD_VERSION}. Updating this install to version ${ZOWE_VERSION}."
-        echo "  Backing up previous Zowe runtime files to ${ZOWE_ROOT_DIR}.${OLD_VERSION}.bak."
-        mv ${ZOWE_ROOT_DIR} ${ZOWE_ROOT_DIR}.${OLD_VERSION}.bak
+install_buildin_components() {
+  component_list="jobs-api files-api api-catalog discovery gateway caching-service apiml-common-lib explorer-ui-server explorer-jes explorer-mvs explorer-uss"
```
1000TurquoisePogs, line 169: no app-server or zss here either?

bin/zowe-configure-component.sh
jackjia-ibm: We were planning to have `desktopPlugins` and `zisPlugins` in the design. If we merge into `appfwPlugins`, the `install-app.sh` will need to take full responsibility to handle appfw plugins installation and zowe package installer won't be able to handle it because it doesn't have visibilities of the differences from manifest. Could you confirm this is the intention?

1000TurquoisePogs:
I've concluded reviewing and while the code may work, I haven't tested it with any external components yet. But, do note that using the name `desktop` to cover both app-server and zss plugins is not correct, and should be changed even if functional.

bin/zowe-configure-instance.sh
jackjia-ibm: We will add app-server and zss once those packages are converted to follow manifest definition. Currently it's backward compatible, we had converted apiml and explorer apis/uis, so we just start from here.

bin/zowe-configure-component.sh
jackjia-ibm: Sorry I think I confused zss plugin with zis plugin. I will change desktopPlugins to appfwPlugins if it's not accurate.

bin/zowe-install-component.sh
jackjia-ibm: Added to docs site PR1526.

install/zowe-install.sh
jackjia-ibm: Yes for now. We will need to convert app-server/zss packages to follow manifest definition before the installer can proper handle it. So for now, the zlux installation is still in old way.

jackjia-ibm:
A quick verification on the last change: https://wash.zowe.org:8443/view/Zowe/job/zowe-install-test/view/change-requests/job/PR-1671/26/.

PR #1725, opened 2020-10-07
Allow servers to additionally log to files

1000TurquoisePogs:
This is a copy, paste, and adaptation of the log file code that app-server and zss use.
If approved, the existing app-server and zss code will be removed so that we can rely on this without duplication.

Currently logging is handled by JCL with the exception of app-server and zss which unless otherwise configured will log to files in $INSTANCE_DIR/logs with the names zssServer and appServer.

This code brings log file ability to mediation layer and explorers too, and possibly external components. It can be toggled off, and I'm not sure whether the default should be on or off.

The following features exist
* Log file name includes component name
* Log file name includes timestamp
* Log files can be placed in any folder desired, but defaults to $INSTANCE_DIR/logs
* Each components logs can be kept according to a maximum count. The default is 4. After the maximum is reached, the oldest will be deleted... this keeps the log directory manageable.

The log files are per-run, so a log file can become arbitrarily big the longer each component is running. This code does not attempt to compact or split files while a component is running.

Questions and problems unresolved:
* ~This doesnt seem to capture STDERR. Attempts to capture STDERR may prevent STDERR from being utilized in JCL, so I am not sure if this is an acceptable compromise, but I'm fine with putting STDOUT & STDERR into a single stream and thus a single file.~
My mistake. I remember seeing a lot more logs back in the early days of zowe and wondered where they were. Seems this PR really does capture everything, its just that things are less verbose than they used to be.

* It would make sense if each file was named after the component, but app-server and zss have logs that predate this PR, so for backwards compatbility their log names are a little different
* Should there be files for each stage of a component, such as validate & configure also going into log files? It could be like $COMPONENT-$STAGE-TIMESTAMP.log

### DIFF
--- a/bin/internal/prepare-workspace.sh
+++ b/bin/internal/prepare-workspace.sh
@@ -96,9 +96,9 @@ get_log_filename() {
     COMPONENT_NAME=$component_id
   fi
 
-  if [ "${COMPONENT_NAME}" = "zss"]; then
+  if [ "${COMPONENT_NAME}" = "zss" ]; then
     COMPONENT_NAME="zssServer" #backwards compatibility
-  elif [ "${COMPONENT_NAME}" = "app-server"]; then
+  elif [ "${COMPONENT_NAME}" = "app-server" ]; then
     COMPONENT_NAME="appServer" #backwards compatibility
   fi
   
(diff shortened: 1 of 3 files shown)


1000TurquoisePogs:
In today's arch call people mentioned it would be cool to have the logs contain more verbose info than the job output.
Seems to me that because our logs are fairly unified in format, there's a simple enhancement to make:

user specification of most verbose level they'd like to see in job (file will be at max verbosity)
at this point, how to make servers print verbosely is not unified, and for performance servers should probably not be defaulting to their most verbose output.
however, whatever their output is, the user could customize how much of it goes into job output. since files dont have much limitation on output size, we dont need to constrain the file output, just the job output.

1000TurquoisePogs:
question to solve: how can we write to a file with 1 verbosity, but to jobs with another verbosity?
i think it will work like this:
`start.sh | tee myfile.txt | grep myfilter`
that way, the file gets everything from STDOUT, and tee keeps STDOUT unmodified until grep filters it, then the remaining STDOUT goes into the job output.

PR #1726, opened 2020-10-08
Client certificate setup for testing x509 certificate auth

achmelo:
Signed-off-by: jandadav <janda.david@gmail.com>

https://github.com/zowe/zowe-install-packaging/pull/1713
https://github.com/zowe/docs-site/pull/1443

#### PR type
- [ ] Bugfix 
- [x] Feature 

- The purpose of this PR is to enable additional client certificate and CA to be populated in keystores/keyring to enable certificate authentication

- [ ] Yes
- [x] No

#### Tests:
- Current staging for comparison
https://wash.zowe.org:8443/job/zowe-install-test/job/staging/680/testReport/


playbooks/roles/configure/tasks/add_client_certificate_acf2.yml
```diff
@@ -0,0 +1,44 @@
+---
+# this playbook runs SZWESAMP(ZWEKRING)
```
stevenhorsman, line 2: Update this comment

playbooks/roles/configure/tasks/upload_client_certificates.yml
```diff
@@ -0,0 +1,52 @@
+---
+# DESC
+
+- name: Check if work_dir_remote has a value
+  fail:
+    msg: work_dir_remote is required
+  when: work_dir_remote is not defined or work_dir_remote is none or work_dir_remote == ''
```
stevenhorsman, line 7: I think you have a few more variables used and you could consider using the existing task for validation eg.
```
# check variables
- import_role:
    name: common
    tasks_from: validate_variables
  vars:
    variable_list:
    - work_dir_remote
    - work_dir_local
    - inventory_hostname
```

tests/installation/src/__tests__/basic/install-api-gen.ts
```diff
@@ -9,9 +9,9 @@
  */
 
 import {
-    checkMandatoryEnvironmentVariables,
-    installAndGenerateApiDocs,
-  } from '../../utils';
+  checkMandatoryEnvironmentVariables, identifySecuritySystem,
```
stevenhorsman, line 12: Can you separate `identifySecuritySystem` onto a new line to match the linting of the rest of the variables please

tests/installation/src/__tests__/basic/install.ts
```diff
@@ -9,11 +9,11 @@
  */
 
 import {
-  checkMandatoryEnvironmentVariables,
+  checkMandatoryEnvironmentVariables, identifySecuritySystem,
   installAndVerifyConvenienceBuild,
   showZoweRuntimeLogs,
 } from '../../utils';
-import { TEST_TIMEOUT_CONVENIENCE_BUILD } from '../../constants';
+import {SECURITY_SYSTEM_ACF2, TEST_TIMEOUT_CONVENIENCE_BUILD} from '../../constants';
```
stevenhorsman, line 16: `SECURITY_SYSTEM_ACF2` isn't used, so shouldn't be imported

tests/installation/src/__tests__/extended/security-systems/convenience/acf2.ts
```diff
@@ -13,7 +13,7 @@ import {
   installAndVerifyConvenienceBuild,
   showZoweRuntimeLogs,
 } from '../../../../utils';
-import { TEST_TIMEOUT_CONVENIENCE_BUILD } from '../../../../constants';
+import {SECURITY_SYSTEM_ACF2, TEST_TIMEOUT_CONVENIENCE_BUILD} from '../../../../constants';
```
stevenhorsman, line 16: You seem to be switching between `const security_system = identifySecuritySystem();` and hardcoding. I don't mind which you use, but it would be good to have consistency. If you are adding multiple imports then the linting style followed is to separate each import on a new line:
```
import { 
  TEST_TIMEOUT_CONVENIENCE_BUILD, 
  SECURITY_SYSTEM_ACF2,
} from '../../../../constants';
```
These issues should be resolved in all the test files

stevenhorsman (changes requested):
Initial rounds of code review - just a few inconsistencies noted.

playbooks/roles/configure/tasks/upload_client_certificates.yml
```diff
@@ -0,0 +1,52 @@
...
+      copy:
+        src: APIML_External_Certificate_Authority.cer
+        dest: "{{ work_dir_local }}/{{ inventory_hostname }}/"
+      delegate_to: localhost
+
+    - import_role:
+        name: common
+        tasks_from: upload_file
+      vars:
+        filename_to_upload: APIML_External_Certificate_Authority.cer
```
jackjia-ibm, line 21: Would like to convert the client certificate file paths/names to variables so the user who consumes the playbook can use them to configure their own client certificates, not the one prepared by us.

Similar this also applies to USER-cert.cer.

tests/installation/src/utils.ts
```diff
@@ -367,3 +367,14 @@ export async function showZoweRuntimeLogs(serverId: string, extraVars: {[key: st
     debug(`showZoweRuntimeLogs failed: ${e}`);
   }
 };
+
+export function identifySecuritySystem(): string {
```
jackjia-ibm, line 371: I think we don't need to identify security system from test cases based on `TEST_SERVER` name, the `zos_security_system` variable can be configured per host. By keeping `marist` to use the default value `RACF`, we can add `playbooks/host_vars/marist-2.yml` with content of `zos_security_system: ACF2` and `playbooks/host_vars/marist-3.yml` with content of `zos_security_system: TSS`. This way, it's easier to extend when we have more ACF2/TSS servers.

Just we need to remove the marist line from `playbooks/.gitignore`, which was added to avoid accidentally publish server password.

jackjia-ibm (changes requested):
Thanks @achmelo  and @jandadav for the PR. Please see the comments. Also I don't see test case updates which consume the client certificate uploaded, will that be in another PR?

playbooks/roles/configure/tasks/upload_client_certificates.yml
jackjia-ibm: Once we defined these new variables, we can update `playbooks/README.md` `Other Build Variables` section to instruct the users how to update them.

achmelo:
convenience build https://wash.zowe.org:8443/job/zowe-install-test/view/change-requests/job/PR-1726/81/
smpe ptf https://wash.zowe.org:8443/job/zowe-install-test/view/change-requests/job/PR-1726/84/

playbooks/roles/start/tasks/wait_for_zowe.yml
```diff
@@ -42,17 +42,17 @@
...
+#  uri:
+#    url: "https://{{ zowe_external_domain_name }}:{{ zowe_zlux_port }}"
+#    follow_redirects: all
+#    method: GET
+#    validate_certs: false
+#  register: _result
+#  until: _result.status == 200
+#  retries: "{{ wait_for_zowe_service_retries | int }}"
+#  delay: "{{ wait_for_zowe_service_delay | int }}"
+#  delegate_to: localhost
```
stevenhorsman, line 55: What's the reason for this being commented out?

tests/installation/src/utils.ts
```diff
@@ -20,7 +20,7 @@ import {
   ZOWE_FMID,
   ANSIBLE_ROOT_DIR,
   SANITY_TEST_REPORTS_DIR,
-  INSTALL_TEST_REPORTS_DIR,
+  INSTALL_TEST_REPORTS_DIR, SECURITY_SYSTEM_ACF2, SECURITY_SYSTEM_RACF, SECURITY_SYSTEM_TSS,
```
stevenhorsman, line 23: Now the security system has been moved to the host vars I don't think these need to be imported

jackjia-ibm (changes requested):
Thanks for the fixes! And sorry for one more I could miss in last review.

playbooks/roles/start/tasks/wait_for_zowe.yml
achmelo: good catch, thanks

tests/installation/src/utils.ts
achmelo: we dont, I removed them

playbooks/group_vars/marist.yml
```diff
@@ -26,3 +26,5 @@ zowe_smpe_volser: ZOWE02
 zowe_zlux_terminal_telnet_port: 623
 # optional, this is the folder where we store Zowe FMIDs on the z/OS server
 zowe_fmids_dir_remote: /ZOWE/zowe-fmids
+# Restrict permissions to keystores after the certificate setup script
+zowe_lock_keystore: false
```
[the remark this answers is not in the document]
achmelo: sure, I was not certain where to put them, is it now correct?

jandadav:
Hi @jackjia-ibm @stevenhorsman We are looking to merge this PR, as it is a prereq for the https://github.com/zowe/zowe-install-packaging/pull/1713. Your comments should be adressed.

playbooks/group_vars/marist.yml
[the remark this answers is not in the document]
stevenhorsman: It sounds like Jack was suggesting the value defaults to `true` in `playbooks/roles/configure/defaults/main.yml`. Do you need it set to false for some specific use cases? Could we set it in the test for that?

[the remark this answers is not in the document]
achmelo: This have to be false as default in playbooks. We need to add CA certificate to apiml truststore for testing x509 authentication.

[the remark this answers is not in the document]
stevenhorsman: So, Jack's point is that we want to be careful about trying to test as close to the customer's defaults as we can, which means the default should be true. I'm suggesting that we set it to false for the specific tests that are going to be checking for client certificates, which I guess are in a subsequent PR? Why does this not work?

stevenhorsman (approved):

playbooks/group_vars/marist.yml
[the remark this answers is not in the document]
stevenhorsman: @jackjia-ibm - for your reference I had a chat with Andrej and David and got a better understanding. The current way that it is set-up/planned is that `zowe_lock_keystore` is set to `false` for all automation as it is required for the configuration of client certificate support. The test plan for that is to add a new test that runs in all of the suites (convenience, ptf, keyring etc) and check client certificates, so the setting would be required to `false` on everything. My understanding that there was going to be a new extended test suite for client certificate support wasn't correct, which might have enabled just setting `zowe_lock_keystore: false` for that suite.
Due to this I think we have to have `zowe_lock_keystore: false` in all the automated tests otherwise it won't work, so setting this in the defaults/main.yml is probably the least confusing place.
This PR is holding up other work, so I'll merge it now, but please let me/us know if we've missed anything here and we can spin it into the next/a. new PR.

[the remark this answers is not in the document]
stevenhorsman: After getting Jack's opinion and learning about how the installation and sanity tests are designed to act independently. 
Summary is:
- playbooks provide result exactly same as manual installation with possibilities for customization — same as what we suggested in docs.zowe.org
- installation test defines the test case. in our case, this test case should be customize the installation to use x509 client certificate and the corresponding test cases should pass
- sanity test should be able to test any zowe instance, no matter it’s installation by playbooks or not
(see slack convo for more details)

I think that gives us the following options:

1. No new extended test suite:
- Setting `'zowe_lock_keystore': false` in all the installation test scripts `tests/installation/src/__tests__ `
- A new toggle something like `CLIENT_CERTIFICATE_CONFIGURED` set to `true` in `playbooks/roles/verify/tasks/main.yml`

2. 1+ extended test suite for client certificates
- Setting `'zowe_lock_keystore': false` in the specific test suite(s)
- A new toggle something like CLIENT_CERTIFICATE_CONFIGURED set to `not zowe_lock_keystore` in `playbooks/roles/verify/tasks/main.yml`

Both cases:
- `zowe_lock_keystore: true` in `playbooks/roles/configure/defaults/main.yml`
- When the new test is added, it only runs if `CLIENT_CERTIFICATE_CONFIGURED` is defined and `true`

jackjia-ibm (approved):
Thanks all!

PR #1756, opened 2020-10-27
Record Zowe install attributes for later ZIS plugin installs

1000TurquoisePogs:
ZIS plugins need to be installed into a dataset that Zowe will be using, so that they can be loaded without asking the user to list all these locations or customize JCL.
A good place would be a new dataset that is in addition to the datasets we currently have. We can make an empty one for the installer to put content into.
However, currently, no dataset information is remembered. If we create one, the instance does not know where it is.

We must do 4-5 things
1) create a dataset for this purpose
2) modify the installer to record this info in a temporary location that instance creation can retrieve
3) have some manual intervention in case the temporary location got wiped
4) modify the zis plugin installer to utilize this new dataset
5) put the zis plugin installer in the instance directory, current users were finding it in root directory which is not friendly.


1000TurquoisePogs:
So far I've taken a shot at number 2.

nosrednayduj:
It would thrill me to pieces to have all of the zowe data sets recorded somewhere on USS, so I could get started without having to ask the user where their stuff was.

1000TurquoisePogs:
Closing this in order to focus on splitting this ticket into smaller tasks.

PR #1763, opened 2020-10-30
Add zlinux build to build process

1000TurquoisePogs:
In this PR, a new stage is made to build docker for zlinux. I decided to make this a stage separate from docker for linux, so you can see progress better, but it's a very similar task with the exception that zlinux needs to do the actions over ssh, while the linux build happens within a docker node inside of our primary jenkins docker node.
The zlinux build completes more quickly than the linux build, since it's a separate system that is ssh into.
The result ends up on artifactory and optionally dockerhub. For artifactory, you can see the output here: https://zowe.jfrog.io/zowe/libs-snapshot-local/org/zowe/1.17.0-PR-1763/ where zlinux contains the string "s390x", while linux contains "amd64".

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -15,6 +15,7 @@ node('ibm-jenkins-slave-nvm') {
 
   def pipeline = lib.pipelines.generic.GenericPipeline.new(this)
   def manifest
+  def zowePaxUploaded
 
   pipeline.admins.add("jackjia", "stevenh", "joewinchester", "markackert")
 
@@ -172,7 +173,7 @@ sed -e 's#{BUILD_BRANCH}#${env.BRANCH_NAME}#g' \
       if (params.BUILD_DOCKER) {
         // this is a hack to find the zowe.pax upload
         // FIXME: ideally this should be reachable from pipeline property
-        def zowePaxUploaded = sh(
+        zowePaxUploaded = sh(
           script: "cat .tmp-pipeline-publish-spec.json | jq -r '.files[] | select(.pattern == \".pax/zowe.pax\") | .target'",
           returnStdout: true
         ).trim()
@@ -253,10 +254,6 @@ sed -e 's#{BUILD_BRANCH}#${env.BRANCH_NAME}#g' \
       if (params.BUILD_DOCKER) {
         // this is a hack to find the zowe.pax upload
         // FIXME: ideally this should be reachable from pipeline property
-        def zowePaxUploaded = sh(
-          script: "cat .tmp-pipeline-publish-spec.json | jq -r '.files[] | select(.pattern == \".pax/zowe.pax\") | .target'",
-          returnStdout: true
-        ).trim()
         echo "zowePaxUploaded=${zowePaxUploaded}"
         if (zowePaxUploaded == "") {
           sh "echo 'content of .tmp-pipeline-publish-spec.json' && cat .tmp-pipeline-publish-spec.json"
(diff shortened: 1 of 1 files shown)


1000TurquoisePogs:
complete build here https://wash.zowe.org:8443/job/zowe-install-packaging/job/PR-1763/24/

Jenkinsfile
```diff
@@ -182,6 +182,84 @@ sed -e 's#{BUILD_BRANCH}#${env.BRANCH_NAME}#g' \
...
+               sudo docker login -u ${USERNAME} -p ${PASSWORD} &&
+               sudo docker push ${USERNAME}/zowe-v1-lts:s390x
+            """
+          }
+          sshCommand remote: Z_SERVER, command: \
+          """
+             rm -rf zowe-build/${env.BRANCH_NAME}_${env.BUILD_NUMBER}
+          """
+          //TODO need way to cleanup docker to save space, dont know how to auto-yes here
+          //              docker system prune -y
```
jackjia-ibm, line 240: It's `docker system prune -f`.

Jenkinsfile
```diff
@@ -15,6 +15,7 @@ node('ibm-jenkins-slave-nvm') {
 
   def pipeline = lib.pipelines.generic.GenericPipeline.new(this)
   def manifest
+  def zowePaxUploaded
```
jackjia-ibm, line 18: Current way of uploading artifacts works fine, but I found for each build, it appears 2-3 builds here https://zowe.jfrog.io/zowe/webapp/#/builds/zowe-install-packaging%20::%20PR-1763/25/1604121016053. It's because we call `uploadArtifacts` several times. If we want to make it better, can define

```
def zoweArtifacts = []
```

Then at the end of packaging operation, assign it with:

```
zoweArtifacts = [
      '.pax/zowe.pax',
      '.pax/zowe-smpe.zip',
      '.pax/smpe-promote.tar',
      '.pax/pd.htm',
      '.pax/smpe-build-logs.pax.Z',
      '.pax/AZWE*'
    ]
```

Then change docker `uploadArtifacts` to `zoweArtifacts.add('zowe-v1-lts.s390x.tar')` and `zoweArtifacts.add('zowe-v1-lts.amd64.tar')`.

Last move publish stage after docker build stages and change to:

```
pipeline.publish(artifacts: zoweArtifacts)
```

This is optional. Or let me know if you think it's good and would like me to add a commit.

jackjia-ibm (approved):
LGTM

Jenkinsfile
```diff
@@ -161,10 +162,9 @@ sed -e 's#{BUILD_BRANCH}#${env.BRANCH_NAME}#g' \
     ]
   )
 
-  // 
   pipeline.createStage(
```
jackjia-ibm, line 165: It seems the 2 docker build stages can run in parallel. Please let me know if the zLinux build takes too long and we need to convert them into parallel stages.

1000TurquoisePogs:
I talked with Jack and found the suggestions challenging to implement
* parallel should be worked on, but I believe it requires an update to another repo, so it will be tackled next.
* multi artifactory result issue is due to the nature of the build process, by which the pax is needed on the docker systems for building (leading to min 2) but then an issue by which zlinux and docker-in-docker are different nodes, being unable to share the upload process (resulting in 3 artifactory results)

Since this only effects the Jenkinsfile, I'm skipping the test on this one because there's no change in the actual code.

PR #1783, opened 2020-11-10
Add sanity tests for the caching service apis

taban03:
Signed-off-by: at670475 <andrea.tabone@broadcom.com>
Bound to https://github.com/zowe/api-layer/pull/925
This PR allows to deploy the API ML Caching Service on Marist and to run its sanity tests.

#### PR type
- [ ] Bugfix 
- [x] Feature 

#### Relevant issues

Fixes 

- 

- [ ] Yes
- [x] No

---

---

---

---
_multi-line description_

Doc issue/PR number:


taban03:
Successful tests here https://wash.zowe.org:8443/job/zowe-install-test/view/change-requests/job/PR-1783/35/

stevenhorsman:
@jackjia-ibm - I think we need to decide whether we want to have a new extended test that has the caching service switched on before we can finalise this change. If caching-service is optional than I don't see how that can be avoided even if it adds more test time

taban03:
> @jackjia-ibm - I think we need to decide whether we want to have a new extended test that has the caching service switched on before we can finalise this change. If caching-service is optional than I don't see how that can be avoided even if it adds more test time

Good point Steven. Any opinion @balhar-jakub ?

balhar-jakub:
I am not sure if the Caching Service can be optional later on. 
This is more about hiding it until it is time for the Caching service to be in production.

stevenhorsman:
Thanks for the info Jakub, do you have an estimate on when we might be turning the caching service on by default in the production release?

scripts/instance.template.env
```diff
@@ -41,9 +41,11 @@ APIML_SECURITY_AUTH_PROVIDER=zosmf
 # List of discovery service URLs separated by comma
 ZWE_DISCOVERY_SERVICES_LIST=https://${ZOWE_EXPLORER_HOST}:${DISCOVERY_PORT}/eureka/
 # Enable debug logging for Api Mediation Layer services
-APIML_DEBUG_MODE_ENABLED=false
+APIML_DEBUG_MODE_ENABLED=true
 ####################
 # caching service
+# if true, caching service is deployed on marist along with the rest of API ML and its sanity tests are run
+ZOWE_CACHING_SERVICE_START=true
```
jackjia-ibm, line 48: This value seems not updated based on the configuration of `zowe_apiml_caching_service_start`. Also if the default value is `true`, then it means it will be turned on by default for all users. Would like to confirm with Jakub what's the best time to default this to be true.

Another question is, if the user has this entry in `instance.env` and is set to `false`. Then later on when we want to switch it to `true`, the upgrade may not work well because it's hard to determine if this value is added by previous migration script, or it's customized by the user already and stay it as `false` is the intention. To avoid we have to switch this false/true back and forth, maybe we don't add this to `instance.env.template`. What we can do is if the user is using Ansible, we modify `playbooks/roles/configure/tasks/configure_instance.yml` to append this value.

tests/sanity/test/apiml/test-caching-service-apis-via-gateway.js
```diff
@@ -0,0 +1,115 @@
...
+
+let getToken = async () => {
+  const uuid = testUtils.uuid();
+  testUtils.log(uuid, 'ZOWE_CACHING_SERVICE_START value ' + process.env.ZOWE_CACHING_SERVICE_START);
+  return await testUtils.login(uuid);
+};
+
+describe('test caching service via gateway', function() {
+  before('verify environment variables', function () {
+    if (process.env.ZOWE_CACHING_SERVICE_START !== undefined && process.env.ZOWE_CACHING_SERVICE_START == 'false') {
```
jackjia-ibm, line 41: Will need to modify `playbooks/roles/verify/tasks/main.yml` to properly pass the value to the test. The test is running on Linux side, it doesn't get the environment variables defined in `instance.env`.

```
ZOWE_CACHING_SERVICE_START: "{{ zowe_apiml_caching_service_start|string|lower }}"
```

I think the check `!== undefined` here also implicates that `ZOWE_CACHING_SERVICE_START` default value is `true`. This should be consistent if we want to change the default value from `instance.env.template`.

tests/sanity/test/apiml/test-caching-service-apis-via-gateway.js
```diff
@@ -0,0 +1,115 @@
...
+  expect(response.data).to.not.be.empty;
+};
+
+let assertStatusNoContent = (response) => {
+  expect(response.status).to.equal(204);
+};
+
+let getToken = async () => {
+  const uuid = testUtils.uuid();
+  testUtils.log(uuid, 'ZOWE_CACHING_SERVICE_START value ' + process.env.ZOWE_CACHING_SERVICE_START);
```
jackjia-ibm, line 35: In test console log, it says: `API ML Tests ZOWE_CACHING_SERVICE_START value undefined`.

jackjia-ibm (changes requested):
Please see my comments if those make sense to you. Thanks Andrea!

playbooks/roles/start/tasks/wait_for_zowe.yml
```diff
@@ -102,3 +102,27 @@
...
+#  uri:
+#    url: "https://{{ zowe_external_domain_name }}:{{ zowe_apiml_gateway_port }}/api/v1/cachingservice/cache"
+#    follow_redirects: none
+#    method: GET
+#    validate_certs: false
+#  register: _result
+#  until: _result.status == 200
+#  retries: "{{ wait_for_zowe_service_retries | int }}"
+#  delay: "{{ wait_for_zowe_service_delay | int }}"
+#  delegate_to: localhost
```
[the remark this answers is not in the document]
taban03: The endpoint I'm checking is protected and requires JWT Token. Not sure how to do it, that's why I was commenting them out

taban03:
> Please see my comments if those make sense to you. Thanks Andrea!

Thank you Jack, taking a look on them right now

playbooks/roles/start/tasks/wait_for_zowe.yml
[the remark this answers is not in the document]
taban03: I can probably ping another unprotected endpoint, I will change it

balhar-jakub:
@stevenhorsman I expect the Caching service to be turned on by default approximately in 1.19 version of Zowe.

balhar-jakub:
@jackjia-ibm @stevenhorsman Is there any way to keep the Caching service turned off by default for the production while keeping it on for testing on Marist?

tests/sanity/test/apiml/test-caching-service-apis-via-gateway.js
taban03: Ok, I've added the parameter to the `playbooks/roles/verify/tasks/main.yml` as suggested by you, now it should be fine.

stevenhorsman:
> @jackjia-ibm @stevenhorsman Is there any way to keep the Caching service turned off by default for the production while keeping it on for testing on Marist?

I think the best way for now is Jack indicates - remove the ZOWE_CACHING_SERVICE_START entirely from instance.env (and make sure that things processing it look for a positive value of "true", rather than not "false" so that blank doesn't switch it on), but add it as true in `playbooks/roles/configure/tasks/configure_instance.yml` (similar to what Andrei did with ZWED_agent_jwt_token parms)

taban03:
> > @jackjia-ibm @stevenhorsman Is there any way to keep the Caching service turned off by default for the production while keeping it on for testing on Marist?
> 
> I think the best way for now is Jack indicates - remove the ZOWE_CACHING_SERVICE_START entirely from instance.env (and make sure that things processing it look for a positive value of "true", rather than not "false" so that blank doesn't switch it on), but add it as true in `playbooks/roles/configure/tasks/configure_instance.yml` (similar to what Andrei did with ZWED_agent_jwt_token parms)

Thanks Steven, I've pushed the changes, let me know if I got the suggestions correctly.

playbooks/roles/start/tasks/wait_for_zowe.yml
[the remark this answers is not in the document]
taban03: This check I added I think is slowing down the build. The reason is that the Caching service is starting after this sensible playbook is run, I do not know why.

playbooks/roles/configure/tasks/configure_instance.yml
```diff
@@ -60,5 +60,11 @@
     echo "ZWED_agent_jwt_fallback=true" >> "{{ zowe_instance_dir }}/instance.env"
     echo "ZWED_agent_jwt_enabled=true" >> "{{ zowe_instance_dir }}/instance.env"
 
+# If zowe_apiml_caching_service_start true,
+# caching service is deployed on marist along with the rest of API ML and its sanity tests are run
+- name: Add Caching Service start parameter to instance.env
+  raw: |-
+    echo "zowe_apiml_caching_service_start=true" >> "{{ zowe_instance_dir }}/instance.env"
```
stevenhorsman, line 67: You are injecting `zowe_apiml_caching_service_start` here, but the envar the the apiml use to know whether to start is `ZOWE_CACHING_SERVICE_START` so this won't work

taban03: Yes you're right, I will fix it

taban03:
Tests passed with the new changes: https://wash.zowe.org:8443/job/zowe-install-test/view/change-requests/job/PR-1783/39/consoleFull  and `ZOWE_CACHING_SERVICE_START` is now picked up correctly by the test with `true` value.

playbooks/roles/configure/tasks/configure_instance.yml
```diff
@@ -60,5 +60,11 @@
     echo "ZWED_agent_jwt_fallback=true" >> "{{ zowe_instance_dir }}/instance.env"
     echo "ZWED_agent_jwt_enabled=true" >> "{{ zowe_instance_dir }}/instance.env"
 
+# If zowe_apiml_caching_service_start true,
+# caching service is deployed on marist along with the rest of API ML and its sanity tests are run
+- name: Add Caching Service start parameter to instance.env
+  raw: |-
```
stevenhorsman, line 66: I think you can should the comment and instead add ` when: zowe_apiml_caching_service_start` so this is only injected if the ansible configure variable says so.

smpe/bld/smpe-split.sh
```diff
@@ -104,9 +104,10 @@ list="$list components/api-mediation/api-catalog-services.jar"  # pax01
 list="$list components/api-mediation/discoverable-client.jar"   # pax02
 list="$list components/api-mediation/discovery-service.jar"     # pax03
 list="$list components/api-mediation/gateway-service.jar"       # pax04
-list="$list components/app-server"                              # pax05
-list="$list components/files-api"                               # pax06
-list="$list components/jobs-api"                                # pax07
+list="$list components/api-mediation/caching-service.jar"       # pax05
```
stevenhorsman, line 107: I think this might the ptf generation. In the past there were checks to ensure that the correct number of pax segments were used. Have you run the ptf build and tests?

taban03: I didn't, I didn't have a clear clue when I changed this, cause the build was failing that time. I will try to run the ptf build

tests/sanity/test/apiml/test-caching-service-apis-via-gateway.js
```diff
@@ -0,0 +1,114 @@
...
+  expect(response.status).to.equal(204);
+};
+
+let getToken = async () => {
+  const uuid = testUtils.uuid();
+  testUtils.log(uuid, 'ZOWE_CACHING_SERVICE_START value: ' + process.env.ZOWE_CACHING_SERVICE_START);
+  return await testUtils.login(uuid);
+};
+
+describe('test caching service via gateway', function() {
```
stevenhorsman, line 39: A potentially neater way, rather than having an if is to use an conditional skip eg:
```
const cachingServiceEnabled = process.env.ZOWE_CACHING_SERVICE_START !== undefined && process.env.ZOWE_CACHING_SERVICE_START == 'true'
(cachingServiceEnabled ? describe : describe.skip)('test caching service via gateway', function() {
```
?

smpe/bld/smpe-split.sh
stevenhorsman: It looks like from https://wash.zowe.org:8443/view/Zowe/job/zowe-install-packaging/view/change-requests/job/PR-1783/32 this has broken the ptf build. Try undoing it and rerun the build

tests/sanity/test/apiml/test-caching-service-apis-via-gateway.js
taban03: it works, thanks! I will replace it

smpe/bld/smpe-split.sh
taban03: Done

playbooks/roles/configure/tasks/configure_instance.yml
taban03: Done

jackjia-ibm (approved):
Looks good to me! Thanks Andrea for all the fixes.

PR #1842, opened 2020-12-14
Bulk rename of zowe docker image from zowe-v1-lts to server-bundle

stevenhorsman:
- Cherry pick in change from #1828 into rc branch

PR #1851, opened 2020-12-16
Hardcode use of ompzowe as repo destination of docker, rather than using username

1000TurquoisePogs:
Currently a draft PR to see if this generates the right content.
This PR is to solve the issue in which a dockerhub organization can be pushed to by multiple users with permissions, but you can't login to an organization as the organization name.
Therefore, migrating ompzowe from a user to an organization requires a new user with permissions to push content to ompzowe, which results in the need to keep username and the term ompzowe separate.


1000TurquoisePogs:
I found that https://github.com/zowe/zowe-install-packaging/pull/1842 was hasty in that it ended up renaming directories rather than just the end result, so the building was broken.
This fixes that as well as the original issue about org name.

Jenkinsfile
```diff
@@ -210,15 +210,15 @@ sed -e 's#{BUILD_BRANCH}#${env.BRANCH_NAME}#g' \
              mkdir -p zowe-build/${env.BRANCH_NAME}_${env.BUILD_NUMBER} &&
              cd zowe-build/${env.BRANCH_NAME}_${env.BUILD_NUMBER} &&
              git clone --branch master https://github.com/zowe/zowe-dockerfiles.git
-             cd zowe-dockerfiles/dockerfiles/zowe-release/s390x/server-bundle &&
+             cd zowe-dockerfiles/dockerfiles/zowe-release/s390x/zowe-v1-lts &&
```
NakulManchanda, line 213: should this be `server-bundle` instead? @1000TurquoisePogs

1000TurquoisePogs: No because it's the folder structure here https://github.com/zowe/zowe-dockerfiles/tree/master/dockerfiles/zowe-release/s390x
The name of the image and the name of the directory have diverged.

jackjia-ibm:
Based on the changes I believe the PR should work. But since I didn't really run any test on it, I couldn't mark it as approved. I would suggest who did real test to do a proper review. `Jenkinsfile.release` is not critical here since we run promote/release from staging branch, we have chance to branch out and fix the problem if we notice failures after RC is confirmed.

I have 2 suggestions:

1. Would recommend we keep the folder name and image name converged. Maybe I didn't get the reason or benefit we keep the folder name to be zowe-v1-lts like before? Would it be ok to also rename the directory to be `server-bundle`?
2. Would like to suggest make `ompzowe` as a variable instead of hardcoded. When I verify the release process, I had concern to overwrite then ompzowe image, so I changed credential temporarily and published to my own account, just to test the process. We may still need this convenience in the future to verify our changes on release process.

Another issue would like to mention is tagging and it's out of scope of this PR. We need to manually tag the zowe-dockerfiles repo branch for a release since they are not in same repo of zowe-install-packaging, which will be tagged automatically when promote formal release. If we do want to keep this for several releases, I strongly recommend to merge zowe-dockerfiles into zowe-install-packaging to better branching sync and versioning.

stevenhorsman:
When running the tests parts of the APIML don't seem to work and the log shows:
```
 Exception in thread "main" java.lang.IllegalArgumentException: File /usr/include/java_classes/IRRRacf.jar must exist
            at org.springframework.boot.loader.data.RandomAccessDataFile$FileAccess.openIfNecessary(RandomAccessDataFile.java:234)
            at org.springframework.boot.loader.data.RandomAccessDataFile$FileAccess.<init>(RandomAccessDataFile.java:216)
            at org.springframework.boot.loader.data.RandomAccessDataFile$FileAccess.<init>(RandomAccessDataFile.java:206)
            at org.springframework.boot.loader.data.RandomAccessDataFile.<init>(RandomAccessDataFile.java:49)
            at org.springframework.boot.loader.jar.JarFile.<init>(JarFile.java:88)
            at org.springframework.boot.loader.archive.JarFileArchive.<init>(JarFileArchive.java:60)
            at org.springframework.boot.loader.archive.JarFileArchive.<init>(JarFileArchive.java:56)
            at org.springframework.boot.loader.PropertiesLauncher.getArchive(PropertiesLauncher.java:500)
            at org.springframework.boot.loader.PropertiesLauncher.getClassPathArchives(PropertiesLauncher.java:476)
            at org.springframework.boot.loader.PropertiesLauncher.getClassPathArchives(PropertiesLauncher.java:446)
            at org.springframework.boot.loader.Launcher.launch(Launcher.java:49)
            at org.springframework.boot.loader.PropertiesLauncher.main(PropertiesLauncher.java:593)
```
There are also other warnings eg the log directory being in the wrong place.

stevenhorsman:
I think this issue is blocking release candidate as well, which we are hoping to build tomorrow

1000TurquoisePogs:
"/usr/include/java_classes/IRRRacf.jar" would be another file which must be conditional else it's not going to work in docker. Hopefully it's fixed tomorrow?

1000TurquoisePogs:
Aside from the switch from the `-cp` command to the `-Dloader.path` command, use of IRRRacf existed for many versions prior and docker was working fine. Is this a docker problem and would the solution to be to revert to use of `-cp`? Just ideas.

https://github.com/zowe/api-layer/pull/581

https://github.com/zowe/api-layer/pull/1010

Seems like a minor change but don't know the reason/significance.

stevenhorsman:
I'm going to merge this into staging based on the fact that the staging version of the docker build is completely broken, so this is progress even if the docker image doesn't work and it doesn't impact the convenience, or ptf builds, then create a new issue for the test failure.

stevenhorsman (approved):

PR #1871, opened 2021-01-06
Zowe resource class and APIML resource

plavjanik:
Signed-off-by: Petr Plavjanik <petr.plavjanik@broadcom.com>

This PR introduces ZOWE resource class that will be used by security resources protecting sensitive information provided by Zowe.

It is done in the same way as ZLUX documentation suggests for ZLUX - https://docs.zowe.org/stable/user-guide/mvd-configuration.html#defining-the-racf-zowe-class.

It defines one resource for API ML that controls access to API with detailed information about API services.

#### PR type
- [x] Feature 

#### Relevant issues

Related to https://github.com/zowe/api-layer/issues/873

1. Define new resource class ZOWE (the name is hard-coded for now because ZLUX uses hardcoded value ZOWE) and resource in ZWESECUR
  - [x] RACF
  - [x] TSS
  - [x] ACF2
2. Updates workflows

- Yes
- [x] No

---
* Affected function: General security *

---
* Description: Zowe resource class introduced to control access to information about API services *

---
* Part: ZWESECUR  *

---
Zowe resource class has been introduced to control access to detailed information about API services.
If you need to enable access to this information to users of Zowe, you need to apply commands in the
section DEFINE ZOWE RESOURCE PROTECTION of ZWESECUR job.

Doc issue/PR number: https://github.com/zowe/docs-site/pull/1520


plavjanik:
@stevenhorsman, @OnnoVdT, @1000TurquoisePogs Hello Onno, Sean, and Steven, this is a draft of the ZWESECUR job that addresses ZOWE resource class. It is using the same RACF command that is documented for ZLUX https://docs.zowe.org/stable/user-guide/mvd-configuration.html#defining-the-racf-zowe-class. It defines the first resource for APIML. It is for the detailed service information that should not be opened without a SAF resource check.

The first version had the resource class configurable. Later, I have noticed that ZLUX has the class hardcoded. So it is hardcoded in the jobs too.

The ACF2 commands are not ready yet.

1000TurquoisePogs:
Do you already have code to read the class? Because, ZSS can tell you about any profile in the Zowe class if you want. It's limited to just the Zowe class because security reviewers didn't think it was a good idea to have a rest api that could allow broader access to esm info.

files/jcl/ZWENOSEC.jcl
```diff
@@ -203,6 +203,15 @@
   LISTGRP  &ADMINGRP. OMVS
   DELGROUP &ADMINGRP.
 
+/* REMOVE ZOWE RESOURCE CLASS ...................................... */
+
+/* remove profile in ZOWE class                                      */
+  RDELETE ZOWE APIML.SERVICES
+/* remove ZOWE class                                                 */
+  RDELETE CDT ZOWE
```
plavjanik, line 211: It might be already defined on the system before. In this case, we should remove it.

plavjanik:
> Do you already have code to read the class? Because, ZSS can tell you about any profile in the Zowe class if you want. It's limited to just the Zowe class because security reviewers didn't think it was a good idea to have a rest api that could allow broader access to esm info.

Hi Sean,

we are using two implementations: 
1. Java APIs - This has better performance, fewer possibilities to break, and makes it possible to have API ML without other components  
2. ZSS auth service - This will allow the usage of Docker containers for API ML with using ZSS for native z/OS APIs

I agree that limiting the service for ZOWE class is the right thing. Providing a generic API to check access to all SAF resources would allow users to do more than they can do now (unauthorized z/OS user cannot check access to SAF resources).

files/jcl/ZWENOSEC.jcl
```diff
@@ -203,6 +204,17 @@
   LISTGRP  &ADMINGRP. OMVS
   DELGROUP &ADMINGRP.
 
+/* REMOVE ZOWE RESOURCE CLASS ...................................... */
+/* uncomment commands to below if the ZOWE class has been created    */
+/* by the ZWESECUR job                                               */
+
+/* remove profile in ZOWE class                                      */
+/* RDELETE ZOWE APIML.SERVICES                                       */
+/* remove ZOWE class                                                 */
```
OnnoVdT, line 213: We worked with the RACF folks, and ZOWE will be a standard SAF class provided by z/OS, starting at v2.5, so there there is no deleting it anymore. Like with TSS, add a bigger comment to this sample command
```
/* uncomment commands to below if the ZOWE class has been created    */
/* by the ZWESECUR job                                               */
```

workflows/templates/ZWESECUR.vtl
```diff
@@ -400,6 +402,37 @@
   LISTGRP &HLQ.
   LISTDSD PREFIX(&HLQ.) ALL
 
+/* DEFINE ZOWE RESOURCE PROTECTION ................................. */
+
+/* - Defines new resource class for Zowe that protects access to     */
```
OnnoVdT, line 407: mention that the class might already exist, in which case this step is not required

OnnoVdT (changes requested):

files/jcl/ZWENOSEC.jcl
plavjanik: Hi Onno, thank you for the review. All three ESMs have these commands commented and preceded with an explanatory comment now.

workflows/templates/ZWESECUR.vtl
plavjanik: I have added it.

OnnoVdT (approved):

JirkaAichler (approved):
Haven't tested personally but everything looks good.

Are there some test results of the installation (this job running)?

Joe-Winchester (approved):

PR #1911, opened 2021-01-21
Migrate server-bundle build code into this repo

1000TurquoisePogs:

#### PR type 
- [ ] Bugfix 
- [ ] Feature 
- [X] Other... Please describe: Moving code from one repo to another

- zowe-dockerfiles contained the build instructions used for docker, but were executed in this repo. Moving the files here will allow them to be under the same version control. While migrating, I simplified the docker files by consolidating the s390x and amd64 arch folders by renaming the prereq image dockerfile by arch extension, and making the server-bundle dockerfile take a build argument for which arch is being built.

The previous folder name for building server-bundle was zowe-v1-lts, so now it is being renamed to just 'server-bundle' since the versioning will be tracked in this repo by tags and branches.

- [ ] Yes
- [X] No

I am testing to confirm if we need to use the node() argument in the jenkinsfile anymore, since it is the same node name as the overall jenkins build is using.


1000TurquoisePogs:
Since the zlinux build needs a branch to point to, I couldnt leave it as staging to run the test, so I substituted it for my branch temporarily, and then reverted. You can see the build successes at https://wash.zowe.org:8443/job/zowe-install-packaging/job/PR-1911/8/

Jenkinsfile
```diff
@@ -210,16 +210,16 @@ sed -e 's#{BUILD_BRANCH}#${env.BRANCH_NAME}#g' \
           """
              mkdir -p zowe-build/${env.BRANCH_NAME}_${env.BUILD_NUMBER} &&
              cd zowe-build/${env.BRANCH_NAME}_${env.BUILD_NUMBER} &&
-             git clone --branch master https://github.com/zowe/zowe-dockerfiles.git
-             cd zowe-dockerfiles/dockerfiles/zowe-release/s390x/zowe-v1-lts &&
+             git clone --branch ${env.BRANCH_NAME} https://github.com/zowe/zowe-install-packaging.git
```
jackjia-ibm, line 213: We can zip the containers folder and upload to zLinux server to avoid picking github url and branch name manually. This will always make sure the code on zLinux is always same as current build. If we checkout branch again, there is a chance the branch is updated during the build, then zLinux version has latest code, but the other docker has older version.

Also maybe we need to sanitize the branch name to make it safe to be a directory name, or create tmp folder with random alphanums.

Based on the process, making zLinux a Jenkins node could be better, so Jenkins can handle the folders, checkout etc. But anyway this way should work.

containers/server-bundle/Dockerfile
```diff
@@ -0,0 +1,102 @@
+ARG BUILD_PLATFORM
```
jackjia-ibm, line 1: Nice we have the ARG here to combine 2 Dockerfiles. Thanks Sean!

jackjia-ibm (changes requested):
Thanks a lot Sean. Also really like to see 2 Dockerfiles are merged into 1. I didn't validate the Dockerfile and utils which I assume they are same as other repo.

The only change I think we need is how to get the source code on zLinux. Thanks.

jackjia-ibm (approved):
LGTM, thanks Sean

PR #1939, opened 2021-02-03
Introduce size of the Cache in the environment vavriables.

balhar-jakub:
The caching service startup was failing because it needed to add two properties for a startup. The properties are added with the default values in this PR.


jackjia-ibm (approved):
Convenience test passed! https://wash.zowe.org:8443/view/Zowe/job/zowe-install-test/view/change-requests/job/PR-1939/5/

Thanks Jakub!

PR #1947, opened 2021-02-05
Add optional docker image with dependency sources

1000TurquoisePogs:
- [X] Necessary documentation (if appropriate) have been added / updated

#### PR type
- [ ] Bugfix
- [X] Feature

- [ ] Yes
- [X] No

This PR does a few things
* Updates the README of the server-bundle container
* Updates the dependency image to always run an update when building (to get the latest versions) and to use a newer version of nodejs
* Adds a new dockerfile to add sourcecode of dependencies into the server-bundle image, but as an optional alternative to the server-bundle image so you can have it with or without sources. This is needed for license compliance, but may be desirable to users too.


containers/server-bundle/Dockerfile.nodejava.amd64
```diff
@@ -3,20 +3,15 @@ FROM openjdk:8-jdk-slim
 RUN apt-get update && apt-get install -y --no-install-recommends \
 		ca-certificates curl wget \
     groff iproute2 moreutils pax && \
-	  rm -rf /var/lib/apt/lists/*
+    apt-get dist-upgrade -y
```
1000TurquoisePogs, line 6: This should protect against docker images being a little outdated

containers/server-bundle/Dockerfile.nodejava.amd64
```diff
@@ -42,7 +37,7 @@ RUN set -ex \
     gpg --keyserver pool.sks-keyservers.net --recv-keys "$key" ; \
   done
 
-ENV NODE_VERSION 12.18.4
+ENV NODE_VERSION 12.20.1
```
1000TurquoisePogs, line 40: We could go to v14, but when?

containers/utils/download-zowe-build.sh
```diff
@@ -16,10 +16,14 @@ cd /root/zowe
 
 if [ "$ZOWE_BUILD" = "latest" ]; then
   echo "Downloading most recent release ..."
+  apt-get update && apt-get install -y wget --no-install-recommends
   wget -q "https://zowe.jfrog.io/zowe/libs-release-local/org/zowe/[RELEASE]/zowe-[RELEASE].pax" -O "$ZOWE_TARGET_NAME"
+  apt-get purge -y wget
```
1000TurquoisePogs, line 21: we remove wget that we use during build time, but maybe its something we should keep in the image for use by plugins?

containers/utils/get-deb-sources.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+# SPDX-License-Identifier: MIT
```
swinslow, line 2: As the original author of an earlier version of this script (see https://gitlab.com/swinslow/contcompl/-/blob/master/take2/get-sources.sh) -- if you would like to mark this as the following instead, that would be fine with me:

```
SPDX-License-Identifier: EPL-2.0 OR MIT
```

Just offering that as an option, for better alignment with Zowe's own license  :)

jackjia-ibm (approved):
Looks good to me, no obvious issues identified, source builds are showing in artifactory. Will see how release goes.

PR #1973, opened 2021-02-22
Removing zlux install script (moving to components)

1000TurquoisePogs:
Uses https://github.com/zowe/zss/pull/258 and https://github.com/zowe/zlux-app-server/pull/146 to remove the zlux install script that instead should be done by the install scripts of the app-server and zss components.


1000TurquoisePogs:
This is ready for review now.
In addition to the automated testing, the manual testing I did was to install, review file contents, and run.
Basically, I wanted to see that we had the same files as before, with the same encoding, for the /components/zss and /components/app-server folders, and this looks to be true.

lchudinov (approved):

PR #1986, opened 2021-03-02
Assign zowe docker user to same group as mounted volumes

1000TurquoisePogs:

#### PR type
- [x] Bugfix 
- [ ] Feature 

Recent changes to the docker image to run as user=zowe presented an issue where mounted volumes might have group permissions that did not match the docker image due to the docker image not having the same groups.
However, users can control whether docker mounts are read or read-write regardless of file permissions, so if the volume is writable, zowe should be able to make use of group-level permissions. Therefore, this change assigns zowe to the group seen within each mount folder before starting the servers. This has no impact on the host system, its just aligning the container to be more compatible with the mounts. The alternative would have been to tell users to set up group ids before running zowe in order to match what exists out of the box on the container, which is unecessary work if it can be automated such as in this PR.

- [ ] Yes
- [x] No

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes to the Zowe Installer will be documented in this file.
 <!--Add the PR or issue number to the entry if available.-->
+## `1.20.0`
+
+### New features and enhancements
+- server-bundle docker image has been optimized to greatly reduce the size without impacting the content and usability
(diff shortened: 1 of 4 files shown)


jackjia-ibm (approved):

1000TurquoisePogs:
Looks like the test in the build was aborted because it doesnt apply to the docker one, which was run separately. So, merging this.

PR #2008, opened 2021-03-12
Signing docker container

1000TurquoisePogs:

#### PR type
- [ ] Bugfix 
- [x] Feature 

#### Relevant issues

Resolves https://github.com/zowe/zlux/issues/626

- Sign docker images according to the documentation found in https://docs.docker.com/engine/security/trust/

- [ ] Yes
- [x] No

This feature should not break existing use of docker because users can get content regardless of signed status when signing verification is off, which is what they do today. But if signing verification is turned on, the old unsigned images will not work while this one will.


Jenkinsfile.release
```diff
@@ -594,6 +603,11 @@ node('zowe-jenkins-agent-dind-wdc') {
           "docker manifest push ompzowe/server-bundle:latest\n" +
           "docker manifest create ompzowe/server-bundle:v${params.ZOWE_RELEASE_VERSION} ${manifestAmends}\n" +
           "docker manifest push ompzowe/server-bundle:v${params.ZOWE_RELEASE_VERSION}"
+        sh "export DOCKER_CONTENT_TRUST=1\n" +
+          "printf \"${DECRYPTCODE}\n\" | docker trust sign ompzowe/server-bundle:amd64\n" +
+          "printf \"${DECRYPTCODE}\n\" | docker trust sign ompzowe/server-bundle:s390x\n" +
+          "printf \"${DECRYPTCODE}\n\" | docker trust sign ompzowe/server-bundle:latest\n" +
```
1000TurquoisePogs, line 609: I dont know if all of these can be combined into one command. I will try.

Jenkinsfile.release
```diff
@@ -594,6 +603,11 @@ node('zowe-jenkins-agent-dind-wdc') {
...
+          "printf \"${DECRYPTCODE}\n\" | docker trust sign ompzowe/server-bundle:v${params.ZOWE_RELEASE_VERSION}"
```
1000TurquoisePogs, line 610: I dont know if you need to sign meta tags or if they are signed due to the signing of the original amd64 and s390x images. I will try to see the difference and update the ticket.

Jenkinsfile.release
```diff
@@ -566,11 +566,20 @@ node('zowe-jenkins-agent-dind-wdc') {
...
-      )]) {
+      withCredentials([
+        usernamePassword(
+          credentialsId: 'ZoweDockerhub',
+          usernameVariable: 'USERNAME',
+          passwordVariable: 'PASSWORD'
+        ),
+        sshUserPrivateKey(
+          credentialsId: 'DockerUserSigningKey',
+          keyFileVariable: 'USERSIGNKEY',
```
1000TurquoisePogs, line 577: i want to write the key to a file in a specific place docker uses, so i will try using sshuserprivatekey as the store first, but may change do a different credential store if it isnt the right type to store 3 values in one object: key, keyname, and passphrase.

Jenkinsfile.release
```diff
@@ -566,11 +566,20 @@ node('zowe-jenkins-agent-dind-wdc') {
...
+          passphraseVariable: 'SIGNPASSPHRASE',
+          usernameVariable: 'SIGNKEYNAME'
+        )
+      ]) {
+        sh "echo \"${USERSIGNKEY}\" > ~/.docker/trust/private/${SIGNKEYNAME}.key"
```
1000TurquoisePogs, line 582: the keyname appears to be important for docker to be able to find the right key to use, so keeping the name intact is important, as is the location of the file.

jackjia-ibm (approved):
Thanks Sean

jackjia-ibm (approved):
I didn't test it because just realized it will update Docker image latest tag. We can merge and fix any issues found when we promote. The `Jenkinsfile.release` is always executed from `staging` branch, so we can fix any issues without affecting release branches (rc/master).

PR #2039, opened 2021-03-26
Change the endpoint used for getting version of API ML

balhar-jakub:
The version endpoint was copied to a location /application/version to prevent routing to another instance for the information.


jackjia-ibm (approved):
`curl -k https://host:7554/application/version` works well for me. Thanks Jakub!

PR #2098, opened 2021-04-19
Add env vars to test zss-thru-apiml behavior

1000TurquoisePogs:
This is just a test to see if we can fix the ZWED151W error about zss-thru-apiml timeouts occurring https://wash.zowe.org:8443/view/Zowe/job/zowe-install-test/view/change-requests/job/PR-2094/12


1000TurquoisePogs:
Closing this as complete except that the timeout value is still configurable and the value may be too low by default (30seconds) so we can tweak that if needed.

PR #2161, opened 2021-05-26
Fix portability of swagger support plugins

1000TurquoisePogs:
Without relativeTo plugins for our two new meta-plugins that add proper swagger support, the plugins break if ROOT_DIR is changed.
With the addition of the relativeTo pointers here, the plugins become portable just as our other plugins are, so this is the behavior we want.


NakulManchanda (approved):

1000TurquoisePogs:
Merging this in to fix everything.

NakulManchanda (approved):

PR #2173, opened 2021-05-31
use non-default jwtsecret label to catch more issues

jackjia-ibm:


jackjia-ibm:
Test passed https://wash.zowe.org:8443/job/zowe-install-test/job/PR-2173/1/console

PR #2195, opened 2021-06-11
Test for cookie issues in app server

lchudinov:
Test for cookie issues in app server


balhar-jakub:
@lchudinov Could we close this PR?

PR #2199, opened 2021-06-14
Record ZIS params in instance

1000TurquoisePogs:
This PR is a simplification of a part of https://github.com/zowe/zowe-install-packaging/pull/1756 which only deals with how to get ZIS parameters into instance.env

These things are known at install time, but could be customized post-install.
Therefore, though we can automate getting the data into instance.env by recording install output into a temp file, and reading it back at instance creation time.
However, because these params can be customized, automation not be desired.
So, there are arguments to specify them explicitly.

Also, the instance creation script argument parsing was reworked to allow for more flexibility in input. The behavior is backwards compat.

How to test:
If you use zowe as you normally did, you should end up with LOADLIB and PARMLIB ZIS parameters in instance.env when you create a new instance. This doesnt update old instances.
However, if for some reason you customize or things go wrong with the temp file logic, you can explicitly specify loadlib, parmlib, or just the dsn prefix by the -l, -p, and -d flags to the configure instance script.

For example

Failure or what is seen on an instance upgrade:
```
ZWES_ZIS_LOADLIB=
ZWES_ZIS_PARMLIB=
ZWES_ZIS_PARMLIB_MEMBER=ZWESIP00
```

Success when no parameters are given and automation works
```
ZWES_ZIS_LOADLIB=myprefix.SZWEAUTH
ZWES_ZIS_PARMLIB=myprefix.SAMPLIB
ZWES_ZIS_PARMLIB_MEMBER=ZWESIP00
```

Success when parameters were specified
`/bin/zowe-configure-instance.sh -c inst/zis10 -l abc -p def`

```
ZWES_ZIS_LOADLIB=abc
ZWES_ZIS_PARMLIB=def
ZWES_ZIS_PARMLIB_MEMBER=ZWESIP00
```


1000TurquoisePogs:
Note: recording these params in instance.env is not essential, so if a failure occurs the zowe setup can still continue. In fact, these parameters can be added to instance.env at any later time manually.

DivergentEuropeans (commented):
This case doesn't work for me:
Success when parameters were specified
`/bin/zowe-configure-instance.sh -c inst/zis10 -l abc -p def`
ZWES_ZIS_LOADLIB=abc
ZWES_ZIS_PARMLIB=def
ZWES_ZIS_PARMLIB_MEMBER=ZWESIP00
I used
`./zowe-configure-instance.sh -c ~/release3/zowe-instance/ -l abc -p def` and I don't see those 3 generated in my instance.env

1000TurquoisePogs:
what do you see?

1000TurquoisePogs:
ZWES_ZIS_PARMLIB_MEMBER=ZWESIP00 will be there unconditionally

1000TurquoisePogs:
so if you dont see it, whats up?
New

DivergentEuropeans:
For some reason, I see none of these parameters at all. It doesn't seem to be editing the instance.env
```
Checking existing /u/ts6321/release3/zowe-instance//instance.env for updated properties
No updates required
```
Maybe that explains why my instance.env isn't updating? Should I delete it and retry?

DivergentEuropeans:
I did delete it, and it generated a new one, and it didn't have those properties there. Not sure here what's going on or if I'm doing something wrong

DivergentEuropeans (approved):
I was testing this, improperly last time. Works well for me

DivergentEuropeans (changes requested):
This needs to replace uses of LOADLIB with PLUGINLIB

PR #2209, opened 2021-06-25
Fix default zlux server json

lchudinov:
This PR fixes zowe/zlux#658

#### PR type

- [x] Bugfix 

#### Relevant issues
Fixes zowe/zlux#658

This PR fixes the issue when `ZOWE_ZSS_SERVER_TLS=true` but zss still runs on https port because the https port setting comes from default `server.json` file.

- [ ] Yes
- [x] No


1000TurquoisePogs:
I think node.https, node.mediationLayer, agent.http, agent.https, privilegedServerName, and implementationDefaults could all be removed and would help protect against accidental outdated configuration causing bugs in both servers. instance.env and convert-env.sh handles all of these for the most part.

1000TurquoisePogs (approved):
This should fix common errors about server.json being out of sync, by just having less content there.

PR #2211, opened 2021-07-06
Create dataset for ZIS plugins during installation procedure

timgerstel:
Signed-off-by: Timothy Gerstel <tgerstel@rocketsoftware.com>

#### PR type
- [ ] Bugfix 
- [X] Feature 

#### Relevant issues

Fixes https://github.com/zowe/zlux/issues/668

- This pull request implements the following feature:
- Create dataset for ZIS plugins during Zowe install procedure such that ZIS datasets are separated from ZIS plugin datasets.

- [ ] Yes
- [X] No

---

Zowe MVS Installation

---
- Create dataset for ZIS plugins

---

---
_multi-line description_

Doc issue/PR number:


scripts/zowe-install-MVS.sh
```diff
@@ -133,6 +133,22 @@ else
     echo "  $script failed to create ${ZOWE_DSN_PREFIX}.SZWESAMP, RC=$rc" >> $LOG_FILE
 fi
 
+# 3. {datasetprefix}.SZWEPLUG
+
+tsocmd "delete '${ZOWE_DSN_PREFIX}.SZWESAMP' " >> ${LOG_FILE} 2>&1
```
1000TurquoisePogs, line 138: why is a ticket labeled create doing a delete?
i can't think of a reason this is here so please explain.

scripts/zowe-install-MVS.sh
```diff
@@ -133,6 +133,22 @@ else
...
+
+# TODO replace by allocate-dataset.sh call to resuse VOLSER support
+tsocmd "allocate new da('${ZOWE_DSN_PREFIX}.SZWEPLUG') " \
+    "dsntype(library) dsorg(po) recfm(u) lrecl(0) bLKSIZE(32760)" \
```
1000TurquoisePogs, line 142: that seems correct, to basically have it be identical in attribute to SZWEAUTH. do you have a test that shows this has been successfully created?

1000TurquoisePogs (changes requested):
I don't think a delete operation should be in this PR, since the point of making a plugin dataset is that it would be the one dataset that never gets deleted during an upgrade.

1000TurquoisePogs (approved):

PR #2216, opened 2021-07-13
Revert "Create dataset for ZIS plugins during installation procedure"

1000TurquoisePogs:
Reverts zowe/zowe-install-packaging#2211


1000TurquoisePogs:
We  discovered that this solution wont work with SMPE. What we really want to do is create this dataset post-SMPE involvement so that it doesnt get tracked by SMPE. But, that requires moving this logic into a different stage of scripting, perhaps a new one, so we'll leave that for another time.

Issue #2170, opened 2021-05-31
Containerization Spike: investigate certification requirements


jackjia-ibm:
There are several certification programs out in market. Investigate what requirements/rules are applicable for Zowe.

- OCI compliance – should we care about it, or anything special rules we need to follow
- [Docker Certified publisher and images](https://docs.docker.com/docker-hub/publish/certify-images/)
- [Red Hat Container Certification](https://connect.redhat.com/en/partner-with-us/red-hat-container-certification)
- [IBM Cloud Pak Certification](https://playbook.cloudpaklab.ibm.com/requirements-by-type/) extract items we need to follow

Expected Result: checklist of rules in these categories
- OCI compliance
- Docker Certified images
- Red Hat Container Certification
- IBM Certification
- Zowe conformance recommendation
- Zowe conformance requirement

CarsonCook:
# OCI Specifications

Workflow that allows users to ru nan image with no additional arguments.

An OCI implementation downloads the OCI image and unpacks it into an OCI Runtime filesystem bundle and is ran by the OCI runtime.

Summary of requirements:
* [Image manifest](#image-manifest) that references image contents and dependencies.
* [Image layers](#image-layer) that describe the container filesystem.
* [Image configuration](#image-configuration) that describes how an image is executed as a container.
* [Container configuration](#container-configuration) that describes how a container implements standard operations.
* Adherence to specific [container lifecycle requirements](#lifecycle).
* Support of [standard container operations](#operations).

## Image Specification

Contains sufficient information to launch the applicaton on the target platform. Create an OCI image that contains an image manifest, layer serialization, and image configuration.

### Image Index

Higher level manifest that points to specific image manifests ideal for certain platforms. OPTIONAL.

Follow the below structure. Fields are required unless notes otherise. Further details and other optional fields are [here](https://github.com/opencontainers/image-spec/blob/master/image-index.md).
```json
{
    "schemaVersion": 2, // must be 2
    "manifests": [ // May be an empty array
        {
            "mediaType": "application/vnd.oci.image.manifest.v1+json",
            "size": 123, // number of bytes of the raw content
            "digest": "sha256:hex", //digest of the content
            "platform": { // the platform field/object is optional
                "architecture": "ppc64", // Required if platform object exists - the CPU architecture the image is built to run on, should use architecture values listed here: https://golang.org/doc/install/source#environment
[...]

PR #2220, opened 2021-07-14
Users/jack/containerization conformance

### LINKED ISSUE #2171: Containerization Spike: component image conformance
Discuss and propose conformance related to Zowe containerization.

General rules for all Zowe components
- Base images (compliance, license)
- Folder structure (image and runtime)
- Start command
- Volume sharing
- Configuration & Secrets
- Preinstalled Linux tools
- Versioning recommendations

Special rules for component with plugin capability – AppServer
- Persistent Volume
- How Plugins should be imported

How to define zowe-install-packaging image:
- Folder structure
- Volume sharing

jackjia-ibm:

#### PR type
- [ ] Bugfix 
- [ ] Feature 
- [x] Other... Please describe: documentation

#### Relevant issues

Fixes #2171

- containerization conformance draft

- [ ] Yes
- [x] No


containers/conformance.md
```diff
@@ -0,0 +1,220 @@
...
+### Base Image
+
+You are free to choose base image based on your requirements.
+
+For Zowe core components, here are our recommendations:
+
+- [Red Hat Universal Base Image 8 Minimal](https://developers.redhat.com/articles/ubi-faq?redirect_fragment=resources#ubi_details)
+- [Ubuntu](https://hub.docker.com/_/ubuntu)
+
+The image should be as small as possible.
```
1000TurquoisePogs, line 20: consider rephrasing:
The image should contain as few software packages as possible for security, and be as small as possible such as by reducing package count and layers.

containers/conformance.md
```diff
@@ -0,0 +1,220 @@
...
+
+### Multi-CPU Architecture
+
+- Zowe core components must release images based on both `amd64` and `s390x` CPU architecture.
+- Zowe core component images must use multiple manifests to define the image supports multiple CPU architecture.
+- CPU architecture like `amd64` and `s390x` must NOT appear in released component image tag.
```
1000TurquoisePogs, line 26: why not? it helps in certain cases and its not a problem with aliases
latest =  latest-ubuntu = 1.23-ubuntu = 1.23-ubuntu-amd64 = latest-amd64
i think the real *MUST* is components must use manifest when in a registry, so that you can alias latest-amd64 and latest-s390x to just "latest".

containers/conformance.md
```diff
@@ -0,0 +1,220 @@
...
+- 1.23.0-alpine.pr-1234
+- 1.23.0-ubi.users-john-test1
+
+### Files and Directories
+
+This is required folder structure for all Zowe components:
+
+```
+/licenses
+/app
```
1000TurquoisePogs, line 88: can we call it "component" instead? our documentation will get very confusing about what an app is, and what a component is. it looks like a component to me since it has manifest.json within.

containers/conformance.md
```diff
@@ -0,0 +1,220 @@
...
+
+These environment variable(s) must be set as a fixed value in the image:
+
+- `ZOWE_COMPONENT_ID`: this is the Zowe component ID for current image. For example: `ENV ZOWE_COMPONENT_ID=gateway`.
+
+### User `zowe`
+
+In the Dockerfile, a `zowe` user and group must be created. Example command `RUN groupadd -r zowe && useradd --no-log-init -r -g zowe zowe`.
+
+`USER zowe` must be specified before the first `CMD` or `ENTRYPOINT`.
```
1000TurquoisePogs, line 117: i had a problem with doing this in tech preview. volume mounts had certain user:group permissions that either required the person running the container to have matched the uid/gid up with what the container would have OR have the container initially start as root, reassign the container's zowe uid to be within (from container viewpoint) the gid seen in the volume, and THEN switch to user zowe to run zowe. therefore, i couldnt do `USER zowe` but did run zowe as user zowe, after initialization as root.

containers/conformance.md
```diff
@@ -0,0 +1,220 @@
...
+
+This section is mainly for information. No actions required for components except where it's specified explicitly.
+
+Below sections are mainly targeting Kubernetes or OpenShift environment. Starting Zowe containers in Docker environment with `docker-compose` is in plan and may change some of the requirements.
+
+### General rules
+
+**Components MUST:**
+
+- NOT be started as root user in the container.
```
1000TurquoisePogs, line 133: above concern

containers/conformance.md
```diff
@@ -0,0 +1,220 @@
...
+- listen on ONLY one port in the container.
+- NOT rely on hardcoded directory names like `/app`.
+- be cloud vendor neutral and must NOT rely on features provided by specific cloud vendor.
+- NOT rely on host information such as `hostIP`, `hostPort`, `hostPath`, `hostNetwork`, `hostPID` and `hostIPC`.
+- MUST accept either `instance.env` or `zowe.yaml` as configuration file, same as running on z/OS.
```
1000TurquoisePogs, line 138: or just plain env vars?

containers/conformance.md
```diff
@@ -0,0 +1,220 @@
...
+
+Below sections are mainly targeting Kubernetes or OpenShift environment. Starting Zowe containers in Docker environment with `docker-compose` is in plan and may change some of the requirements.
+
+### General rules
+
+**Components MUST:**
+
+- NOT be started as root user in the container.
+- listen on ONLY one port in the container.
+- NOT rely on hardcoded directory names like `/app`.
```
1000TurquoisePogs, line 135: above it says /app is required

containers/conformance.md
```diff
@@ -0,0 +1,220 @@
...
+- be cloud vendor neutral and must NOT rely on features provided by specific cloud vendor.
+- NOT rely on host information such as `hostIP`, `hostPort`, `hostPath`, `hostNetwork`, `hostPID` and `hostIPC`.
+- MUST accept either `instance.env` or `zowe.yaml` as configuration file, same as running on z/OS.
+
+### Files and Directories
+
+In runtime, the Zowe content are organized in this structure:
+
+```
+/zowe
```
1000TurquoisePogs, line 145: tech preview had `/home/zowe`. should we in order to keep the existing documentation or will we just change all that doc anyway

1000TurquoisePogs (commented):
Giving my 2 cents on what you have so far

Another thing to note is image signing. It appears to be something only supported by dockerhub (or maybe any proper registry?) but gives that extra level of security.

containers/conformance.md
jackjia-ibm: We can use same `/home/zowe` directory as tech preview.

containers/conformance.md
jackjia-ibm: Yes /app is required but the application should not assume the runtime is /app. That's what it means here. I will try to rephrase it.

containers/conformance.md
jackjia-ibm: If we move to v2, I'm thinking to remove instance.env support. So try not to surprise the extenders here. They should expect zowe.yaml could be used as configuration which aligns to zOS deployment.

containers/conformance.md
jackjia-ibm: This may cause trouble if we want to get certification. Will keep this in mind and try if I can solve the issue.

The issue could be caused by in tech preview, the workspace volume is touched off container. We should avoid that. All changes to persistent volume should come from container.

But we may need to align uid and gid for all images. Will add this part.

containers/conformance.md
jackjia-ibm: Sounds good idea, will change.

containers/conformance.md
jackjia-ibm: This is for formal release image. For dev/test purpose, I guess it's ok.

containers/conformance.md
jackjia-ibm: Sounds great. Will change.

containers/conformance.md
1000TurquoisePogs: unless there's some special utility i am not aware of, what i noticed is that if you migrate a z/os instance to linux (and i'm working on a utility to make that easy) or even if you just make a new directory, that directory is owned by someone other than the docker user.
lets say, user "rob" with id=100, in group "dev", gid=200
on the linux system, you see `rwxrwxr-x rob:dev instance`
in docker, you see `rwxrwxr-x 100:200 instance` but the zowe user is for example, id=1001 in gid=1002
so, the zowe user wouldnt be able to write to the instance, which can be a problem.

in that scenario, the only solution i found was to tell the user to manually `chmod` or `chown` to have the uid/gid match a known and hardcoded container uid/gid. I had picked 8544 to match the app-server port, and was going to tell users to make sure to do a `chown -R youruser:8544 instance` before running... it could work, but automatic config is better, which is what i ended up doing by
1) startup as root
2) read gid
3) add zowe to group in container (no effect on host system, except that now zowe can read the instance it otherwise couldnt)
4) run as zowe

containers/conformance.md
```diff
@@ -0,0 +1,227 @@
+# Zowe Conformance Criteria Related to Containerization
+
+**DRAFT**
+
+These conformance criteria are applicable for all Zowe components intending to run in a containerized environment. The containerized environment could be Kubernetes or OpenShift running on Linux or Linux on Z. This may also apply to `docker-compose` running on Linux, Windows, Mac OS, or zCX.
+
+## Image
+
+In general, the image should follow [Best practices for writing Dockerfiles](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/). Below requirements are in addition to the list.
```
CarsonCook, line 9: The image SHOULD follow, or MUST follow?

containers/conformance.md
```diff
@@ -0,0 +1,227 @@
...
+
+This is the required folder structure for all Zowe components:
+
+```
+/licenses
+/component
+  +- README.md
+```
+
+- `/licenses` folder holds all license related files. It should include at least the license information for current application. It's recommended to include a license notice file for all pedigree dependencies. All licenses files must be in UTF-8 encoding.
```
CarsonCook, line 92: It SHOULD include at least...or it MUST include at least...

containers/conformance.md
```diff
@@ -0,0 +1,227 @@
...
+
+```
+/component
+  +- manifest.json or manifest.yaml
+  +- /bin/<lifecycle-scripts>
+  +- <other-application-files>
+```
+
+- `/component/manifest.(json|yaml)` is recommended for Zowe components. The format of this file is defined at [Zowe component manifest](https://docs.zowe.org/stable/extend/packaging-zos-extensions/#zowe-component-manifest). Components must use the same manifest file as when it's running on z/OS.
+- `/component/bin/<lifecycle-scripts>` should remain the same as what it is when running on z/OS. Components must use the same lifecycle scripts when it's running on z/OS.
```
CarsonCook, line 105: SHOULD remain the same... or MUST remain the same

containers/conformance.md
```diff
@@ -0,0 +1,227 @@
...
+
+### Environment Variable(s)
+
+These environment variable(s) must be set as a fixed value in the image:
+
+- `ZOWE_COMPONENT_ID`: this is the Zowe component ID for current image. For example: `ENV ZOWE_COMPONENT_ID=gateway`.
+
+### User `zowe`
+
+In the Dockerfile, a `zowe` user and group must be created. The `zowe` user `UID` and group `GID` should be defined as `ARG` and with default value `UID=20000` and `GID=20000`. Example commands:
```
CarsonCook, line 115: SHOULD be defined as...or MUST be defined as

containers/conformance.md
```diff
@@ -0,0 +1,227 @@
...
+      +- /workspace
+    +- /keystore
+      +- zowe-certificates.env
+```
+
+- `/home/zowe/runtime` is a shared volume initialized by the `zowe-launch-scripts` container.
+- `/home/zowe/extension/<component-id>` is a symbolic link to the `/component` directory. `<component-id>` is `ZOWE_COMPONENT_ID` defined in `ENV`.
+- `/home/zowe/instance/(instance.env|zowe.yaml)` is a Zowe configuration file and MUST be mounted from a ConfigMap.
+- `/home/zowe/keystore/zowe-certificates.env` is optional if the user is using `instance.env`. If this configuration exists, it MUST be mounted from a ConfigMap.
+- Any confidential environment variables, for example, a Redis password, in `instance.env` or `zowe.yaml` should be extracted and stored as Secrets. These configurations must be imported back as environment variables.
```
CarsonCook, line 166: SHOULD be extracted and stored as Secrets...or MUST be extracted...

containers/conformance.md
```diff
@@ -0,0 +1,227 @@
...
+
+- The `zowe-launch-component` image contains necessary scripts to start Zowe components in the Zowe context.
+- This image has a `/home/zowe` directory and it will be shared and mounted to all Zowe component containers as `/home/zowe/runtime`.
+- In Kubernetes and OpenShift environments this step is defined with [`initContainers` specification](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/).
+
+### Zowe Workspace Directory
+
+- Zowe workspace directory `/home/zowe/instance/workspace` will be defined as a persistent volume.
+- Components writing to this directory should be aware of the potential conflicts of same-time writing by multiple instances of the same component.
+- Components writing to this directory should NOT write container specific information to this directory as it may potentially overwritten by another container.
```
CarsonCook, line 185: SHOULD NOT write container specific...or MUST NOT write container specific

containers/conformance.md
jackjia-ibm: Will keep it as SHOULD here. I assume even if they don't follow those best practices, it won't affect us to get certification.

CarsonCook: As of now it won't affect docker certification because they don't really have requirements. I suppose it could affect us in the future, however.

jackjia-ibm: If that's the case, we may need to import the rules here explicitly just in case Docker modify the best practices page. Will check the potential overlaps.

1000TurquoisePogs:
Going down the epic list to find tasks I can assist with and just adding note here: May need to say more about how components containing plugins should be written to be conformant. If I learn more I will either suggest improvements here or in another PR.

containers/conformance.md
```diff
@@ -0,0 +1,235 @@
...
+  * support both `amd64` and `s390x` architecture.
+- [Red Hat Universal Base Image 8 Minimal](https://developers.redhat.com/articles/ubi-faq?redirect_fragment=resources#ubi_details)
+- [Ubuntu](https://hub.docker.com/_/ubuntu)
+
+The image should contain as few software packages as possible for security, and should be as small as possible such as by reducing package count and layers.
+
+### Multi-CPU Architecture
+
+- Zowe core components must release images based on both `amd64` and `s390x` CPU architecture.
+- Zowe core component images must use multiple manifests to define if the image supports multiple CPU architecture.
```
balhar-jakub, line 29: What do you mean by manifest here?

containers/conformance.md
```diff
@@ -0,0 +1,235 @@
...
+```
+### Required Labels 
+LABEL name="APPLICATION NAME" \
+      maintainer="EMAIL@ADDRESS" \
+      vendor="COMPANY NAME" \
+      version="VERSION NUMBER" \
+      release="RELEASE NUMBER" \
+      summary="APPLICATION SUMMARY" \
+      description="APPLICATION DESCRIPTION" \
+```
```
balhar-jakub, line 46: Do we need to provide the actual Dockerfile? As we don't actually use it per se. We would need to grab it from JIB.

containers/conformance.md
jackjia-ibm: It's the docker manifest https://docs.docker.com/engine/reference/commandline/manifest/.

containers/conformance.md
jackjia-ibm: We do need the Dockerfile(s) used to build components, it will be useful to inspect license, generate sources image, or do some scans. But for Dockerfile of base images, we don't really need.

Do you mean you don't have the Dockerfile if you are using JIB to create the images?

containers/conformance.md
```diff
@@ -0,0 +1,235 @@
...
+Here are our recommendations for core Zowe components:
+
+- Zowe base images:
+  * base on Ubuntu and Red Hat Universal Base Image,
+  * provide common dependencies including JDK and/or node.js,
+  * support both `amd64` and `s390x` architecture.
+- [Red Hat Universal Base Image 8 Minimal](https://developers.redhat.com/articles/ubi-faq?redirect_fragment=resources#ubi_details)
+- [Ubuntu](https://hub.docker.com/_/ubuntu)
+
+The image should contain as few software packages as possible for security, and should be as small as possible such as by reducing package count and layers.
```
jackjia-ibm, line 24: Make more specific requirements on base images:
- must be bash
- xargs, etc
- what about (o)ping, (o)netstat, (chtag), etc. - may make them all pre-reqs for base images they choose.

containers/conformance.md
```diff
@@ -0,0 +1,235 @@
...
+```
+
+- `/component/manifest.(json|yaml)` is recommended for Zowe components. The format of this file is defined at [Zowe component manifest](https://docs.zowe.org/stable/extend/packaging-zos-extensions/#zowe-component-manifest). Components must use the same manifest file as when it's running on z/OS.
+- `/component/bin/<lifecycle-scripts>` must remain the same as what it is when running on z/OS.
+
+### Environment Variable(s)
+
+These environment variable(s) must be set as a fixed value in the image:
+
+- `ZOWE_COMPONENT_ID`: this is the Zowe component ID for current image. For example: `ENV ZOWE_COMPONENT_ID=gateway`.
```
jackjia-ibm, line 118: Maybe we can require manifest and extract from manifest `.name`.

containers/conformance.md
balhar-jakub: We don't have one per se. We can provide one if needed as this is basically the Docker image:
https://github.com/GoogleContainerTools/jib/blob/master/docs/faq.md#what-would-a-dockerfile-for-a-jib-built-image-look-like

jackjia-ibm: I see. This rule comes from one of the RedHat Certification criteria, https://github.com/zowe/zowe-install-packaging/issues/2170#issuecomment-857948519. It's unclear if Dockerfile is required or not, maybe labels in image is also acceptable.

This rule isn't critical for starting APIML in Zowe context. So if JIB can fit in other conformance rules listed here, we can change the rule to `These descriptive labels are required in the Dockerfile or image:`.

PR #2276, opened 2021-08-03
Fix temporary directory issue

1000TurquoisePogs:
#### PR type
- [x] Bugfix 

In the previous release, a temporary file was recorded during install time (not spme) which was used at configure time to attempt to automate the recording of ZIS parameters for use in plugin install automation.
The problem is that the temp file was stored in a directory such that if multiple users were doing installs and configures of Zowe (seemingly exclusively dev situation) the permissions of the directory might prevent a new temp file from being made.
This PR fixes the issue by just not creating a temp directory, and instead just places the temp file directly under `/tmp`. This should avoid write permission issues, but allow read permissions to continue to be the deciding factor on if this automation works.

- [ ] Yes
- [x] No


lchudinov:
@1000TurquoisePogs  I get `ls: FSUM6785 File or directory "/tmp/zowe--install-*" is not found` with the pax.

lchudinov (approved):
Works fine to me now.

PR #2282, opened 2021-08-09
Us766623/psi trigger

pavelzlatnik:
Jenkins Trigger modification to support creation of PSI for ZOWE together with SMPE creation 

#### PR type
- [ ] Bugfix 
- [x] Feature 

Modification of Jenkins trigger to support also PSI creation from SMPE environment
(Now only with dummy PSI script which just echo message. Real script that creates PSI would be added after check that pipeline is working as expected).
- 

- [ ] Yes
- [X] No

### DIFF
--- a/.pax/post-packaging.sh
+++ b/.pax/post-packaging.sh
@@ -23,7 +23,7 @@ set -x
 
 function call_PSIcreate {  
   
-    echo "called function to create PSI from trigger"  
+    echo "Called function to create PSI from trigger."  
     cd ${CURR_PWD}/psi/bld
     chmod 775 ./prepare-PSI.sh
     ./prepare-PSI.sh
--- a/.pax/post-packaging.sh
+++ b/.pax/post-packaging.sh
@@ -21,6 +21,16 @@
 #######################################################################
 set -x
 
+function call_PSIcreate {  
+  
+    echo "called function to create PSI"  
+    cd ${CURR_PWD}/psi/bld
+    chmod 775 ./prepare-PSI.sh
+    ./prepare-PSI.sh
+  
+  return 0
+ }
+
 SCRIPT_NAME=$(basename "$0")
 CURR_PWD=$(pwd)
 
@@ -267,6 +277,13 @@ mv "${SMPE_BUILD_SHIP_DIR}/${SMPE_PROMOTE_TAR}" ${SMPE_PROMOTE_TAR}
 # leave fixed name for promote.tar to simplify automated processing during promote
 echo "mv fmid.zip ${SMPE_FMID_ZIP}" > rename-back.sh.1047
 iconv -f IBM-1047 -t ISO8859-1 rename-back.sh.1047 > rename-back.sh
+
+#PSI creation from SMPE
+run_PSI_create=true
+
+if [ "$run_PSI_create" = true ] ; then
+    call_PSIcreate 
+fi
 
 echo "[$SCRIPT_NAME] done"
 
--- a/.pax/post-packaging.sh
+++ b/.pax/post-packaging.sh
@@ -279,9 +279,9 @@ echo "mv fmid.zip ${SMPE_FMID_ZIP}" > rename-back.sh.1047
 iconv -f IBM-1047 -t ISO8859-1 rename-back.sh.1047 > rename-back.sh
 
 #PSI creation from SMPE
-run_PSI_create=true
 
-if [ "$run_PSI_create" = true ] ; then
+echo "PSI build is set to : [$BUILD_PSI]"
+if [ "$BUILD_PSI" == "yes" ]; then
     call_PSIcreate 
 fi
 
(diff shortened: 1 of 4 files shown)


psi/bld/prepare-PSI.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
```
OnnoVdT, line 1: I see you're creating a new 'root' directory, psi. These dirs need special considerations to ensure they do not show up in the pax version. I advise to make psi a subdir of the smpe directory, which is still logically correct since you plan on using SMPE as base for PSI.

.pax/post-packaging.sh
```diff
@@ -268,6 +278,13 @@ mv "${SMPE_BUILD_SHIP_DIR}/${SMPE_PROMOTE_TAR}" ${SMPE_PROMOTE_TAR}
 echo "mv fmid.zip ${SMPE_FMID_ZIP}" > rename-back.sh.1047
 iconv -f IBM-1047 -t ISO8859-1 rename-back.sh.1047 > rename-back.sh
 
+#PSI creation from SMPE
+run_PSI_create=true
```
OnnoVdT: You need to know where the SMPE build placed the rel-fales & SMPMCS to use as input for the PSI creation. Like the SMPE build, you also need to have a config file to manage where the PSI work area is, and where the output goes. 
Instead of driving a new script from post-packaging.sh, PSI creation could be triggered from smpe.sh with a startup argument (test "$run_PSI_create" = true && opt="$opt -S"). Doing the invocation here allows you to reuse get-config.sh to get the SMPE variables (and append it to add your own).

The way I envision it, if smpe.sh receives the -S argument, it executes 2 additional scripts, smpe-install-smpe.sh to install the smpe image, and smpe-psi.sh to create the PSI from this image. Finally post-packaging.sh can pick up the psi file and stage it for shipment to artifactory.

The PSI build could use the SMPE config setup for this, and instead of driving a new script from post-packaging.sh, PSI creation could be triggered from smpe.sh with a

OnnoVdT:
If I read the Git data correctly, you're trying to merge with master instead of staging, so that needs fixing.

OnnoVdT:
Are we building PSI only for GA, or do you envision a PSI on each PTF build as well? The latter is easier for new customers, but does require us to install GA and then apply the PTF when staging PSI input data during a PTF build.

OnnoVdT (changes requested):

psi/bld/prepare-PSI.sh
```diff
@@ -0,0 +1,15 @@
...
+#version=1.0
+
+ 
+function PSIcreation {  
+  
+    echo "Called function to create PSI in our script."
```
jackjia-ibm, line 7: Sorry I didn't start review the PR since this looks like dummy function for now. Are we planning to implement this function in future PRs? or will improve with this PR?

pavelzlatnik:
> If I read the Git data correctly, you're trying to merge with master instead of staging, so that needs fixing.

I will make new branch against staging in that case

psi/bld/prepare-PSI.sh
pavelzlatnik: ok I will make it under smpe folder

.pax/post-packaging.sh
pavelzlatnik: I have tried that based on this comment in new branch. Just was not sure about variable run_PSI_create (this was dummy one from initial commit, not part of pullReq), so I have used $BUILD_PSI" == "yes" which I have added to Jenkins trigger in later commit. Also I was not sure about smpe-install-smpe.sh, did you mean by that smpe-install.sh script, or some new one which does not exist now?

psi/bld/prepare-PSI.sh
pavelzlatnik: We have whole repository in our environment which is building PSI for ZOWE, and we want to somehow incorporate it to zowe environment, namely this sript https://github.gwd.broadcom.net/MFD/PSI-automation/blob/master/ZOWE/create.sh
image

.pax/post-packaging.sh
pavelzlatnik: > You need to know where the SMPE build placed the REL-files & SMPMCS to use as input for the PSI creation. Like the SMPE build, you also need to have a config file to manage where the PSI work area is, and where the output goes.
> Instead of driving a new script from post-packaging.sh, PSI creation could be triggered from smpe.sh with a startup argument (test "$run_PSI_create" = true && opt="$opt -S"). Doing the invocation here allows you to reuse get-config.sh to get the SMPE variables (and append it to add your own).
> 
> The way I envision it, if smpe.sh receives the -S argument, it executes 2 additional scripts, smpe-install-smpe.sh to install the smpe image, and smpe-psi.sh to create the PSI from this image. Finally post-packaging.sh can pick up the psi file and stage it for shipment to artifactory.

There, https://github.com/zowe/zowe-install-packaging/pull/2296#issuecomment-901222902

pavelzlatnik:
Created new branch against staging branch with changes based on feedback 
https://github.com/zowe/zowe-install-packaging/pull/2296

PR #2366, opened 2021-09-29
Add missing scripts

1000TurquoisePogs:

#### PR type
- [x] Bugfix 

Certain useful scripts for app-server and zss are not currently available in the instance directory. This could lead to users finding the scripts in component directories, using them, and then possibly later having a problem if we decide to move or rename files within component directories, since they are internal & undocumented.
This attempts to fix the problem by putting useful scripts into the instance, where we will not make name or location changes that will cause issues unless there is some kind of major version change. 

- [ ] Yes
- [x] No

I don't like the location of install-app.sh, as people should have components and use the component installer instead of install-app. But, I can't move it right now, it would be a breaking change for some.

Also, the component installers should be here too. But I don't know, the scripts may need adjustment to live in the instance location.


jackjia-ibm (approved):
Thanks Sean

PR #2523, opened 2021-12-13
update app server, zss var names consistently

1000TurquoisePogs:
This PR is from https://github.com/zowe/zowe-install-packaging/pull/2478 but made to resolve the many conflicts.


1000TurquoisePogs:
I'll coordinate fixes on whatever breaks after merging this.

Issue #2608, opened 2022-01-28
Init certificates can fail in automation


1000TurquoisePogs:
Posted for an extender trying out v2.

Try running `zwe init certificates` in an environment that has few env vars such as automation, and you may find this result:
```
zwe init certificate --allow-overwritten --update-config -c /myzowe/zowe.yaml
-------------------------------------------------------------------------------
>> Creating certificate authority "local_ca"

>>>> Generate PKCS12 format local CA with alias local_ca:
  * Exit code: 127
  * Output:
    keytool: pkeytool 4: pkcs12_create_certificate_authority 18: /myzowe/bin/commands/certificate/pkcs12/create/ca/index.sh 40: .: zwecli_process_command 4: /myzowe/bin/zwe 96: .: zwecli_inline_execute_command 9: /myzowe/zowe_runtime/bin/commands/init/certificate/index.sh 198: .: zwecli_process_command 4: /myzowe/zowe_runtime/bin/zwe 96: FSUM7351 not found
Error ZWEL0168E: Failed to create certificate authority "local_ca".
```

When I look at the code I see: https://github.com/zowe/zowe-install-packaging/blob/v2.x/staging/bin/libs/certificate.sh#L35
keytool comes from $JAVA_HOME/bin, but we call it directly here. Likely the cause of failure is that PATH didnt contain it.
I imagine the fix is just to reference $JAVA_HOME when possible.

jackjia-ibm:
Please validate if this is still an issue in v2.0.0-TP2. I made some changes to check whether java is available on certain commands.

jackjia-ibm:
Please reopen if this is still an issue.

PR #2609, opened 2022-01-28
Fix certificate init by checking keytool location

1000TurquoisePogs:

#### PR type
- [x] Bugfix 

#### Relevant issues
Fixes https://github.com/zowe/zowe-install-packaging/issues/2608

Small check for if we can really use `keytool` or need to reference `JAVA_HOME` to find it instead

- [x] No


bin/libs/certificate.sh
```diff
@@ -28,11 +28,23 @@ ZWE_PRIVATE_DEFAULT_CERTIFICATE_VALIDITY="3650"
...
+if [ "${keytool_on_path}" != "true" ]; then
+  if [ -n "${JAVA_HOME}" ]; then
+    keytool_bin="${JAVA_HOME}/bin/keytool"
+  else
+    print_message "WARNING: keytool cannot be found, please put on PATH or define JAVA_HOME"
+  fi
+else
+  keytool_bin=keytool
+fi
+
```
jackjia-ibm, line 42: This way will introduce 2 new variables and be seen by components.

Another though on this, we can put `require_java` into `zwe certificate pkcs12` sub-commands. `zwe init certificate` will call `zwe certificate` commands behind the scene. `require_java` will check JAVA_HOME and make sure it's in PATH, otherwise it will fail with error message.

What do you think?

Issue #2435, opened 2021-10-28
Spike: propose v2 changes and discuss with community


jackjia-ibm:

jackjia-ibm:
Initial [Zowe v2 Install Proposals.pptx](https://github.com/zowe/zowe-install-packaging/files/7513201/Zowe.v2.Install.Proposals.pptx) be discussed

jackjia-ibm:
Updated version of [Zowe v2 Install Proposals.pptx](https://github.com/zowe/zowe-install-packaging/files/7636262/Zowe.v2.Install.Proposals.pptx) as of Dec 1st, 2021

PR #2640, opened 2022-02-07
Improved zwe command help format and add examples

### LINKED ISSUE #2574: zwe positional statement syntax unclear
When I executed `./zwe init --update-config --allow-overwritten apfauth` I thought I would just be doing the `apfauth` step, but instead the entire init was done.
The correct command appears to be  `./zwe init apfauth --update-config --allow-overwritten`
`zwe init --help` doesn't show what the syntax should be, so it seems like we need to enhance the help a bit, or enhance the argument reading. one or the other.

In `man` output, command format is the first line shown. We could follow that standard.

Such as if the help said:

```
  Format

  zwe init [sub-command] [option [option]...]

  Example

  zwe init certificate -v -l /my/log/output


------------------
Global parameters:
  --help|-h: boolean, optional
    Display this help.
  --debug|--verbose|-v: boolean, optional
    Enable verbose mode.
  --trace|-vv: boolean, optional
    Enable trace level debug mode.
  --log-dir|--log|-l: string, optional
    Write logs to this directory.
  --config|-c: string, optional
    Path to Zowe configuration zowe.yaml file.

------------------
Parameters for command "init":
  --allow-overwritten: boolean, optional
    Allow overwritten existing MVS data set.
  --skip-security-setup: boolean, optional
    Whether should skip security related setup.
  --security-dry-run: boolean, optional
    Whether to dry run security related setup.
  --update-config: boolean, optional
    Whether to update YAML configuration file with initialization result.

------------------
Available sub-command(s):
  - apfauth
  - certificate
  - mvs
  - security
  - stc
  - vsam
```

adityadr-joy:

#### PR type
- [x] Bugfix 
- [ ] Feature 

#### Relevant issues
Fixes #2574 

- Updated zwe help format 
- Added synopsis 
- Added examples
- Improved parameter formatting 

- [ ] Yes
- [x] No

---

---

---

---
_multi-line description_

Doc issue/PR number:


bin/commands/.help
```diff
@@ -1,3 +1,9 @@
-A command line utility helps you managing Zowe instance.
+Synopsis
+    zwe [sub-command [sub-command]...] [parameter [parameter]...]
```
jackjia-ibm, line 2: I believe `Synopsis` section can be automatically generated.

This variable `${ZWE_CLI_COMMANDS_LIST}` is the sub-command list. Then we can always append `[parameter [parameter]...]` after it.

jackjia-ibm (commented):
Seems you introduced new `.examples` file, could you also update `bin/README.md` to reflect the changes https://github.com/zowe/zowe-install-packaging/tree/v2.x/staging/bin#command-assistant-files. Thanks.

bin/commands/.help
jackjia-ibm: Also I feel `Name` and `Synopsis` seems duplicated info to me. What do you think?

bin/commands/sample/sub/.help
```diff
@@ -1 +1,7 @@
-A sample sub-command.
+Synopsis
+    zwe sample sub [sub-command] [parameter [parameter]...]
```
1000TurquoisePogs, line 2: Question for @jackjia-ibm actually: shouldn't we hide this from an end-user? Extensions aren't going to add commands to `zwe` right, so this command only helps us core zowe devs. May be better as documentation in a readme or something.

bin/commands/version/.help
```diff
@@ -1 +1,6 @@
-Display Zowe version.
+Synopsis
+    zwe version [parameter]...
```
1000TurquoisePogs, line 2: is there a parameter? I don't see one

1000TurquoisePogs (changes requested):
2 help issues I saw.

bin/commands/sample/sub/.help
jackjia-ibm: With original idea, I do think we may allow extenders to expand `zwe` command. I remember JoeW also thought about this. The thought was listed in the initial proposal https://github.com/zowe/zowe-install-packaging/issues/2435#issuecomment-983931109 page 3. But it's not priority and we have so many topics, so we didn't really discuss details.

1000TurquoisePogs: Ok, we should talk about this more sometime later. I'm unsure how much we should do here without just telling people to use the Zowe CLI.

bin/commands/version/.help
adityadr-joy: version doesn't have its own parameters but it can be executed with global parameters like 
zwe version -h
zwe version -v

bin/commands/certificate/pkcs12/.help
```diff
@@ -1 +1,6 @@
-Manage PKCS12 format keystore and truststore.
+Synopsis
+    zwe certificate pkcs12 [sub-command] [parameter [parameter]...]
```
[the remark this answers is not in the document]
adityadr-joy: Resolved it

bin/commands/.help
adityadr-joy: You are right that `Name` and `Synopsis` seems duplicate for deep commands.

But In most of the man or help pages command title is always displayed first before any other explanation so Instead of keeping it under `Name` heading it can be displayed individually.

Let me know if it looks redundant to you I can remove it.

bin/commands/.help
```diff
@@ -1,3 +1,5 @@
-A command line utility helps you managing Zowe instance.
+Description
```
jackjia-ibm, line 1: After removing `Synopsis` from `.help`, i guess we don't need this `Description` line. What do you think?

bin/commands/.help
jackjia-ibm: It looks good to me. Thanks Aditya!

bin/commands/.help
adityadr-joy: Right, Fixed it.

jackjia-ibm (approved):
Awesome work! Thanks a lot Aditya!

1000TurquoisePogs (approved):

PR #2651, opened 2022-02-10
use zowe.extensionDirectory from config instead of target-dir parameter for zwe components install 

### LINKED ISSUE #2633: zwe component install requires target-dir
```
> ./zwe components install -c ~/zowe.yaml --help
Zowe server command: zwe components install

Install a Zowe module.

------------------
Global parameters:
  --help|-h: boolean, optional
    Display this help.
  --debug|--verbose|-v: boolean, optional
    Enable verbose mode.
  --trace|-vv: boolean, optional
    Enable trace level debug mode.
  --log-dir|--log|-l: string, optional
    Write logs to this directory.
  --config|-c: string, optional
    Path to Zowe configuration zowe.yaml file.

------------------
Parameters for command "components install":
  --target-dir|--target|-d: string, required
    Target directory where the component will be installed.
  --component-file|--component|-o: string, required
    Path to the component package or directory.
  --auto-encoding|-e: string, optional
    If we want to automatically tagging the module files.

------------------
Available sub-command(s):
  - extract
  - process-hook

> ./zwe components install -c ~/zowe.yaml --component /mycomponent
Error ZWEL0106E: target-dir parameter is required
```

It seems that target-dir is supposed to be the extensionDir, is that right?
Then, this seems like a bug in that zowe.yaml contains extensionDir, so specifying target-dir separately seems unnecessary.
Or, am I supposed to use a different command to install components?

adityadr-joy:

#### PR type
- [ ] Bugfix 
- [ ] Feature 

#### Relevant issues

Fixes #2633 

`zwe components install` command gets target directory parameter from zowe.yaml instead of target-dir parameter. 

- [ ] Yes
- [ ] No

---

---

---

---
_multi-line description_

Doc issue/PR number:

### DIFF
--- a/bin/commands/components/.examples
+++ b/bin/commands/components/.examples
@@ -1,1 +1,1 @@
-    zwe components install -d path/to/target/dir -o /path/to/component/package
+    zwe components install -c /path/to/zowe.yaml -o /path/to/component/package
--- a/bin/commands/components/install/.errors
+++ b/bin/commands/components/install/.errors
@@ -1,1 +1,2 @@
 ZWEL0156E|156|Component name is not initialized after extract step.
+ZWEL0180E|180|Zowe extension directory (zowe.extensionDirectory) is not defined in Zowe YAML configuration file.
--- a/bin/commands/components/install/.examples
+++ b/bin/commands/components/install/.examples
@@ -1,4 +1,4 @@
-    zwe components install -d path/to/target/dir -o /path/to/component/package
+    zwe components install -c /path/to/zowe.yaml -o /path/to/component/package
 
-    zwe components install extract -d /path/to/target/dir -o /path/to/component/package
+    zwe components install extract -c /path/to/zowe.yaml -o /path/to/component/package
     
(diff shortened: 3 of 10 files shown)


DivergentEuropeans (dismissed):
From this code, it looks like you removed support for the target-dir arg completely i.e. if you don't have extensionsDir present in your config, you throw an error. Shouldn't it default to zowe.yaml config & then check for command line args (like this one) when they're present or when zowe.yaml not found?

DivergentEuropeans (commented):
Didn't mean to put official "Request changes"

1000TurquoisePogs:
> From this code, it looks like you removed support for the target-dir arg completely i.e. if you don't have extensionsDir present in your config, you throw an error. Shouldn't it default to zowe.yaml config & then check for command line args (like this one) when they're present or when zowe.yaml not found?

No. At runtime, the servers only have the value of extensionsDir to find components. So, if you were able to specify anything else via the target-dir parameter, you could install a component into a place that will not be known to zowe and so it will be as if it does not exist.

1000TurquoisePogs (approved):
This looks good to me. Check with Jack to ensure this is the right way to require config and extract extensionDir from it, but this looks correct to me.

jackjia-ibm (approved):
Looks good to me too. Thanks a lot Aditya!

PR #2691, opened 2022-02-25
Users/jack/container workspace dir

jackjia-ibm:
Sort out workspace mount point.


containers/conformance.md
```diff
@@ -173,29 +173,30 @@ In the runtime, the Zowe content is organized in this structure:
...
         +- /<component-id>
     +- /instance
-      +- instance.env or zowe.yaml
+      +- zowe.yaml
       +- /logs
-      +- /tmp
       +- /workspace
     +- /keystore
-      +- zowe-certificates.env
+      +- keystore.p12
```
1000TurquoisePogs, line 180: I don't think these files should be guaranteed to be here.
What we could say is "keystore content will be in the /keystore directory"
whether thats a p12, a cer, or both is really up to the users and would be specified in the zowe.yaml

containers/kubernetes/workloads/api-catalog-deployment.yaml
```diff
@@ -100,20 +100,20 @@ spec:
...
-              mountPath: "/home/zowe/logs"
+              mountPath: "/home/zowe/instance/logs"
             - name: zowe-workspace
-              mountPath: "/home/zowe/workspace"
+              mountPath: "/home/zowe/instance/workspace"
       initContainers:
         - name: init-zowe
           # image: zowe-docker-release.jfrog.io/ompzowe/zowe-launch-scripts:2-ubuntu
-          image: zowe-docker-snapshot.jfrog.io/ompzowe/zowe-launch-scripts:2.0.0-ubuntu.v2-x-staging
+          image: zowe-docker-snapshot.jfrog.io/ompzowe/zowe-launch-scripts:2.0.0-ubuntu.users-jack-container-workspace-dir
```
1000TurquoisePogs, line 116: keeping this or going back to v2.x/staging?

containers/conformance.md
jackjia-ibm: Agreed. changes wording.

containers/kubernetes/workloads/api-catalog-deployment.yaml
jackjia-ibm: It's for testing purpose. Will be reverted before merge.

PR #2705, opened 2022-03-01
Tie registration of appfw plugins to component enabled status

1000TurquoisePogs:
Fixes the bug of api catalog not appearing on first startup, as well as long time bug of explorer ui visibility in desktop not being tied to whether or not they're running. This is fixed by moving plugin registration code into appfw (https://github.com/zowe/zlux-app-server/pull/201) and removing it from zowe install packaging.

### DIFF
--- a/bin/libs/component.sh
+++ b/bin/libs/component.sh
@@ -350,21 +350,6 @@ process_component_appfw_plugin() {
       all_succeed=false
       break
     fi
-    appfw_plugin_id=$(read_json "${appfw_plugin_path}/pluginDefinition.json" ".identifier")
-    if [ -z "${appfw_plugin_id}" ]; then
-      print_error "Cannot read identifier from App Framework plugin ${appfw_plugin_path}/pluginDefinition.json"
-      all_succeed=false
-      break
-    fi
-
-    # copy to workspace/app-server/pluginDirs
-    appfw_plugin_workspace_path="${ZWE_zowe_workspaceDirectory}/app-server/pluginDirs/${appfw_plugin_id}"
-    mkdir -p "${appfw_plugin_workspace_path}"
-    cp -r "${appfw_plugin_path}/." "${appfw_plugin_workspace_path}/"
-
-    # install app
-    "${ZWE_zowe_runtimeDirectory}/components/app-server/share/zlux-app-server/bin/install-app.sh" "${appfw_plugin_workspace_path}"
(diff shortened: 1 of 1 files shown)


jackjia-ibm (changes requested):
This will not work in containerization use case for few reasons:

1. app-server container cannot reach other extensions/plugins runtime directory when running in container. `<workspace>/app-server/pluginsDir` is the only reliable place to find plugin code.
2. plugins could be applied/started much later than app-server. Which means when app-server is starting, the plugin may not in `<workspace>/app-server/pluginsDir` either.

This is similar to APIML static vs dynamic registration. Ideally the plugins should do `dynamic` registration.

One problem I think we may have is pluginsDir is using plugin ID, which isn't directly related to the component name in zowe.yaml.

1000TurquoisePogs:
We need to visit containers in a future PR.

The problem this and other PRs address is appfw did not previously track component enable status, so it didnt know whether to show or hide a plugin.

The current fix is to have appfw do plugin registration/deregistration every startup by iterating through every component folder to read its manifest to find which plugins exist, and register if enable=true, else deregister.

The existing container plugin-copy code is not compatible with these changes, because this register/deregister code needs to see the component folders to relate component enabled status to plugin registration status.

I am missing some information to come up with a fix for containers.
This might work:
1. Detect if we're running in a container or not, to
2. Conditionally copy entire component folders into a place equivalent to "extensionDirectory". Ideally, literally extensionDirectory would be best but if needed, `workspaceDirectory/app-server` can be utilized again. I just figure other components probably will want to see extensions in the future, so app-server specific hacks here don't seem appropriate.
But depending on when `process_component_appfw_plugin` is run, there are other solutions.


Also note on your comments:

>     2. plugins could be applied/started much later than app-server. Which means when app-server is starting, the plugin may not in `<workspace>/app-server/pluginsDir` either.
app-server can scan its plugin registration folder at a user-defined interval of time, so there can be a solution that accounts for startup time differences.
 
> This is similar to APIML static vs dynamic registration. Ideally the plugins should do `dynamic` registration.
APIML registration is purely a network operation. App-server serves files. I'm not seeing how this is related. File copies must take place in the container scenario.

> One problem I think we may have is pluginsDir is using plugin ID, which isn't directly related to the component name in zowe.yaml.
Component names are not unique, plugin IDs are. The app-server and zss do not see components, they see plugins, so the component inspection is only really needed to find where plugins are and whether they should be registered or deregistered.

PR #2737, opened 2022-03-17
Replace home=/tmp with home=/u/&USER

1000TurquoisePogs:
Since multiple people can use /tmp, using /tmp as a home directory can result in unpredictable behavior, as seen with a recent end user in which it caused zowe to never start and instead have a loop of shell script spawning.

I also think there doesn't need to be an AUXUSER section at all, since AUX is never directly started, but is started from ZIS typically. But, perhaps a fix for a different PR.


1000TurquoisePogs:
I haven't tested this change yet and I'm not sure if I can set the home dir to one that doesn't exist prior but I guess the worst case is people need to set HOME to a unique place not shared by other users.

PR #2747, opened 2022-03-30
Remove doing a workspace copy of plugins when on zos

1000TurquoisePogs:
In TP2 we added logic to app-server which will call install-app or not depending upon component enable status.
Containers work a bit differently and for that, we were using the logic that I'm making conditional in this PR.
If this logic were not conditional, a copy would be made, wasting space and disabling the PC bit for zss plugins, and install-app would be pointing to a misleading location on z/os.

### DIFF
--- a/bin/libs/component.sh
+++ b/bin/libs/component.sh
@@ -339,33 +339,23 @@ process_component_appfw_plugin() {
     # apply values if appfw_plugin_path has variables
     appfw_plugin_path=$(parse_string_vars "${appfw_plugin_path}")
     appfw_plugin_path=$(cd "${appfw_plugin_path}"; pwd)
-
-    if [ ! -r "${appfw_plugin_path}" ]; then
-      print_error "App Framework plugin directory ${appfw_plugin_path} is not accessible"
-      all_succeed=false
-      break
-    fi
     if [ ! -r "${appfw_plugin_path}/pluginDefinition.json" ]; then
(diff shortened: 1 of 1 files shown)


1000TurquoisePogs:
See https://github.com/zowe/zowe-install-packaging/pull/2705 for conversation

bin/libs/component.sh
```diff
@@ -339,33 +339,23 @@ process_component_appfw_plugin() {
...
-    appfw_plugin_workspace_path="${ZWE_zowe_workspaceDirectory}/app-server/pluginDirs/${appfw_plugin_id}"
-    mkdir -p "${appfw_plugin_workspace_path}"
-    cp -r "${appfw_plugin_path}/." "${appfw_plugin_workspace_path}/"
-
-    # install app
-    "${ZWE_zowe_runtimeDirectory}/components/app-server/share/zlux-app-server/bin/install-app.sh" "${appfw_plugin_workspace_path}"
-    # FIXME: do we know if install-app.sh fails. if so, we need to set all_succeed=false
+    if [ "${ZWE_RUN_ON_ZOS}" != "true" ]; then
+      # for containers, copy to workspace/app-server/pluginDirs and run install-app. on zos, this is done at startup.
+      appfw_plugin_workspace_path="${ZWE_zowe_workspaceDirectory}/app-server/pluginDirs/${appfw_plugin_id}"
```
jackjia-ibm, line 350: Seems the line of assigning value to `appfw_plugin_id` is removed. So it could be empty here.

PR #2764, opened 2022-04-08
Add server common schema and id property to server base

1000TurquoisePogs:
Very small PR, just aligning schema content to what was discussed with Jack and in https://github.com/zowe/docs-site/pull/2108


1000TurquoisePogs:
I got confirmation from Jack that this is good to merge in.

PR #2787, opened 2022-04-19
Added rbacProfileIdentifier and cookieIdentifier parameter in config and schema

aditya-ranshinge:
Signed-off-by: Aditya Ranshinge <aranshinge@rocketsoftware.com>

- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Necessary documentation (if appropriate) have been added / updated
- [x] DCO signoffs have been added to all commits, including this PR

#### PR type
What type of changes does your PR introduce to Zowe? _Put an `x` in the box that applies to this PR. If you're unsure about any of them, don't hesitate to ask._ 
- [x] Bugfix 
- [ ] Feature 
- [ ] Other... Please describe:

#### Changes proposed in this PR

Added rbacProfileIdentifier and cookieIdentifier parameter in config and schema

PR #2803, opened 2022-04-21
2787 for RC

1000TurquoisePogs:
https://github.com/zowe/zowe-install-packaging/pull/2787 but for RC


1000TurquoisePogs:
This is necessary for the behavior of cookies and RBAC of ZSS & app-server in RC2, so I'm merging this in

PR #2822, opened 2022-04-26
Users/jack/smpe build

jackjia-ibm:
This should fix extattr missing issue when copying from extract to staging folder.


OnnoVdT:
$extract = $stage comment:
smpe-isntall.sh:180-191 - add $extract = $stage test to verify product install
```
# verify everything is installed
if test $removeInstall -eq 1
then
  if test "$extract" = "$stage"
  then
    echo "-- SKIP verifying product install" 
  else
    echo "-- verifying product install"
    orphan=$(find $extract ! -type d)
    if test "$orphan"
    then
      echo "** ERROR $me not all files are moved to $stage or $mvsI"
      echo "$orphan"                      # quotes preserve line breaks
      test ! "$IgNoRe_ErRoR" && exit 8                           # EXIT
    fi    #
  fi    # $extract != $stage
fi    # $removeInstall
```

jackjia-ibm:
[Onno Van den Troost](https://app.slack.com/team/W46QNNWT1)  [3:30 PM](https://ibm-systems-z.slack.com/archives/DHM91NSMN/p1651001445439609)
I now see why extract and stage are separate
the SMPE files are added via another pax, which must be extracted and post-processed. Initially there were 3 input files (node if I recall correctly), but that one was removed before GA of the SMPE version.
In v1, both zowe.pax and smpe.pax leave trash behind, so the installed data was copied to a clean slate, $stage. In v2, zowe.pax is clean, but smpe.pax is not. if we install with $extract = $stage then the smpe.pax trash is packaged as well (install script and 2 empty directories)
[3:35](https://ibm-systems-z.slack.com/archives/DHM91NSMN/p1651001728165529)
so if we wnat to prevent the copy of installed zowe.pax, I must either rework smpe.pax to not leave trash, and ensure it doesn't overwrite anything the install of zowe.pax already laid down (troublesome), or rework so that only smpe.pax uses an extract dir and then copy to stage

[Jack Jia](https://app.slack.com/team/WAXAJKE91)  [3:36 PM](https://ibm-systems-z.slack.com/archives/DHM91NSMN/p1651001760948049)
no need to difficult changes for now
3:36
i think redundant copying is not ideal, but it’s ok, as long as the copied result is correct

Onno Van den Troost  3:36 PM
agreed, for now I would just add a comment on why we copy
will send you that as review comment
have to drop for a minute to pick up the kids

PR #2832, opened 2022-05-02
Add ZIS plugin install automation as part of zwe component install

DivergentEuropeans:
Signed-off-by: Leanid Astrakou <lastrakou@rocketsoftware.com>

#### PR type
- [ ] Bugfix 
- [x] Feature 

#### Relevant issues

Fixes 

- 

- [ ] Yes
- [x] No

How to Test?

1. Get ZIS plugin (one with /loadlib and /samplib)
2. Construct your component to include ZIS plugin (see documentation for ZIS plugin schema here: https://github.com/zowe/docs-site/blob/6f533c7ac3e3d0cb3ff83a5a8b29626abe051a86/docs/appendix/server-component-manifest.md - see `zisPlugins`)
3. Run zwe components install on component with ZIS plugins


AdarshdeepCheema (approved):

bin/libs/component.sh
```diff
@@ -408,6 +408,154 @@ process_zss_plugin_install() {
   fi
 }
 
+process_zis_plugin_install() {
+  if [ "${ZWE_RUN_ON_ZOS}" = "true" ]; then
+    zwes_zis_pluginlib=$(read_yaml "${ZWE_CLI_PARAMETER_CONFIG}" ".zowe.setup.dataset.authPluginLib")
+    zwes_zis_parmlib=$(read_yaml "${ZWE_CLI_PARAMETER_CONFIG}" ".zowe.setup.dataset.parmlib")
+    zwes_zis_parmlib_member="ZWESIP00"
```
1000TurquoisePogs: very important TODO here. we have a schema file now, https://github.com/zowe/zowe-install-packaging/blob/v2.x/master/schemas/zowe-yaml-schema.json
No more undocumented, hardcoded things.
Find a place for this parameter and add it in the schema and example file https://github.com/zowe/zowe-install-packaging/blob/v2.x/master/example-zowe.yaml so that it is visible and configurable by users.
I suggest, zowe.setup.dataset.parmlibMembers.zis=ZWESIP00

1000TurquoisePogs (changes requested):

bin/libs/component.sh
```diff
@@ -408,6 +408,154 @@ process_zss_plugin_install() {
...
+            if [ "$num" != "" ]; then
+              remove_key_value_at_line_number "$num" "$parmlib_member_as_unix_file"
+              add_key_value_at_end_of_file "$samplib_key_value" "$parmlib_member_as_unix_file"
+              changed=1
+            else
+              # The key doesn't exist. Just add the key-value pair to the end of the file.
+              add_key_value_at_end_of_file "$samplib_key_value" "$parmlib_member_as_unix_file"
+              changed=1
+            fi
+          fi
```
jackjia-ibm: Think this whole section can be wrapped into a function `update_uss_parmlib_key(parmlib_file, key, value)`.

bin/libs/component.sh
```diff
@@ -408,6 +408,154 @@ process_zss_plugin_install() {
...
+  fi
+
+  changed=0
+
+  if [ -d "${component_dir}${plugin_path}" ]; then
+    if [ -d "${component_dir}${plugin_path}/loadlib" ] && [ -d "${component_dir}${plugin_path}/samplib" ]; then
+      for module in $(ls ${component_dir}${plugin_path}/loadlib); do
+        copy_uss_to_mvs "${component_dir}${plugin_path}/loadlib/${module}" "$zwes_zis_pluginlib"
+        if [ $? != 0 ]; then
+          exit 1
```
jackjia-ibm: We shouldn't `exit` but should write error message and `return`, then let caller to decide what to do.

bin/libs/component.sh
```diff
@@ -408,6 +408,154 @@ process_zss_plugin_install() {
...
+    zwes_zis_parmlib_member="ZWESIP00"
+    print_trace "- Checking for zis plugins and verifying them"
+    component_dir="${1}"
+    
+    iterator_index=0
+    zis_plugin_id=$(read_component_manifest "${component_dir}" ".zisPlugins[${iterator_index}].id" 2>/dev/null)
+    zis_plugin_path=$(read_component_manifest "${component_dir}" ".zisPlugins[${iterator_index}].path" 2>/dev/null)
+    while [ -n "${zis_plugin_path}" ]; do
+      cd "${component_dir}"
+      zis_plugin_install "${zis_plugin_path}" "${zwes_zis_pluginlib}" "${zwes_zis_parmlib}" "${zwes_zis_parmlib_member}" "${zis_plugin_id}" "${component_dir}"
```
jackjia-ibm: The caller should check whether the function fails or not, display error messages, and instruct what the user can do.

bin/libs/component.sh
```diff
@@ -408,6 +408,154 @@ process_zss_plugin_install() {
...
+      iterator_index=`expr $iterator_index + 1`
+      zis_plugin_path=$(read_component_manifest "${component_dir}" ".zisPlugins[${iterator_index}].id" 2>/dev/null)
+      zis_plugin_id=$(read_component_manifest "${component_dir}" ".zisPlugins[${iterator_index}].id" 2>/dev/null)
+    done
+  fi
+}
+
+# $1 - needle
+# $2 - haystack
+is_substr_of() {
```
jackjia-ibm: Would recommend to move to `bin/libs/string.sh`.

bin/libs/component.sh
```diff
@@ -408,6 +408,154 @@ process_zss_plugin_install() {
...
+  cp "$1" "//'$2'"
+  return $?
+}
+
+# $1 = data_set_name
+# $2 = file_name
+copy_mvs_to_uss() {
+  cp "//'$1'" "$2"
+  return $?
+}
```
jackjia-ibm: Would recommend to move to `bin/libs/fs.sh` or `bin/libs/zos-dataset.sh`.

bin/libs/component.sh
```diff
@@ -408,6 +408,154 @@ process_zss_plugin_install() {
...
+
+zis_plugin_install() {
+  plugin_path="${1}"
+  zwes_zis_pluginlib="${2}"
+  zwes_zis_parmlib="${3}"
+  zwes_zis_parmlib_member="${4}"
+  plugin_id="${5}"
+  component_dir="${6}"
+  parmlib_member_as_unix_file="./${zwes_zis_parmlib_member}.txt"
```
jackjia-ibm: Would recommend to use `create_tmp_file` defined in `bin/libs/fs.sh`. Put into current directory will become unpredictable depends on where the user kicks off the `zwe` command.

bin/libs/component.sh
```diff
@@ -408,6 +408,154 @@ process_zss_plugin_install() {
...
+            else
+              # The key doesn't exist. Just add the key-value pair to the end of the file.
+              add_key_value_at_end_of_file "$samplib_key_value" "$parmlib_member_as_unix_file"
+              changed=1
+            fi
+          fi
+        done < "${component_dir}${plugin_path}/samplib/${params}"
+      done
+    fi
+  echo "Successfully installed ZIS plugin: ${plugin_id}"
```
jackjia-ibm: Use `print_message` function defined in `bin/libs/common.sh` so it will comply verbose level and logging control.

bin/libs/component.sh
```diff
@@ -408,6 +408,154 @@ process_zss_plugin_install() {
...
+      for params in $(ls ${component_dir}${plugin_path}/samplib); do
+        if [ ! -f "${component_dir}${plugin_path}/samplib/${params}" ]; then
+          exit 2
+        fi
+        while read samplib_key_value; do
+          prefix=$(echo "$samplib_key_value" | cut -c -2)
+          if [ "$prefix" = "//" ] || [ "$prefix" = "* " ] || [ "$prefix" = "" ]; then
+            continue
+          fi
+          grep -x "$samplib_key_value" "$parmlib_member_as_unix_file" > /dev/null
```
jackjia-ibm, line 535: Better to capture output so we can control what will be displayed to terminal.

This output can be displayed in trace by using `print_trace` function (or `print_debug` for debug mode).

bin/libs/component.sh
```diff
@@ -408,6 +408,154 @@ process_zss_plugin_install() {
...
+# $1 - needle
+# $2 - haystack
+is_substr_of() {
+  is_substring=$(echo "$2" | grep -c "$1")
+  return "$is_substring"
+}
+
+# $1 - file_name
+# $2 - data_set_name
+copy_uss_to_mvs() {
```
jackjia-ibm: I think there is version for this purpose defined as `copy_to_data_set` in `bin/libs/zos-dataset.sh`.

bin/libs/component.sh
jackjia-ibm: A unique exit code is recommended and should documented in `.errors` file. For example,

```
print_error "Error ZWEL0200E: failed to copy USS file %s to MVS data set %s."
return 200
```

jackjia-ibm (commented):
Thanks Lenny! Would recommend to add more different level of debugging information with `print_message`, `print_debug`, `print_trace` or `print_error` to help on troubleshooting.

bin/libs/component.sh
DivergentEuropeans: Makes sense, ok

schemas/zowe-yaml-schema.json
```diff
@@ -32,6 +32,17 @@
                   "type": "string",
                   "description": "Zowe PARMLIB"
                 },
+                "parmlibMembers": {
+                  "type": "object",
+                  "additionalProperties": false,
+                  "description": "Holds Zowe PARMLIB members",
```
pavelzlatnik: I would change description to "Holds Zowe PARMLIB members for plugins", to distinguish from "parmlib": {
                  "type": "string",
                  "description": "Zowe PARMLIB"

schemas/zowe-yaml-schema.json
```diff
@@ -32,6 +32,17 @@
...
+                  "properties": {
+                    "zis": {
+                      "type": "string",
+                      "description": "PARMLIB member for ZIS plugins"
```
1000TurquoisePogs: Needs better limits than just "string".

https://github.com/zowe/zowe-install-packaging/blob/v2.x/master/schemas/server-common.json#L20
Create an object here-ish, named $anchor=zoweDatasetMember
minlength=1, maxlength=8, pattern=alphanumerics
It's technically possible to have unprintable characters for member names, but it's quite crazy so let's just stick to alphanumerics.
Then back over here,
```
"properties": {
  "zis": {
    "$ref": "/schemas/v2/server-common#zoweDatasetMember",
    "description": "PARMLIB member used by ZIS"
  }
}
```

1000TurquoisePogs (dismissed):

schemas/zowe-yaml-schema.json
DivergentEuropeans: Sure, makes sense!

schemas/server-common.json
```diff
@@ -24,6 +24,14 @@
       "minLength": 3,
       "maxLength": 44
     },
+    "datasetMember": {
+      "$anchor": "zoweDatasetMember",
+      "type": "string",
+      "description": "A 44-char all caps dotted ZOS name",
```
pavelzlatnik: Hi, I think this description is for dataset name, not for member name, it should be "A 1-8char all caps dataset member name"

DivergentEuropeans: Yes! Curses

nosrednayduj:
It looks like you are not implementing a feature from the V1 ZIS plug-in install scripts, which is the ability to provide an environment variable as a value, to be substituted. E.g.,
IZP.ZSSP.AUXNAME=$ZWE_zowe_setup_security_stcs_aux
Implemented in V1 components/zss/bin/zis-parmlib-update.sh function resolve_env_parameter
We use this feature heavily, please do not forget it.  Thank you.

@DivergentEuropeans : I hope I addressed this with https://github.com/zowe/zowe-install-packaging/pull/2947/commits/28018bc27ee4357ebaf1a043e21c02aee2080820

jackjia-ibm (approved):
Thanks a lot Lenny for working and testing this.

bin/libs/component.sh
DivergentEuropeans: Unfortunately it doesn't work the same way with print_trace 
I tried
print_trace "$(grep -x "$samplib_key_value" "$parmlib_member_as_unix_file" > /dev/null)"
and
print_trace "$(grep -x "$samplib_key_value" "$parmlib_member_as_unix_file" )"

1000TurquoisePogs:
My understanding is this is an improvement for those who need zis plugins and will do no harm, so with everyone's testing i'm merging this.

1000TurquoisePogs:
Thank you everyone, tested as working!

PR #2881, opened 2022-05-19
Update schema files 

1000TurquoisePogs:

#### PR type
- [x] Bugfix 
- [x] Feature 

When testing with configmgr, I realized the schema files needed a few tweaks.

I also noticed that manifest had no schema, but needed one. So I read through the docs.zowe.org spec and then tested the code. I found minor differences between the doc and the code, and I also found some new features not yet in the documentation. So, I think this new schema file could be used to revise the documentation on manifest files.

I added a little file for a trivial schema. Perhaps component developers could copy it for their own projects, or use it in the field temporarily if something was wrong.

- [x] No

It's a requirement for configmgr PR (to be made and linked here soon)

Not yet. temporarily, we could update the doc manually but at some point perhaps the schema could be used to automate the doc site content.


schemas/server-common.json
```diff
@@ -8,7 +8,7 @@
       "$anchor": "zoweSemverVersion",
       "type": "string",
       "description": "A semantic version, see https://semver.org/",
-      "pattern": "^(0|[1-9]\\d*)\\.(0|[1-9]\\d*)\\.(0|[1-9]\\d*)(?:-((?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\\.(?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\\+([0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?$"
+      "pattern": "^(0|[1-9]\\d*)\\.(0|[1-9]\\d*)\\.(0|[1-9]\\d*)(-*[a-zA-Z][0-9a-zA-Z\\-\\.]*)?(\\+[0-9a-zA-Z\\-\\.]*)?$"
```
1000TurquoisePogs: These mean equivalent things but have been simplified for performance and compatibility with a wider variety of regexp interpreters.

Issue #2879, opened 2022-05-19
Demo on zowe-install-packaging integrated with configmgr


1000TurquoisePogs:
As mentioned in recent weeks, today I'm able to demo the work several of us have been contributing to,
In this demo, I'll show how the new code is a solution to https://github.com/zowe/zowe-install-packaging/issues/2825 and can eventually solve https://github.com/zowe/zowe-install-packaging/issues/2073 / https://github.com/zowe/community/issues/1389 (in zwe, not components)
This is largely to address the items in https://github.com/zowe/zowe-install-packaging/issues/2662 though parmlib support (https://github.com/zowe/zlux/issues/777) is not yet here although we believe that's less work to finish than what we've done so far.

1000TurquoisePogs:
Recording here: https://zoom.us/rec/share/AiSwEA5_7pNXSitLsQuAnklz3Pfx3_h-VMstCi7xrc1s_Memnaspr3GvQq-p67ZP.6QIAGOHvR1exnXyd

Issue #2662, opened 2022-02-15
v2 server+extension configuration 


1000TurquoisePogs:
This is a ticket to track all threads regarding making v2 configuration:
- Verifiable as valid through automated tools
- Trackable when additions/removals/changes occur between versions
- Familiar for unix and mvs experts alike
- Concise
- Extendable for products to build onto
- Dependency-less. We have a nodejs dependency today that we want to drop so that server setup needs no extra installed software
- Editable through intelligent editors. Not just syntax highlighting, but to have hoverhelp, dropdown options, warnings, all the features you'd expect from an IDE
- Easy to clone. It is difficult to clone a v1 instance to another system due to hardcoded paths for example, so keeping non-portable attributes separate from portable ones, and making as many attributes portable is idea

...maybe more? Please comment on additional desires we have.

This has so far resulted in the following design that is under implementation:
- Zowe should take a list of configuration "files" as an argument to the STC startup
- The files are not exclusively yaml. To make MVS experts happy too, we want to support at minimum both yaml and parmlib.
- The list is comma separated
- There is a documented way in which the different file types can contain the same type of information, such as port numbers
- The list gets merged into a single output configuration. Files on the right hand side have their properties overridden by the files to the left
- The output will include attributes that are references in a Schema
- json-schema will be the technology, as it is standard and is already in use elsewhere in zowe (CLI)
- Schema can have sections that are open-ended, or to be filled in by extenders, so although the attributes are  backed by a schema, the schema itself can be as restrictive or as open as desired.
- The schema will serve as one type of documentation. Our docs site can derive valid attributes and their descriptions from it. Our docs site can generate change notes by sensing the differences between versions.
- We should stick to json schema features that are widely supported by popular editor plugins/schema inspectors, as this will allow us to use or create powerful wizards and IDE tools for quickly and simply customizing zowe servers
- Templating within config files would be very nice, such as a component referencing a java lib via {{java.java_home}}/lib/some-jar.... but we should stick to a template style that is compatible with json schema tools, or else make statements such as that schema validation happens after template resolution, or that templates are considered strings if they look like strings, so that editors can continue to function.

balhar-jakub:
Thanks for capturing the requirements. Just one comment on one of the requirements. 

[...]

PR #2885, opened 2022-05-20
zwe ts edition, with configmgr

1000TurquoisePogs:
This PR introduces the configmgr into the zowe build, and uses it in an all-new port of the existing zwe scripts over to typescript which the configmgr can execute.
The ts code (running as js after build) acts nearly identical to the existing shell code, and it should be possible to get 100% compatibility in the near future. 

**The typescript code is not running under node. configmgr has its own js interpreter so there is no prereq or dependency**

configmgr has advantages by itself, such as validating configuration against schemas, templating within the config file, and allowing for multiple configuration files in a tiered list that gets merged into 1 unified config for server use, so perhaps we could put our defaults in one file and overrides in another.

The ts code builds upon these advantages to also enable compile-time checks so we can spot & fix bugs more easily, but also it's much, much faster than the shell code at doing equivalent actions.

So far, the ts code has only been written to implement "zwe internal start prepare". No other phases are implemented yet but can be over time.

**This code has no impact unless you opt-in to it.** If you are just using `zwe` the same way you always did, you will get the same behavior as before.
If you instead set the env var `ZWE_CLI_CONFIGMGR=true`, then `zwe` commands will use the ts+configmgr code paths, at least for the `zwe` commands we've implemented so far. The rest will just run the usual shell scripts.

**This has no change to components with/without `ZWE_CLI_CONFIGMGR=true` in that they get the same execution environment, and barring bugs, same workspace contents and same env vars.**

Each ts file exists in the same directory as the shell counterparts.
In the `bin/commands` area, each `zwe` command starts as a shell script and can switch over to TS depending on the env var. When it does this, at the moment a `cli.ts` is used to bridge the gap between shell code and ts code.

**This PR depends on various small PRs from other repos, see https://github.com/zowe/zowe-install-packaging/issues/2662 for the list**

**A recording showing how this code works and what benefits it as is here https://github.com/zowe/zowe-install-packaging/issues/2879**

## Testing
Try running `zwe start`, you should see no difference in behavior.
Now get a clean workspace directory and run  `zwe internal start prepare`
Then compare it to the workspace made by `ZWE_CLI_CONFIGMGR=true zwe internal start prepare`
Additionally, try the new features:

- Make a typo in zowe.yaml, like a bad port or a typo in the keyring name type PKCS12, or missing a required property.  The "prepare" should fail with a warning saying verification failed and pointing to what in particular the issue was.
- Try templating in zowe.yaml (format `${{ js-goes-here }}`). Try for example, mapping the externalPort to the gateway port, such as by setting the externalPort to ` ${{ components.gateway.port }}`
- Try multiple config files. instead of `zwe internal start prepare -c zowe.yaml` do `ZWE_CLI_CONFIGMGR=true zwe internal start prepare -c FILE(override.yaml):FILE(base.yaml)`

If you'd like to inspect the result of templating and merging, look for the file `workspaceDirectory/.env/zowe-merged.yaml`.
Also try looking at the envs in `workspaceDirectory/.env` in general. **If you spot a difference, let us know!**

**note: you can't go between ZWE_CLI_CONFIGMGR=true and ZWE_CLI_CONFIGMGR=false once you start using templating and multi-file, because the shell scripts dont support it. You can swap between these modes if you use just an ordinary zowe.yaml though**

#### PR type
- [x] Feature 

#### Relevant issues
See https://github.com/zowe/zowe-install-packaging/issues/2662
And https://github.com/zowe/zowe-install-packaging/issues/2879

- Bundles the 'configmgr' binary for zos. Configmgr should compile on other platforms but we'll get there soon.
- Updates schema files upon testing them
- Uses configmgr to read the schema files to validate zowe.yaml and manifest.yaml files
- Uses configmgr's builtin JS engine to run this logic and overall all logic of "zwe internal start prepare" within the context of TS scripts. This PR contains a TS equivalent of most shell scripts within /bin/libs and the behavior is nearly identical (just a few TODOs in code) and can be 100% identical in the near future.
- This PR does not apply the TS code to every single zwe command, but it can be done over time.
- The TS code and configmgr do not execute unless the env var "ZWE_CLI_CONFIGMGR=true". Although the configmgr should be backward compatible, in case of bugs we can just disable it. For now, we can consider it opt-in feature until we get more coverage & confidence in it.

- [x] No

I will be updating the README of the bin folder to explain how this code works, soon.


bin/commands/internal/start/index.ts
```diff
@@ -0,0 +1,42 @@
+/*
```
1000TurquoisePogs, line 1: This file isnt in use yet, I could either remove it or finish it. Currently only `zwe internal start prepare` uses the TS code.
This file is "excluded" in the TS build script as a result.

1000TurquoisePogs:
I just realized the pipeline isnt building the TS code yet, no build instruction given. I'll be updating the PR to run the "npm run prod" soon.

Joe-Winchester (approved):
I've done some testing of this and it looks good to me.  I can't see any regression.

struga0258 (approved):

timgerstel (approved):
Code was presented and reviewed via webex

PR #2941, opened 2022-06-28
[WIP] configmgr ts part2

1000TurquoisePogs:
Updates to enable:
zwe components install
zwe components enable
zwe components disable
zwe internal start
zwe internal get-launch-components
zwe internal config get
opt-in to configmgr via yaml instead of env var.

"set" operations require writing to the left-most config file, whether its yaml or parmlib, so that's still to do.
jq parsing is temporarily lazy, since it's about to be exposed from configmgr itself so I'll switch to that.


1000TurquoisePogs:
wrong origin.

PR #2945, opened 2022-07-07
configmgr ts part3 - start

1000TurquoisePogs:
This PR enables `zwe internal start` when using zwe ts, on z/os. A part 4 PR may still be needed for containers, but this should now allow wide use of zwe ts on z/os.
This PR fixes or adds missing env vars as compared to the part 1 and 2 zwe ts PRs. The list of env vars each component receives is now nearly identical, with only some zwe internal env vars not seen but they were not for use by components to begin with.

How to test? Just run `zwe internal start` with or without configmgr enabled (--configmgr or zowe.useConfigmgr=true) and see if it behaves the same as it does when configmgr is off. In my testing, all servers came up, apiml gateway showed green checkmarks, and desktop apps worked.


bin/commands/internal/start/component/index.ts
```diff
@@ -0,0 +1,110 @@
...
+
+export function execute(componentId: string, runInBackground: boolean=false) {
+  std.setenv('ZWE_CLI_PARAMETER_COMPONENT', componentId);
+  
+  const coreComponents = std.getenv('ZWE_PRIVATE_CORE_COMPONENTS_REQUIRE_JAVA');
+  if (coreComponents && coreComponents.includes(componentId)) {
+    java.requireJava();
+  }
+
+  //node.requireNode(); //needed for what?
```
1000TurquoisePogs: remove

1000TurquoisePogs: or is it needed in the library code?

bin/commands/internal/start/component/index.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  // check and sanitize ZWE_CLI_PARAMETER_HA_INSTANCE
+  config.sanitizeHaInstanceId();
+
+  const WORKSPACE_DIRECTORY = ZOWE_CONFIG.zowe.workspaceDirectory;
+  if (!WORKSPACE_DIRECTORY) {
+    common.printErrorAndExit("Error ZWEL0157E: Zowe workspace directory (zowe.workspaceDirectory) is not defined in Zowe YAML configuration file.", undefined, 157);
+  }
+
+  // TODO really this should use configmgr
```
1000TurquoisePogs: remove todo, code is clean within that function

bin/commands/internal/start/component/index.ts
```diff
@@ -0,0 +1,110 @@
...
+    const manifest = component.getManifest(COMPONENT_DIR);
+    
+    const privateWorkspaceDir=std.getenv('ZWE_PRIVATE_WORKSPACE_ENV_DIR');
+    //TODO CLI_PARAMETER vars come from cli --parameters, and so should be mapped to execute() parameters. This probably doesnt work.
+    const haInstance = std.getenv('ZWE_CLI_PARAMETER_HA_INSTANCE');
+    
+    // source environment snapshot created by configure step
+    const componentName = pathoid.basename(COMPONENT_DIR);
+
+    if (fs.fileExists(`${privateWorkspaceDir}/${componentName}/.${haInstance}.env`)) {
```
1000TurquoisePogs, line 68: should put this within the start script check because, if there is no start script, is there really a reason to load the env vars?

1000TurquoisePogs: are the vars loaded for this component visible to another component? with configmgr? without?

bin/libs/component.ts
```diff
@@ -490,13 +490,19 @@ export function processComponentApimlStaticDefinitions(componentDir: string): bo
           const resolvedContents = resolveShellTemplate(contents);
 
           const zweCliParameterHaInstance=std.getenv("ZWE_CLI_PARAMETER_HA_INSTANCE");
-          const outPath=`${STATIC_DEF_DIR}/${componentName}.${sanitizedDefName}.${zweCliParameterHaInstance}.yaml`;
+          const outPath=`${STATIC_DEF_DIR}/${componentName}.${sanitizedDefName}.${zweCliParameterHaInstance}.yml`;
```
1000TurquoisePogs, line 494: this needs a code comment. why did i do this?

bin/libs/component.ts
```diff
@@ -490,13 +490,19 @@ export function processComponentApimlStaticDefinitions(componentDir: string): bo
...
 
           common.printDebug(`- writing ${outPath}`);
 
-
-          xplatform.storeFileUTF8(outPath, xplatform.AUTO_DETECT, resolvedContents);
-          shell.execSync(`chmod`, `770`, outPath);
+          let errorObj;
```
1000TurquoisePogs, line 500: this needs a code comment. why did i do this?

bin/libs/java.ts
```diff
@@ -19,9 +19,9 @@ import * as config from './config';
 const JAVA_MIN_VERSION=8;
 
 export function ensureJavaIsOnPath(): void {
-  let path=std.getenv('PATH');
+  let path=std.getenv('PATH') || '';
```
1000TurquoisePogs: should instead use "'/bin:.:/usr/bin'" not blank string

bin/libs/node.ts
```diff
@@ -31,9 +31,9 @@ std.setenv('__UNTAGGED_READ_MODE','V6');
 
 
 export function ensureNodeIsOnPath(): void {
-  let path=std.getenv('PATH');
+  let path=std.getenv('PATH') || '';
```
1000TurquoisePogs: should instead use "'/bin:.:/usr/bin'" not blank string

bin/utils/ObjUtils.ts
```diff
@@ -49,6 +49,7 @@ export class Flattener {
 
     flatten1(json:any, first:boolean, keyPrefix:string, result:any):void {
         if (Array.isArray(json)){
+            result[keyPrefix] = json.join(',');
```
1000TurquoisePogs: can i make this into a conditional as part of the arg? because not everyone may want this feature...
this is the feature where an entire array will not only be seen as a0, a1, but allso that a=0,1

manifest.json.template
```diff
@@ -12,7 +12,7 @@
   },
   "binaryDependencies": {
     "org.zowe.zlux.zlux-core": {
-      "version": "~2.2.0-V2.X-STAGING-ZLUX-CORE",
+      "version": "~2.2.0-zlux-app-server-PR-230",
```
1000TurquoisePogs: clean up because point to staging once merged

struga0258 (approved):

timgerstel (approved):

bin/commands/internal/start/component/index.ts
1000TurquoisePogs: the difference of with/without configmgr does not matter here, checked the behavior by exporting vars from one component to another and seeing that with launcher, env vars are not shared, but without launcher they are.

PR #2947, opened 2022-07-10
Test zowe-install-packaging PR https://github.com/zowe/zowe-install-packaging/pull/2832

DivergentEuropeans:
PR made to test https://github.com/zowe/zowe-install-packaging/pull/2832 such that pipeline deployment is made - all info available in that link


bin/libs/component.sh
```diff
@@ -416,6 +416,153 @@ process_zss_plugin_install() {
...
+            num=$(get_line_number_of_key "$samplib_key" "$parmlib_member_as_unix_file")
+            if [ "$num" != "" ]; then
+              parsed_zwes_zis_parmlib_keys=$(replace "${zwes_zis_parmlib_keys}" "." "_") # replace . with _ in keyname for working key search
+              parsed_samplib_key=$(replace "${samplib_key}" "." "_") # replace . with _ in keyname for working key search
+              samplib_key_value=$(read_json_string ${parsed_zwes_zis_parmlib_keys} ".${parsed_samplib_key}")
+              if [ "$samplib_key_value" = "list" ]; then
+              # The key is comma separated list
+                parmlib_key_value=$(get_string_at_line_number "$num" "$parmlib_member_as_unix_file")
+                parmlib_value=$(get_value_of_string "$parmlib_key_value")
+                samplib_value=$(get_value_of_string "$samplib_key_value")
```
nosrednayduj: Seems like there should be a separate variable to contain whether or not this should be a list, than its value. Looks to me like you are overloading samplib_key_value and the value will end up being "list" in this case, and that's probably not intended.  Either that or it will get an error because it can't find the value of "list", or something.

bin/libs/fs.sh
```diff
@@ -88,6 +88,17 @@ create_tmp_file() {
...
 }
 
+###############################
+# Copies dataset to Unix file
+#
+#
+# @param string   Dataset name
+# @param string   Unix file name
+copy_mvs_to_uss() {
+  cp "//'$1'" "$2"
```
nosrednayduj, line 98: So this will not work if a PDS is open, even in non-exclusive mode.  cp is just stupid. Zowe V1 has a copy item to PDS that uses a CLIST to do the copying. I copied this and made it more generic for UMS purposes.  I note it does have some issues, in particular it requires the user to have a prefix. But maybe there is some way to rewrite it with some temporary data set that is specified in zowe.yaml.

I can't find it in git hub directly, because I'm sure it's in some older branch. This is where I find it in my V1 installation:
scripts/utils/zowe-install-xmem.sh which calls zowe-copy-to-JES.sh which calls mcopyshr.clist

I can provide the generic script that we created for UMS from this base. It would be great if something like that used everywhere instead of cp. (Especially if the bugs were fixed.)

bin/libs/component.sh
DivergentEuropeans: Correct

bin/libs/component.sh
```diff
@@ -501,66 +513,112 @@ zis_plugin_install() {
...
-                remove_key_value_at_line_number "$num" "$parmlib_member_as_unix_file"
-                add_key_value_at_end_of_file "$samplib_key_value" "$parmlib_member_as_unix_file"
-                changed=1
-              fi
+            update_uss_parmlib_key_value "$samplib_key_value" "$parmlib_member_as_unix_file"
+            if [ $? -ne 0 ]; then
+              print_message "Failed to install ZIS plugin: ${zis_plugin_id}"
+              exit 1
             else
-              # The key doesn't exist. Just add the key-value pair to the end of the file.
```
nosrednayduj: So this ends up being
else
fi
which is kind of vacuous. You should remove the else.

PR #3006, opened 2022-08-25
Component schema validate

1000TurquoisePogs:
#### PR type

- [x] Feature 

- Changes manifest spec to allow component schema to be a list. string still supported.
- Fetch schema set from component during validate stage
- Validate config against component schema for each component to see if any of them have an issue with zowe.yaml
- Exit on validation failure only after checking them all

- [x] No. This enforces the agreement and documentation that components must have schemas which are to be validated.


bin/commands/internal/start/prepare/index.ts
```diff
@@ -405,7 +410,7 @@ export function execute() {
   common.printFormattedInfo("ZWELS", "zwe-internal-start-prepare", `Zowe version: v${zoweVersion}`);
   common.printFormattedInfo("ZWELS", "zwe-internal-start-prepare", `build and hash: ${runtimeManifest.build.branch}#${runtimeManifest.build.number} (${runtimeManifest.build.commitHash})`);
 
-
+os.sleep(30000);
```
timgerstel: magic number, also not clear why this sleep is necessary

1000TurquoisePogs: it was for testing, now removed.

CHANGELOG.md
```diff
@@ -6,6 +6,13 @@ All notable changes to the Zowe Installer will be documented in this file.
 
 - Updated ZWEWRF03 workflow to be up to date with the installed software
 
+## `2.4.0`
+
+### New features and enhancements
+
+- zwe can now validate component configuration through use of configmgr and json-schema. If a component does not have a schema, a warning will be printed. Due to schemas being required since 2.0, this behavior may change in a later version.
+- Components can now provide an array of schema files rather than just one. This allows for better re-use and organization.
```
DivergentEuropeans, line 14: > Components can now provide an array of schema files rather than just one. This allows for better re-use and organization.

Does this need documentation?

bin/libs/component.ts
```diff
@@ -209,6 +209,71 @@ export function getManifest(componentDirectory: string): any {
...
+    if ((status = CONFIG_MGR.loadSchemas(configRevisionName, schemas))) {
+      common.printError(`Error: Could not load schemas ${schemas} for configs ${configPath}, status=${status}`);
+    }
+
+    if ((status = CONFIG_MGR.setConfigPath(configRevisionName, configPath))) {
+      common.printError(`Error: Could not set config path for ${configPath}, status=${status}`);
+    }
+
+    if ((status = CONFIG_MGR.loadConfiguration(configRevisionName))) {
+      common.printError(`Error: Could not load config for ${configPath}, status=${status}`);
```
1000TurquoisePogs, line 257: return false

timgerstel (approved):

DivergentEuropeans (approved):

CHANGELOG.md
1000TurquoisePogs: schema=documentation, that's part of the point of it

PR #3017, opened 2022-08-30
Fix user message when pc bit fails

1000TurquoisePogs:
#### PR type
- [x] Bugfix 
- [ ] Feature 

- When `zwe components install` could not find or set the PC bit of a ZSS plugin, it would print out an example command for fixing the issue. Now, it shows the exact command you could execute to fix the PC bit problem.

- [ ] Yes
- [x] No


bin/libs/component.sh
```diff
@@ -350,7 +350,7 @@ test_or_set_pc_bit() {
       echo "PC bit set successfully."
       return 0
     else
-      echo "PC bit not set. This must be set such as by executing 'extattr +p $COMPONENT_HOME/lib/sys.so' as a user with sufficient privilege."
+      echo "PC bit not set. This must be set such as by executing 'extattr +p ${path}' as a user with sufficient privilege."
```
DivergentEuropeans, line 353: if @param string component directory means path is component directory, shouldn't it be like 
'extattr +p ${path}/lib/sys.so'

I think 1 is incorrect, either the comment#377 or the code

DivergentEuropeans (approved):

bin/libs/component.sh
1000TurquoisePogs: just noting that we talked and concluded that yes `path` was fine as-is.

PR #3021, opened 2022-09-01
parmlib read/write for configmgr

1000TurquoisePogs:
#### PR type

- [x] Feature 

config input can now accept PARMLIB() in addition to already existing FILE(), and handles read & write in this case. 
The syntax looks like `FILE(/my/customizations.yaml):PARMLIB(TEAM.CUSTOM.ZOWE(YAML)):PARMLIB(ORG.CUSTOM.ZOWE(YAML)):FILE(/zowe/defaults.yaml)`
The code cannot handle if the member name is not the same for all PARMLIB entries, and will print an error in that case.
The syntax inside a PARMLIB is identical to the unix files, it's just yaml-in-parmlib.

- [x] No


JoeNemo (approved):
Looks good, although I can't vouch for many things that will only be tested thru integration test.

PR #3110, opened 2022-10-14
adding the .ts file for commands/internal/container/prestop

AdarshdeepCheema:
Signed-off-by: Adarshdeep Cheema <adarshdeep.cheema@ibm.com>

#### PR type
- [ ] Bugfix 
- [ ] Feature 

#### Relevant issues

Fixes 

- 

- [ ] Yes
- [ ] No

---

---

---

---
_multi-line description_

Doc issue/PR number:

### DIFF
--- a/bin/commands/internal/container/prestop/index.ts
+++ b/bin/commands/internal/container/prestop/index.ts
@@ -10,11 +10,8 @@
 */
 
 import * as std from 'std';
-import * as os from 'os';
 import * as fs from '../../../../libs/fs';
-import * as zoslib from '../../../../libs/zos';
 import * as common from '../../../../libs/common';
-import * as stringlib from '../../../../libs/string';
 import * as node from '../../../../libs/node';
 import * as shell from '../../../../libs/shell';
 import * as config from '../../../../libs/config';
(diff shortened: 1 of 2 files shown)


bin/commands/internal/container/prestop/index.ts
```diff
@@ -0,0 +1,63 @@
...
+    if (result.out) {
+      common.printMessage("- deleting");
+      shell.execOutSync('sh', '-c', `rm -f *.${zwe_cli_parameter_hs_instance}.* 2>&1`);
+      common.printMessage("- refreshing api catalog");
+      let apicatalog_host = `api-catalog-service.${std.getenv('ZWE_POD_NAMESPACE')? std.getenv('ZWE_POD_NAMESPACE'): 'zowe'}.svc.${std.getenv('ZWE_POD_CLUSTERNAME')? std.getenv('ZWE_POD_CLUSTERNAME'): 'cluster.local'}`
+      let success =component.refreshStaticRegistration(apicatalog_host,
+      std.getenv('ZWE_components_api_catalog_port'),
+      std.getenv('ZWE_zowe_certificate_pem_key'),
+      std.getenv('ZWE_zowe_certificate_pem_certificate'),
+      std.getenv('ZWE_zowe_certificate_pem_certificateAuthorities'));
```
1000TurquoisePogs, line 49: these come from zowe.yaml.
so, instead please use the configmgr to get them, because it has more validation than just requesting a string from an env var.
For example,

```
common.requireZoweYaml();
const ZOWE_CONFIG=config.getZoweConfig();
const catalogPort=ZOWE_CONFIG.zowe.components.api-catalog ? ZOWE_CONFIG.zowe.components.api-catalog.port : undefined;
```

if something is required by schema, you can guarantee it is in the object ZOWE_CONFIG there. but, a lot of these parameters are marked as optional (on purpose or accident) so we need to check them.

bin/commands/internal/container/prestop/index.ts
```diff
@@ -0,0 +1,63 @@
...
+      if (success) {
+        common.printFormattedDebug("ZWELS", "zwe-internal-container-prestop,execute", ` refreshStaticRegistration success`);
+      } else {
+        common.printFormattedError("Error ZWEL0142E", "zwe-internal-container-prestop,execute", `processComponentDiscoverySharedLibs failure`);
+      }
+    } else {
+      common.printMessage("- nothing to delete");    }
```
1000TurquoisePogs, line 56: bracket in a weird place?

bin/commands/internal/container/prestop/index.ts
```diff
@@ -0,0 +1,63 @@
...
+import * as std from 'std';
+import * as fs from '../../../../libs/fs';
+import * as common from '../../../../libs/common';
+import * as node from '../../../../libs/node';
+import * as shell from '../../../../libs/shell';
+import * as config from '../../../../libs/config';
+import * as component from '../../../../libs/component';
+
+export function execute() {
+  node.requireNode();
```
1000TurquoisePogs, line 21: you may not want this right here. in the shell script, it was assumed most everything required node.
we see in the library file that the catalog thing uses curl.js... so why not put the requireNode() over *there*
i believe the requireNode() function caches the answer about if node exists, so if you call it multiple times it should not harm performance.

1000TurquoisePogs (changes requested):
You're on the right track but note a few things:

* You don't have a way for prestop/index.ts to be called. Check other command directories: you need cli.ts + a modification in index.sh that determines whether or not to do shell code or javascript code. 
You need to put this boilerplate at the top of index.sh https://github.com/zowe/zowe-install-packaging/blob/congifmgrWorkContainer/bin/commands/internal/start/component/index.sh#L14-L17 and a `fi` at the bottom.

By the way... If you are editing a **non internal** index.sh, there's a slightly larger boilerplate about managing the permissions of workspace directory by using a tmpdir, but **internal**/container is internal, so you don't need https://github.com/zowe/zowe-install-packaging/blob/congifmgrWorkContainer/bin/commands/start/index.sh#L14-L22

* Your code did not compile. Go to `build/zwe` and run `npm run dev` to see if your code compiles. I think you are missing a bracket or something. The pipeline build failed due to this, you can see the TSC error in there.

Issue #3142, opened 2022-11-09
Java 11 version string does not pass version check


skurnevich:
**Describe the bug**
Tried to run Zowe with Java 11 in ConfigMgr mode, but get failure on check
<img width="1021" alt="Screenshot 2022-11-09 at 16 54 33" src="https://user-images.githubusercontent.com/61471197/200877819-8c79a4e2-c143-4d1b-88b5-21f2436d61da.png">

Version String Naming changed for that particular version to be `java version "11.0.16" 2022-07-19` and does not start with 1. anymore.

Noticed that [java.ts](https://github.com/zowe/zowe-install-packaging/blob/677a607686e6ee7ecb349dc5925a6f58dd9e61da/bin/libs/java.ts#L125) has incorrect logic and basically checks only that major version is equal to '1'.

Older[ java.sh](https://github.com/zowe/zowe-install-packaging/blob/677a607686e6ee7ecb349dc5925a6f58dd9e61da/bin/libs/java.sh#L124) (without ConfigMgr) seems to be processing this version better

JoeNemo:
fixed in code

PR #3143, opened 2022-11-09
Fix for Java major version check

skurnevich:
Signed-off-by: Sergei Kurnevich <sergei.kurnevich@broadcom.com>

#### PR type

- [x] Bugfix - minor
- [ ] Feature 

#### Relevant issues

Fixes https://github.com/zowe/zowe-install-packaging/issues/3142


1000TurquoisePogs (approved):
This looks good to me. Should we have a max version check as well?

skurnevich:
Hi Sean, I'm not sure what max version we could use for that check. 
Our docs state that only Java 8 is supported, but we've been asked to test Zowe with Java 11, so that is why I spotted the issue and made the version check more generic. However, I do not know what versions could be supported in the future, I am not even sure if we going to support Java 11 in the end.
Do you know what major versions we could check for? 

Just btw, I also wanted to add some regex to match major and minor version values, something like 
`(?!\.)(\d+(\.\d+)+)([-._][\w-]+)?(?![\d.])` 
but it seems that the Java version string format could vary and it can omit minor and path numbers if those are 0, i.e. `openjdk version “11” 2018-08-25` is also potentially possible output format, which makes the effort useless.

PR #3193, opened 2022-12-05
React to component configure bad RCs

1000TurquoisePogs:

#### PR type
- [x] Feature 

This PR checks the RC of running component.configure scripts for when rc is not 0. Previously, when not 0, the component would be silently ignored. Now, at minimum there is a warning message. But the schema is also updated to include the option to have zowe exit instead of continuing in case the user wants increased confidence.

- [x] No

### DIFF
--- a/bin/commands/internal/start/prepare/index.ts
+++ b/bin/commands/internal/start/prepare/index.ts
@@ -362,6 +362,11 @@ function configureComponents(componentEnvironments?: any, enabledComponents?:str
           } else {
             //not printformattederror because it makes silly long strings like "ERROR: 2022-12-07 20:00:47 <ZWELS:50791391> me ERROR (zwe-internal-start-prepare,configure_components) Error ZWEL0317E: Component app-server commands.configure ended with rc=22."
             common.printError(`Error ZWEL0317E: Component ${componentId} commands.configure ended with rc=${result.rc}.`);
+            if (result.out) {
+              const outLines = result.out.split('\n');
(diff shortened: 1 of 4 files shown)


bin/commands/internal/start/prepare/index.ts
```diff
@@ -355,13 +355,16 @@ function configureComponents(componentEnvironments?: any, enabledComponents?:str
                 shell.execSync('chmod', `700`, `"${zwePrivateWorkspaceEnvDir}/${componentName}/.${zweCliParameterHaInstance}.env"`);
               }
             }
-          }
-
-          if (result.rc == 0) {
             const outLines = result.out.split('\n');
             common.printFormattedInfo("ZWELS", "zwe-internal-start-prepare,configure_components", `- commands.configure output from ${componentId} is:`);
             common.printMessage(outLines.filter(line => !line.startsWith('export ')).join('\n'));
```
1000TurquoisePogs, line 360: we should also print out the output, if it exists.

schemas/zowe-yaml-schema.json
```diff
@@ -512,6 +512,11 @@
               "type": "string",
               "description": "Log level for Zowe launch scripts.",
               "enum": ["", "info", "debug", "trace"]
+            },
+            "onComponentConfigureFail": {
+              "type": "string",
+              "description": "Chooses how 'zwe start' behaves if a component configure script fails",
+              "enum": ["warn", "exit"]
```
1000TurquoisePogs: should say default=warn here because thats what the code does

timgerstel (approved):

PR #3196, opened 2022-12-09
Adds permission for RAUDITX to ZWESECUR

JirkaAichler:
Allow Zowe user to use RAUDITX resource that is necessary for API ML to cut SMF records. This will be a new functionality planned in the upcoming release that will improve user authentication auditing.

https://github.com/zowe/api-layer/pull/2691


JirkaAichler:
Opening the new one that is not based on the fork.

PR #3198, opened 2022-12-12
Adds permission for RAUDITX to ZWESECUR

JirkaAichler:
Allow Zowe users to use the RAUDITX resource that is necessary for API ML to cut SMF records. This will be a new functionality planned in the upcoming release that will improve user authentication auditing.

https://github.com/zowe/api-layer/pull/2691


JoeNemo:
Is the need for this privilege an opt-in or a must-have?   Not every user will have this.

JirkaAichler:
> Is the need for this privilege an opt-in or a must-have? Not every user will have this.

It is optional. API ML will just not generate SMF ESM audit records when not configured. Should I update the comments in JCL?

1000TurquoisePogs (approved):
This is consistent with the discussion we had on use of this, so great.

PR #3238, opened 2023-01-23
Suggest zwe components upgrade upon installing a component twice

1000TurquoisePogs:
It's easy for people to encounter error 155 when installing a component twice, because people can be lead to think the way to upgrade a component is just to install a new version using `zwe components install`. the actual command is `zwe components upgrade`. but, error 155 will be printed on doing an install for an already installed component, and so I improved the message to suggest the user to use `upgrade` instead, if they meant to upgrade.

#### PR type
- [x] Feature 

- [ ] Yes
- [x] No

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ All notable changes to the Zowe Installer will be documented in this file.
 
 - Updated ZWEWRF03 workflow to be up to date with the installed software
 
+## `2.7.0`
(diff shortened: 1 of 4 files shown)


DivergentEuropeans (approved):
Good change. I'd suggest adding ${target_dir} as well to make the command literal, as running 'zwe components upgrade' literally would fail

PR #3297, opened 2023-03-01
Prevent rc3 on app-server when symlink moves

1000TurquoisePogs:
app-server can throw rc=3 if startup is missing some auth configuration.
This can happen due to a bug we've noticed with symlinks, where the rarely-used trivial-auth plugin nevertheless is loaded by the server yet not at a relative path, such that if zowe's runtimedirectory is a symlink which has moved, the path becomes incorrect, the server doesnt find the plugin, and then the server refuses to load.
By adding this plugin pointer and putting it at a relative path like every other core plugin, this should resolve the portability issue.

#### PR type
- [x] Bugfix 

- [ ] Yes
- [x] No

### Testing

Run zowe with runtimedir set to a symlink
Then, change the symlink location and rerun zowe.
Or, just go into workspace/app-server/plugins and change this file to some bogus path and watch rc=3 happen.

This pr should fix it.


timgerstel (approved):

DivergentEuropeans (approved):
I found an issue with the build at the install step by @struga0258 was kind enough to test it locally and confirm it builds correctly (unrelated issue from that install step)

PR #3309, opened 2023-03-06
fix launcher schema failure

1000TurquoisePogs:
For some reason, the regex in the schema used for semver checking is not working correctly.
The regex should work fine, according to regex websites, but in reality the launcher's latest version "2.0.10" fails to pass validation.
I restructured the regex to be functionally similar, yet simplified, and this allows the launcher to pass the check.
This pattern also solves https://github.com/zowe/zowe-common-c/issues/352

### DIFF
--- a/schemas/server-common.json
+++ b/schemas/server-common.json
@@ -8,13 +8,13 @@
       "$anchor": "zoweSemverVersion",
       "type": "string",
       "description": "A semantic version, see https://semver.org/",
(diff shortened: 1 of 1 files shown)


struga0258 (approved):

1000TurquoisePogs:
Merging this because I know it helps with the RC, even though the tests are failing due to a different cause.

PR #3322, opened 2023-03-14
Fix list of discovery services

pavel-jares-bcm:
This PR fixes the collection of URLs of discovery services. It should be based on https://github.com/zowe/zowe-install-packaging-tools/blob/692c38383060a1616787cfe4463284552df437bb/config-converter/src/libs/yaml.js#L23. Anyway, both of codes are not accurate.

The code should construct environment property `ZWE_DISCOVERY_SERVICES_LIST`. This value should contain all enabled DS in API ML. The original (sh) code in case of haInstance use the hostname from there and just in case the hostname is missing use the external domain. In the case of non-HA setup, it used an external domain. It is closer to real requirements because the TS version collects all instances and URLs based on the external one (the external one does not make any sense).

There are two possibilities:
- Zowe (API ML) is in HA
  - it is required to use specific hostnames (not external ones, APIML cannot use DVIPA etc.)
- Zowe (API ML) is not in HA
  - zowe.yaml should be without haInstance section and the external domain represent the one instance
  - haInstance are present and contain a concrete value.

It is important to not forget that discovery service could be defined in root components and in components of a specific instance.

This implementation is more strict. For cases in the middle, it rather writes an error message than tries to "imaging" a configuration.

#### PR type
- [x] Bugfix 
- [ ] Feature 

#### Relevant issues

Fixes 

- 

- [x] Yes
- [ ] No

It is not necessary, but I have changed the JSON scheme to validate `hostname` and `sysname` as mandatory. Defining an instance without those values has no sense. Anyway to fix the bug it is not necessary and it could be removed.

---

---

---

---
_multi-line description_

Doc issue/PR number:


JoeNemo (approved):
Approved.

JoeNemo:
@ojcelis , please look at this one, too.

MarkAckert:
Hi @pj892031 , could you fix the missing DCOs? I'll merge after that. Thx

pavel-jares-bcm:
> Hi @pj892031 , could you fix the missing DCOs? I'll merge after that. Thx

Hi @MarkAckert , the DCO was set and all actions are passing. Could you please merge it? Thank you.

PR #3361, opened 2023-04-05
Add stdout during component install

1000TurquoisePogs:

#### PR type
- [x] Feature 

Users report the desire to have components print output during installation.
We started doing this recently during configuration, but not installation.
Intentionally or not, configmgr=false does print stdout during installation, but not configmgr=true.
So, this PR enhances the TS code to add lines similar to those of configure, but for install.

- [x] No


JoeNemo (approved):
Approved.

PR #3418, opened 2023-05-11
Remove trivial-auth from default list of plugins

### LINKED ISSUE #966: Reason to use trivialAuth provider
The ZLUX application contains a couple of authentication providers. One of them is trivialAuth. This provider accepts userId from the JWT token without any validation.

I understand it could be useful for testing purposes, but should it be really used in production? Even as a fallback?

ZLUX should be accessible via Gateway (btw. GW is checking the validity of the token) and the reason to use fallback is that GW is not available.

I guess trivialAuth should be disabled during release.

```ts
module.exports.getUserId = (apimlTkn) => {
  let base64UrlToBase64 = (input) => {
    let result = input.replace(/-/g, '+').replace(/_/g, '/');
    const padCount = result.length % 4;
    if (padCount > 0) {
      if (padCount === 1) {
        throw new Error('bad length of base64url string');
      }
      result += new Array(5 - padCount).join('=');
    }
    return result;
  }

  let userid;
  try {
    const payloadBase64Url = apimlTkn.split('.')[1];
    const payloadBase64 = base64UrlToBase64(payloadBase64Url);
    const payloadString = Buffer.from(payloadBase64, 'base64').toString();
    const payloadObject = JSON.parse(payloadString);
    userid = payloadObject.sub;
  } catch (e) {
    throw new Error(`failed to parse APIML token: ${e}`);
  }
  return userid;
}

  authenticate(request, sessionState) {
    if (request.body && request.body.username) {
      sessionState.username = request.body.username;
      sessionState.authenticated = true;
      return Promise.resolve({ success: true });
    } else if (request.cookies && request.cookies[TOKEN_NAME]) {
      try{
        sessionState.username = apiml.getUserId(request.cookies[TOKEN_NAME]);
      } catch (e) {
        return Promise.resolve({ success: false });
      }
      sessionState.authenticated = true;
      return Promise.resolve({ success: true })
    } else {
      return Promise.resolve({ success: false });
    }
  },
```

1000TurquoisePogs:
"trivial-auth" can be used deliberately, but I haven't seen a user who does this and it's not well documented. Yet, its in the default config. Here, it doesn't lower security because "saf-auth" is also present and that does real saf checks, so this "trivial-auth" is redundant and can only get in the way.
I'm removing it from the list of running plugins by default. People can re-add it if they really need it, but I expect nobody will be disrupted nor need to re-add it.

This is to resolve https://github.com/zowe/zlux/issues/966 and https://github.com/zowe/zlux/issues/970


struga0258 (approved):

1000TurquoisePogs:
I'm bypassing the test build here to save time, we're doing a test after all the dependent PRs are in one place.

PR #3448, opened 2023-06-12
Update ZWENOKYR

Martin-Zeithaml:
1 line change in comment section of JCL. Based on description, it might be confusing, that ZWEKRING is used for defining and ZWENOKYR too.


JoeNemo (approved):
Approved

PR #3506, opened 2023-07-13
build: make APIML_OIDC vars default for all builds

MarkAckert:
Some of our test builds are failing due to missing oidc vars. We toggle OIDC enablement on/off based on `oidc_enable` var, but even when false the Ansible playbooks still expect all oidc vars to be set. This change makes it so APIML_OIDC vars are always set across all installations, and we continue to rely on the `oidc_enable` var to modify the behavior of APIML within the tests. When `oidc_enable` is false, the set variables will be ignored, when true, they will be picked up. 

In the case where the ENV for OIDC is unset (a likely error condition), a dummy value is used which indicates where it was set, in case further debugging is required.


ojcelis (approved):

ojcelis:
Test will defer in RC2

PR #3507, opened 2023-07-13
Applied keystore flags to keep keystore compat with all components

### LINKED ISSUE #3157: Need to automate java compatibility flag for PKCS12 keystore generation
https://github.com/zowe/docs-site/issues/2459 describes a problem when generating keystores of PKCS12 type specifically, when using newer versions of java.
The problem manifests when using ZSS. If the keystore is not backward compatible, ZSS prints:
`ZWES1060W Failed to init TLS environment, rc=1(Handle is not valid)`

[java.sh](https://github.com/zowe/zowe-install-packaging/blob/v2.x/master/bin/libs/java.sh) (*non-configmgr mode*)
and [java.ts](https://github.com/zowe/zowe-install-packaging/blob/v2.x/master/bin/libs/java.ts) (*configmgr mode*)
are where java version checks occur.

[certificate.sh](https://github.com/zowe/zowe-install-packaging/blob/v2.x/master/bin/libs/certificate.sh) is where keytool stuff occurs. There is no configmgr version of this file yet, it's planned for the future.

Between the java files and the certificate.sh file, it seems like we should be assigning the compatibility flags depending on the version of java that is used. I'm also told that java sr7fp25 will use add the flag `-Dkeystore.pkcs12.legacy` to do similar things, so we may need to set flags depending on whether the java version is `<sr7fp15`, `>=sr7fp15 && < sr7fp25`, and `>=sr7fp25`

The acceptance criteria is...
Be able to generate a PKCS12 keystore on every type of java version listed above, without encountering `ZWES1060W Failed to init TLS environment, rc=1(Handle is not valid)`

1000TurquoisePogs:
Users have reported that when using java over certain versions, the pkcs12 keystores zowe generates will not be readable by zss: https://github.com/zowe/zss/issues/607

Using the patch in https://github.com/zowe/zss/issues/607 and recommendations in https://github.com/zowe/zowe-install-packaging/issues/3157, this PR applies the java keystore generation compatibility flags appropriate to the version of java seen.
I couldn't find a java 8 sr7fp25 despite what the above issue said. I did find a sr8fp0 instead.
I noticed with the versions, which fix version they declared in their version output, such as 1.8.0_341
The _341 at the end is the part we care about for version detection.

In this PR, depending on which fix version of 8 we see, we give certain keystore generation flags to keytool.
I assume java 11+ will use the latest flag of 8, but I don't know, but wrote the code with that assumption.

#### PR type
- [x] Bugfix 

Fixes https://github.com/zowe/zowe-install-packaging/issues/3157

- [x] No

### Testing:

Simply, if you create a pkcs12 keystore with a new java today, zss will not start.
With this bugfix, it should!


bin/libs/certificate.sh
```diff
@@ -199,6 +199,8 @@ pkcs12_create_certificate_authority() {
   password="${3}"
   common_name=${4:-${ZWE_PRIVATE_DEFAULT_CERTIFICATE_CA_COMMON_NAME}}
 
+  flags=$(get_java_pkcs12_keystore_flag)
+
   print_message ">>>> Generate PKCS12 format local CA with alias ${alias}:"
   mkdir -p "${keystore_dir}/${alias}"
   result=$(pkeytool -genkeypair -v \
```
achmelo, line 206: Do you think that this flag should be also specified in functions that cover exports from keyring? like keyring_export_to_pkcs12

bin/libs/java.sh
```diff
@@ -132,3 +132,32 @@ validate_java_home() {
...
+    java_version_short=$(echo "${java_version}" | grep ^"openjdk version" | sed -e "s/openjdk version //g"| sed -e "s/\"//g")
+  fi
+  java_major_version=$(echo "${java_version_short}" | cut -d '.' -f 1)
+  java_minor_version=$(echo "${java_version_short}" | cut -d '.' -f 2)
+  java_fix_version=$(echo "${java_version_short}" | cut -d '_' -f 2)
+
+  if [ ${java_major_version} -eq 1 -a ${java_minor_version} -eq 8 ]; then
+    if [ ${java_fix_version} -lt 341 ]; then
+      printf " "
+    elif [ ${java_fix_version} -lt 361 ]; then
```
achmelo, line 152: I was generating keystore with version 351 using "legacy" flag and it ZSS was ok . Did you find that it is not supported in this update?

bin/libs/certificate.sh
1000TurquoisePogs: i was not sure if it was allowed, but probably it is. i can try it.

bin/libs/java.sh
1000TurquoisePogs: i didn't have a version to test with, i was just following what IBM said to do, even though it was a little confusing.

achmelo (approved):

PR #3523, opened 2023-07-24
prototype of template-based yamls as an alternative to basic example

### LINKED ISSUE #1884: Users do not automatically get new or changed config values upon upgrade
Because zowe's server config is based upon starting with an example config file, and editing it, it is likely that users will create a working config file, and then not update it with every upgrade, even though they should because new zowe versions can come with new features and changes in desired defaults.

I say "desired defaults" because, since there isnt a default file, just an example file, there really isn't a default config.

We should move from an example-based config to a default-based config, using the multi-yaml feature here https://docs.zowe.org/stable/user-guide/configmgr-using/ such that the first file in the list can be zowe's default file, so that users will pick up new values & changes in defaults over time.

1000TurquoisePogs:

#### PR type
- [x] Feature 

#### Relevant issues
Resolves https://github.com/zowe/community/issues/1884
Evolution of https://github.com/zowe/zowe-install-packaging/pull/3356

This PR introduces a new way to setup zowe.
Instead of just getting example-zowe.yaml and editing it, there are 2 files:
* A defaults.yaml file, never to be edited, but to be used in a yaml list
* A templated example yaml file. To be edited, but with templates to reduce configuration steps.

The defaults.yaml, though more work due to needing to use the advanced `CONFIG=FILE(/path/to/my-config.yaml):FILE(/path/to/default.yaml)` syntax, gives advantage of users not needing to take action upon upgrades to zowe: default.yaml will change at our discretion to give the user a smooth upgrade experience.

The idea is, the user finds whichever template yaml that best suits their keystore need, and then edits that, which should require less edits overall due to templates.

These yamls now also have new comments which are hopefully more helpful than before.

- [x] No


Joe-Winchester (approved):
This looks great !   I'm not that famliar with the imported keyring scenarios, but the pkcs12 and keyring generated and imported look really nice.

1000TurquoisePogs:
Closed for now due to defaults.yaml used elsewhere. Will revisit defaults.yaml enhancements over time.

PR #3535, opened 2023-07-27
Allow bypassing version check of zOS

1000TurquoisePogs:
Some components of Zowe have a version check on zOS and when a newer zOS is detected than has been verified as working, an error will print and zowe will stop.
Using the environment variable "ZWE_zowe_launcher_unsafeDisableZosVersionCheck" (and captured in the schema zowe.launcher.unsafeDisableZosVersionCheck) then the user can configure Zowe to only warn instead of error.

prs involved:
https://github.com/zowe/launcher/pull/90
https://github.com/zowe/zowe-common-c/pull/390
https://github.com/zowe/zss/pull/632


JoeNemo (approved):
Approved.

PR #3545, opened 2023-08-03
Small typo fix

Martin-Zeithaml:
Couple of trace messages had either wrong suffix or file reference.

Note: Also in the TS files we are referencing functions from shell: e.g. `generate_instance_env_from_yaml_config` instead of `generateInstanceEnvFromYamlConfig`.

### DIFF
--- a/bin/libs/config.ts
+++ b/bin/libs/config.ts
@@ -95,7 +95,7 @@ export function generateInstanceEnvFromYamlConfig(haInstance: string) {
   }
 
   // delete old files to avoid potential issues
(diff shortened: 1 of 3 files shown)


MarkAckert (approved):

1000TurquoisePogs:
Sure, this looks like a fine fix to me.
Why was it like that to begin with: I wanted to do a direct conversion from SH to TS, keeping messages identical.
Over time, i think it doesn't matter much anymore and can even be confusing. So, ok with changing it.

PR #3593, opened 2023-10-06
Ensure workflow files are in ascii in final PAX

### LINKED ISSUE #3591: XML Parser Error
Problem Description:

we failed to load the zOSMF workflow
/usr/lpp/zowe/files/workflows/ZWECONF.xml
with the following error message

The validation of workflow definition file "/tst/prd23c1/usr/lpp/zowe/files/workflows/ZWECONF.xml" failed. See line "1" for errors. More information from the XML parser: "Content is not allowed in prolog."

Using help for message IZUWF0120E The validation of workflow definition file file-name failed. See line line-number for errors. More information from the XML parser: error-data

brings the following information: 

User response
If error-data is the message Content is not allowed in prolog it is likely that a mismatch exists between the encoding specified on the first line of the workload definition file and the actual encoding of the file. This error can occur, for example, if the file was developed in a workstation-based editor that requires the UTF-8 encoding standard, but you transferred the file in ASCII mode to z/OS, which resulted in an automatic conversion to EBCDIC. If this has happened, you can re-transmit the file in binary mode, or edit the file on z/OS and change the encoding to IBM-1047. Then, try adding the workflow again.

In our case this workflow definition file was transferred by SMP/E during SMP/E installation, so there was no transfer done by us.

Analysis of problem:
Per the doc site in [zowe.org](http://zowe.org/), client can use zosmf to configure zowe (i.e. zowe.yaml) using .../runtime/files/workflows/ZWECONF.xml, however, when the file is used in zosmf, client see error as in below screen.  The issue is codepage of ZWECONF.xml is ebcdic but zosmf takes ASCII files.  Client needs to use iconv to convert the file from IBM-1047 to ISO8859-1. This has been raised as issue in multiple client complains. Is there a chance to make zosmf files in ISO8859-1?

Screenshot 2023-10-05 at 10 55 28 AM

MarkAckert:
Addresses #3591 .

#### PR type
- [x] Bugfix 

#### Relevant issues

Fixes  #3591 

Add a post-processing step in `pre-packaging.sh` which converts workflows from EBCDIC to ASCII. This suggestion was chosen because `pre-packaging.sh` operates on the workflow files in EBCDIC via REXX and Shell utilities, and changing those to tolerate ASCII would be more work than converting after they've completed.

Workflows won't have to be converted from EBCDIC to ASCII

### DIFF
--- a/.pax/pre-packaging.sh
+++ b/.pax/pre-packaging.sh
@@ -57,7 +57,7 @@ function _convertEbcdicToAscii
   for ebcdic_file in $files_to_convert; do
     echo "[$SCRIPT_NAME] converting $ebcdic_file to ascii..."
     tmpfile="${ebcdic_file}.tmp"
-    iconv -f IBM-1047 -t ISO-8859-1 "${ebcdic_file}" >"${tmpfile}"
+    iconv -f IBM-1047 -t ISO8859-1 "${ebcdic_file}" >"${tmpfile}"
     if [ "$using_output_dir" == "yes" ]; then
       mv "${tmpfile}" "${output_dir}/${file}"
     else
--- a/.pax/pre-packaging.sh
+++ b/.pax/pre-packaging.sh
@@ -53,7 +53,7 @@ function _convertEbcdicToAscii
     echo "[$SCRIPT_NAME] will convert $input to ascii, results in $output_dir"
   fi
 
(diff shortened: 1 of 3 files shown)


muzikovam (approved):
Workflows are working for me :)

ojcelis (approved):

PR #3617, opened 2023-10-18
Reduce string fields by introducing better typing

1000TurquoisePogs:
This PR makes the zowe schema a little stricter. I haven't seen anyone have trouble with the plain string fields, but ZEN will be more powerful if it has a more insightful schema, so I tried to eliminate string fields when not open ended.


Martin-Zeithaml (approved):

MarkAckert:
Is there any chance this could cause issues for existing Zowe deployments - would a previously accepted `zowe.yaml` be rejected?

Martin-Zeithaml:
Yes, in this unrealistic scenario:
We have allowed setting like this, when for example the admin group name was not checked for chars or length.
```
YAML:
...
    groups:
      admin: ZW3ADMIN_LONGNAME

JCL:
//ZWESECUR JOB 
...
//         SET  ADMINGRP=ZW3ADMIN_LONGNAME
```
Some legacy products (and I know it, I was working on several) does not check the string length and if more is provided, only allowed maximum is used with no warning.
Let's say that some security product works in this style, so it will use just `ZW3ADMIN` and ignores the rest.
With new schema, the zowe validation will fail.

1000TurquoisePogs:
Yes I would like to think the new "stricter" validation isn't actually stricter in that before, you were allowed to choose values that would just fail at runtime, as opposed to failing at validation time (the point of validation... to make errors easier to spot)

But @Martin-Zeithaml 's point is interesting. A program can silently allow bad values. So we could move from allowing people to have harmlessly incorrect configs to having the inability to start zowe.

Do we care? I don't have evidence anyone is doing this with zowe currently, and I feel that us telling people that a group name is 8 characters max when we know it is, is a justifiable stance to have. One could say "its a breaking change that you no longer let me have invalid input"... but is this risk worth avoiding?

MarkAckert:
> Do we care? I don't have evidence anyone is doing this with zowe currently, and I feel that us telling people that a group name is 8 characters max when we know it is, is a justifiable stance to have. One could say "its a breaking change that you no longer let me have invalid input"... but is this risk worth avoiding?

I think the change in this context is fine and we should go ahead with it. Thanks for looking in to this!

1000TurquoisePogs:
Merged without test due to build system issues.

PR #3620, opened 2023-10-20
Chore - remove node14

1000TurquoisePogs:
Node v14 is no longer supported by ibm, and our documentation therefore also says zowe does not support it.
We have no reason to continue testing it in our testbed, so removing references and updating zwe to exclude it.


MarkAckert (approved):

ojcelis (approved):

ojcelis:
Test failed due to zos 2.5 upgrades. Will test in on separate PR.

PR #3650, opened 2023-11-09
Updates on new v3 FMID

ojcelis:

#### PR type
- [ ] Bugfix 
- [ ] Feature 

#### Relevant issues

Fixes 

- 

- [ ] Yes
- [ ] No

---

---

---

---
_multi-line description_

Doc issue/PR number:


smpe/bld/AZWE003.htm
```diff
@@ -1,5 +1,4 @@
 <!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 3//EN">
```
OnnoVdT, line 1: Updating the PD here is not useful, should be more explicit about that in the readme. The PD source is a BookManager script file based on  on a template that is updated once in a while. To pass Shopz validation, you have to use the latest template. I have the PD script source on my system and adjust it each time I'm notified about a template update (together with updating the other 20 or so PDs I maintain). That script is converted to PDF and HTML, twice, once with IBM-specific data and once generic. It is that generic HTML file that is kept in Git so that the doc folks can pick it up.
For now your update will do, but we must replace closer to GA with the final one generated from script source.

smpe/bld/_README.txt
```diff
@@ -34,7 +34,6 @@ Note: SUP should only be updated if most products depending on the
   SUP(AZWE001)
   DELETE(AZWE001)
 3. remove select files from smpe/bld/service
-   - apar-bucket.txt
```
OnnoVdT: the apar-bucket.txt file must also be removed when it is present. So this line should stay in the README.

OnnoVdT (changes requested):

smpe/bld/AZWE003.htm
ojcelis: sounds good. I'll take note of this

smpe/bld/_README.txt
ojcelis: Will update it

OnnoVdT (approved):

PR #3651, opened 2023-11-10
zzow system updates

MarkAckert:
This PR replaces the existing zzow systems (2,3,4) with the new ones (6,7,8). Related PRs were open and merged in zowe-actions.

Test-VT-Terminal was disabled, as the new system default SSH port created an error condition in the test. The error condition is expected and is the correct behavior.


ojcelis (approved):

MarkAckert:
Integration test fails until we merge this in; so we need to ignore the status check.

PR #3659, opened 2023-11-22
Update TLS settings

1000TurquoisePogs:
This PR clarifies that the cipher format is IANA, and removes TLS 1.0 and 1.1 from the enum as we decided there is no reason to allow downgrading unless there's high demand for that.
The minTls was wrongly stated as 1.3, current is 1.2


DivergentEuropeans (approved):

1000TurquoisePogs:
Marist is failing to build zlux at the moment unrelated to the code it seems, so I'm going to merge this for now.

PR #3663, opened 2023-11-28
Add getesm to build

1000TurquoisePogs:
This puts the new utility "getesm" into bin/utils
getesm was added to zowe-common-c here https://github.com/zowe/zowe-common-c/pull/415

it is not currently used by anything and has no side effects.
it could be called by new code or by users to simply print out which ESM they are using to the STDOUT of a terminal.

it could also be used as a smoke test in the event of configmgr or zss failing.
it is a C program just like them, so if they are failing, it can be used to prove if any trivial C program will run on a system.

### DIFF
--- a/.github/workflows/build-packaging.yml
+++ b/.github/workflows/build-packaging.yml
@@ -171,7 +171,7 @@ jobs:
         with:
           manifest-file-path: ${{ github.workspace }}/manifest.json
           default-target-path: .pax/binaryDependencies/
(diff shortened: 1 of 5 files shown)


Martin-Zeithaml (approved):

1000TurquoisePogs:
Test automation seems broken unrelated to this change. Tested this change manually and its OK, so merging.

PR #3672, opened 2023-11-29
Zwe doc should run on push to v2.x/master (release)

MarkAckert:
Zowe `zwe` doc generation currently runs on push to v2.x/rc, which results in unnecessary runs of the tool and an excess of PRs on the docs-site or build failures. v2.x/rc can receive multiple changes in a single release cycle. 

Instead, v2.x/master should be used for generating doc - there will be a single push per release to this branch, albeit it will come closer the live release time.

### DIFF
--- a/.github/workflows/zwe-doc-generation.yml
+++ b/.github/workflows/zwe-doc-generation.yml
@@ -4,7 +4,7 @@ on:
   # Will run this on push when v2 is out
   push:
     branches:
-      - v2.x/rc
+      - v2.x/master
   workflow_dispatch:
 
 env:


anaxceron:
@MarkAckert -- what's the latest that this new method would add the doc to the master branch? I'd like us to NOT get any updates to the master branch on a GA date in case the Doc Squad runs into a problems w/ the branch and needs to troubleshoot.

MarkAckert:
> @MarkAckert -- what's the latest that this new method would add the doc to the master branch? I'd like us to NOT get any updates to the master branch on a GA date in case the Doc Squad runs into a problems w/ the branch and needs to troubleshoot.

It would usually be the day before, or morning of the GA date. For context, this is a PR opened by the automation, which goes into `docs-staging`: https://github.com/zowe/docs-site/pull/3275

anaxceron:
> > @MarkAckert -- what's the latest that this new method would add the doc to the master branch? I'd like us to NOT get any updates to the master branch on a GA date in case the Doc Squad runs into a problems w/ the branch and needs to troubleshoot.
> 
> It would usually be the day before, or morning of the GA date. For context, this is a PR opened by the automation, which goes into `docs-staging`: [zowe/docs-site#3275](https://github.com/zowe/docs-site/pull/3275)

@MarkAckert yep, I've seen the PRs. I'm just not loving the possibility that this could happen on a GA date. When you say "morning," I'm assuming you mean morning around EST (or thereabouts). But we may have someone in China working on the release (as is the case for v2.13) and a PR like this taking place in the morning in the U.S. will likely muck w/ their schedule in China.

It is squad policy to do a "doc freeze" the day before a GA for this very reason. Anything you can do to address this?

MarkAckert:
I'll think about alternatives - it sounds like `v2.x/master` isn't a good solution, and the current one isn't great either. Making this automation run a manual step would also work but would be something we're liable to forget. I'll keep thinking and in the meantime close this PR. Thanks for the feedback!

PR #3674, opened 2023-12-01
Revert "Reduce string fields by introducing better typing"

MarkAckert:
Reverts zowe/zowe-install-packaging#3617


1000TurquoisePogs (approved):
Good for now, will revisit in next release.

PR #3680, opened 2023-12-06
Include 'zowe.job.suffix' to be compatible with prior configs

1000TurquoisePogs:
At previous advice, at least one extender was using 'zowe.job.suffix' within their component to let users customize their jobname.
As we just added 'zowe' section to components, there was a collision in schemas so after discussion it was found good to just have the zowe schema declare the existence of zowe.job.suffix. Right now we say zowe does nothing with it, and it is for extension use as needed.


MarkAckert (approved):

MarkAckert:
All tests passed; merging in after syncing with base branch.

PR #3708, opened 2024-01-19
API ML v3

achmelo:
Add changes to support java 17

#### PR type
- [ ] Bugfix 
- [ ] Feature 

#### Relevant issues

Fixes 

- 

- [ ] Yes
- [ ] No

---

---

---

---
_multi-line description_

Doc issue/PR number:


balhar-jakub:
metrics-service should be removed from V3 PAX

MarkAckert (approved):

MarkAckert:
Java 17 is throwing errors when keytool imports EBCDIC certificates, which were converted from ASCII to EBCDIC as part of our install scripts. Removing the conversion and keeping certs in ASCII fixes the keytool steps in certificate.sh, but the app framework then throws an error trying to use the ASCII certificates at runtime. After manually converting certificates back to EBCDIC, the app framework worked. APIML services worked with the certificates regardless of EBCDIC vs ASCII (I assume because they load certs from keystore). I'm not sure yet if ASCII certificates in the keystore and EBCDIC certificates outside the keystore will result in trust issues between the services, though a mixed codepage path is probably not a good idea.

Ideally we find out if the `keytool -import` command should accept EBCDIC certificates and submit a bug with IBM or fix our import call. 

If keytool no longer supports EBCDIC, then I think we'd want all the services to consume certificates in a consistent format - EBCDIC or ASCII. I don't know if this is feasible, and if we did decide to change certificate codepages to ASCII we'd need to consider this in the context of upgrade - will old keystores with Java 8 and old EBCDIC certificates continue to work with v3? 

@taban03

achmelo:
> Java 17 is throwing errors when keytool imports EBCDIC certificates, which were converted from ASCII to EBCDIC as part of our install scripts. Removing the conversion and keeping certs in ASCII fixes the keytool steps in certificate.sh, but the app framework then throws an error trying to use the ASCII certificates at runtime. After manually converting certificates back to EBCDIC, the app framework worked. APIML services worked with the certificates regardless of EBCDIC vs ASCII (I assume because they load certs from keystore). I'm not sure yet if ASCII certificates in the keystore and EBCDIC certificates outside the keystore will result in trust issues between the services, though a mixed codepage path is probably not a good idea.
> 
> Ideally we find out if the `keytool -import` command should accept EBCDIC certificates and submit a bug with IBM or fix our import call.
> 
> If keytool no longer supports EBCDIC, then I think we'd want all the services to consume certificates in a consistent format - EBCDIC or ASCII. I don't know if this is feasible, and if we did decide to change certificate codepages to ASCII we'd need to consider this in the context of upgrade - will old keystores with Java 8 and old EBCDIC certificates continue to work with v3?
> 
> @taban03

Because certificates in EBCDIC encoding are used only by the app framework to establish trust, we need to convert only truststore exports. Now all components can start without any manual steps. There are still a few tests failing but these tests are for deprecated components which are going to be removed in v3. 
It would be good to review which certificates are still needed and do not produce those that aren't. It is not necessary within this PR.

tests/sanity/test/apiml/test-api-zosmf-via-gateway.js
```diff
@@ -31,13 +31,13 @@ describe('test api mediation layer zosmf api', function() {
     token = await apiml.login();
   });
 
-  it('should be able to get z/OS Info via the gateway port and endpoint (/api/v1/zosmf/info)', async function() {
+  it('should be able to get z/OS Info via the gateway port and endpoint (/ibmzosmf/api/v1/info)', async function() {
```
1000TurquoisePogs: why has the url for zosmf changed to "ibmzosmf"? this looks like a breaking change not listed on zowe.org/vnext

balhar-jakub: To ibmzosmf it changed years ago, the zosmf was kept only for backwards compatibility. 

But you are correct, it needs to be added to the breaking changes that we intend to remove the zosmf

dkelosky: @balhar-jakub - isn't `zosmf/info` part of the endpoint path? https://docs.zowe.org/v2.6.x/extend/extend-apiml/api-mediation-components-of-url/#url-with-empty-basepath

Although it perhaps seems redundant, `zosmf` is part of the API path and applications use it that way ([such as Zowe CLI](https://github.com/zowe/zowe-cli/blob/master/packages/zosjobs/src/JobsConstants.ts#L112)).

Is this change really necessary?

dkelosky: Would this mean that the swagger doc in z/OSMF UI (which includes `zosmf`) would look different than the swagger docc for z/OSMF when viewed through the API ML catalog UI? 

IBM z/OSMF:
image

adam-wolfe: It looks like the test has been defined this way since March 2022: [commit](https://github.com/zowe/zowe-install-packaging/commit/0ca96cddcfe9326456f26ca075127ca2de644705#diff-d851d7cd9626c7d890cd7305ab614dcfdb99f2747e4667eb3ac6ba37b34d9a93)

balhar-jakub: The swagger documentation wouldn't change, it's already coming from the ibmzosmf definition. 

I created a Q&A in GitHub around this here: https://github.com/zowe/api-layer/discussions/3324 

The ibmzosmf static definition was added as part of the cleanup for zosmf to follow the conventions required by conformance. I can't find the exact commit, but the ibmzosmf lived within the context at least as soon as

https://github.com/zowe/api-layer/commit/e27dc32e4d3b1095033f73a203945a1e71723085

which is over 4 years ago and the discussion about this is living here: https://github.com/zowe/api-layer/issues/665

The intention as far as I understand it based on other discussions wasn't to keep these two forever but do the change on the major version boundaries. We missed the opportunity for V2 and as such the intention from our side was to do it on V3 boundary.

MarkAckert:
All builds green, merging in. Thank you!

PR #3716, opened 2024-01-26
fix: update default service id for z/osmf in api layer

pablocarle:
Updates the default z/OSMF service Id to `ibmzosmf` in v3

#### PR type

#### Relevant issues

Related to https://github.com/zowe/api-layer/issues/3295

- Set default to `ibmzosmf` in the exampe zowe.yaml for v3

- [x] Yes
- [ ] No

In existing installations moving to v3, they need to make sure to update the serviceId set for z/OSMF in the gateway configuration from default `zosmf` to `ibmzosmf`, the first one will no longer exist.

---

---

---

---
_multi-line description_

Doc issue/PR number:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 # Change Log
 
 All notable changes to the Zowe Installer will be documented in this file.
+
+## `3.0.0`
+
+### Breaking Changes
+
+- `zowe.yaml` changed its default z/OSMF Service ID definition from `zosmf` to `ibmzosmf`, which may impact Zowe Clients. For more information on this change, please see the API Mediation Layer's 3.0.0 Breaking Changes.
+
 <!--Add the PR or issue number to the entry if available.-->
 ## `2.14.0`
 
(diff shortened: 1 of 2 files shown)


MarkAckert:
Will the service ID change break anything? Either extenders or API clients? Or is this internal-only?  If this is a breaking change, it's fine for v3, but we should make sure to capture it in the CHANGELOG.

pablocarle:
> Will the service ID change break anything? Either extenders or API clients? Or is this internal-only? If this is a breaking change, it's fine for v3, but we should make sure to capture it in the CHANGELOG.

The breaking change happens in the API Mediation Layer, as of v2, there were two zosmf static definitions: with service Id `zosmf` and `ibmzosmf`, the breaking change is that `zosmf` will no longer exist, so clients such as the CLI and others should be made aware of this, to verify they are not going directly to `zosmf`.

It will be included in the changelog for v3 in api-layer, should I aad it also manually to https://github.com/zowe/zowe-install-packaging/blob/v3.x/staging/CHANGELOG.md ?

MarkAckert:
I included a comment in the CHANGELOG pointing users to the APIML doc since the configuration change is coming through zowe.yaml. Thanks for clarifying!

MarkAckert (approved):

PR #3721, opened 2024-01-31
bin/libs/string.sh: trim update

### LINKED ISSUE #3720: bin/libs/string.sh trim too aggressive
**Describe the bug**
`trim` is expected to remove leading and trailing blanks only, but [implementation](https://github.com/zowe/zowe-install-packaging/blob/453dd6937b63f23851a1412aa296657a03a14314/bin/libs/string.sh#L25) via `xargs` is removing more spaces than expected.

**To Reproduce**
1. Use `trim` in shell script with parameter e.g. `  A         B  ` (10 spaces between `A` and `B`)
2. Result is `A B` (one space inside)
3. Use `trim()` from `bin/libs/strings.js` and the result will be `A          B` (10 spaces between `A` and `B`)

**Expected behavior**
Same behavior for both functions. JS version seems to be the correct one.

Martin-Zeithaml:

#### PR type
- [x] Bugfix 

#### Relevant issues

Fixes https://github.com/zowe/zowe-install-packaging/issues/3720

`trim " A          B "` returns `"A          B"`

- [x] Yes

### DIFF
--- a/bin/libs/string.sh
+++ b/bin/libs/string.sh
@@ -22,7 +22,7 @@ trim() {
     input=${1}
   fi
 
-  echo "${input}" | xargs
+  echo "${input}" | sed -e 's/^[[:space:]]*//;s/[[:space:]]*$//'
 }
 
 ###############################


Martin-Zeithaml:
## Test

```
. /zowe/runtime/bin/libs/string.sh
for str in  ""\
            " "\
            "  A"\
            "A  "\
            " A B"\
            "   A    B   "\
            " Ai\niB "\
            "\t\tA\t\tB\t\t"\
            "\tA\tB\t"; do         
  res=$(trim "${str}")
  echo "'${str}' >>> '${res}'"
done
```
## With fix

```
'' >>> ''
' ' >>> ''
'  A' >>> 'A'
'A  ' >>> 'A'
' A B' >>> 'A B'
'   A    B   ' >>> 'A    B'
' Ai
iB ' >>> 'Ai
iB'
'               A               B               ' >>> 'A                B'
'       A       B       ' >>> 'A        B'
```

## Without fix
```
'' >>> ''
' ' >>> ''
'  A' >>> 'A'
'A  ' >>> 'A'
' A B' >>> 'A B'
'   A    B   ' >>> 'A B'
' Ai
iB ' >>> 'Ai iB'
'               A               B               ' >>> 'A B'
'       A       B       ' >>> 'A B'
```

MarkAckert (approved):

MarkAckert:
Code change looks good, Is there a chance any existing code calling `trim()` depended on the old, unintentional stripping of whitespace between characters in the string?

Martin-Zeithaml:
@MarkAckert This was my concern too, that removing the bug will introduce 10 new. I was looking into the code how the shell `trim` is used and seems the original intention was to use it only for words without spaces but surrounded by spaces. I did also small testing, some `zwe` commands, where `trim` is used and no problems were found. I believe this change is safe, otherwise I would delete this PR.

PR #3746, opened 2024-02-28
pswi schema validation tests

MarkAckert:

#### PR type
This PR adds tests which attempt to detect desync between the zowe.yaml schema files and the zowe.yaml which is embedded and configured as part of the z/OSMF workflows. 

This PR adds tests which cover the following cases:

* Comments are desync'd. This test fails fast (early in file), and could be brittle. Please review and comment on if it's worth keeping.
* Fields and values are desync'd. This will cover anything added, deleted, or changed in the example-zowe.yaml, relative to the current workflow with no special configuration provided.
* Simple (default) workflow zowe.yaml schema validation. Currently, the wf zowe.yaml is invalid due to missing java.home and node.home defaults and the tests are adjusting for that.
* Branch coverage for workflow zowe.yaml schema validation. There's some simple if/else logic embedded in the workflow VTL, and this test exercises all possible branching paths to pre-emptively cover an ugly edge case popping up.

In both schema validation cases, we are using a popular JSON schema validator, `ajv`. This is not the same schema validation happening in the `zwe` configMgr tool, but a few manual tests show their results look to be the same. I think these schema tests are very high confidence, but if we want perfect confidence results, we need new server-side configMgr-driven schema validation tests.

To run these tests:
`npm install` and `npx mocha` or `npm run test`. No server-side dependencies required.


MarkAckert:
We went in another direction for PSWI schema validation. We could come back to this if those tests need shoring up, but right now they look good.

PR #3748, opened 2024-03-01
Fix the SMP/e build

MarkAckert:
Adds new members to the SMPMCS for the v3 JCL work.

No update should be required for PSWI; the new members should be picked up through the SMP/e build.


muzikovam (approved):

struga0258 (approved):

MarkAckert:
Merging, status check passed after re-run.

PR #3718, opened 2024-01-29
Migrate 'zwe init' actions to templated JCL

1000TurquoisePogs:
This PR solves a common complaint; Admins need more system changes represented in JCL so it works with their automation, they can understand what Zowe is doing, and can view the success or failure more easily.

It solves this by taking the system actions out of each "zwe init" command, and placing them into a related JCL.
Each new and existing JCL is also simplified by removing as many "SET" commands as possible to instead replace with `{templates}` that reference values found in zowe.yaml

There is a new job called ZWEGENER which when run will take each JCL template, resolve it with verified zowe.yaml content, and place the result into the customer jcl lib.

ZWEGENER needs input about where your zowe.yaml and schema is, as well as the loadlib in order to find the logic that makes it be able to read zowe.yaml, configmgr-rexx.

ZWEGENER can be run by itself to provide a reduced reliance on USS to setup zowe, but there is also a new command "zwe init generate" which will call it too. This now gets automatically called in "zwe init" as a prereq for other "zwe init" subcommands.

Resolves https://github.com/zowe/zowe-install-packaging/issues/1038
Resolves https://github.com/zowe/zowe-install-packaging/issues/2481
Resolves https://github.com/zowe/zowe-install-packaging/issues/2482
Resolves https://github.com/zowe/zowe-install-packaging/issues/2487
Resolves https://github.com/zowe/zowe-install-packaging/issues/2451
Resolves https://github.com/zowe/zowe-install-packaging/issues/3065
Resolves https://github.com/zowe/zowe-install-packaging/issues/3634
Resolves https://github.com/zowe/zowe-install-packaging/issues/3562
Resolves https://github.com/zowe/zowe-install-packaging/issues/3416
Resolves https://github.com/zowe/zowe-install-packaging/issues/3641
Resolves https://github.com/zowe/zowe-install-packaging/issues/2470
Resolves https://github.com/zowe/zowe-install-packaging/issues/3330

PR #3756, opened 2024-03-12
Remove node usage in startup when using configmgr

1000TurquoisePogs:
This PR should allow users to start up zowe without having nodejs as long as they aren't using app-server.

This PR does not help you install zowe without nodejs. That's instead something you can do with https://github.com/zowe/zowe-install-packaging/pull/3718

While investigating the code, I realized that the actions of `config-converter` aren't difficult to replicate in plain JS in the same function we were calling it.

There are 3 things it appears to provide:
* .zowe.json = is a conversion of zowe.yaml with some extras
* .zowe-<ha-instance>.json = adds ha changes on top of that
* componentname/.configs-<ha-instance>.json = adds a "configs" area on top of that, per component that has one.

So since configmgr already transforms the yaml into json, some deep object merges can replace this behavior.

Then with the former config-converter code removed, I checked for remaining uses of node and saw that several other node-based functions were unused in our codebase so they can safely be removed.

TESTING: check the contents of workspace/.env before-after this PR.
If you do a diff to compare the old and new json, you will notice 2 things:
1) items are reordered. this is fine.


Martin-Zeithaml:
## Test
1) No `node` in user's profile
2) No `node` in yaml file
3) App-server disabled

```
$ echo $NODE_HOME

$ cat 3756.yaml | grep -i node

$ cat 3756.yaml | grep -C 1 -i server

  app-server:
    enabled: false

$ zwe start -c 3756.yaml
Temporary directory '/tmp/.zweenv-5074' created.
===============================================================================
>> STARTING ZOWE

>> Job ZWE8SV is started successfully.

Temporary directory /tmp/.zweenv-5074 removed successfully.
```

Martin-Zeithaml (approved):

PR #3760, opened 2024-03-14
add different message ids for VSAM errors

taban03:
Addresses https://github.com/zowe/api-layer/issues/3329 

#### PR type
- [ ] Bugfix 
- [ ] Feature 

#### Relevant issues

Fixes 

- 

- [ ] Yes
- [ ] No

---

---

---

---
_multi-line description_

Doc issue/PR number:

### DIFF
--- a/bin/commands/init/vsam/.errors
+++ b/bin/commands/init/vsam/.errors
@@ -1,3 +1,5 @@
+ZWEL0148E|148|%s (%s) is not defined in Zowe YAML configuration file.
+ZWEL0149E|149|%s (%s) is not defined in Zowe YAML configuration file.
 ZWEL0157E|157|%s (%s) is not defined in Zowe YAML configuration file.
 ZWEL0300W||%s already exists. This data set member will be overwritten during configuration.
 ZWEL0301W||%s already exists and will not be overwritten. For upgrades, you must use --allow-overwrite.
--- a/bin/commands/init/vsam/.errors
+++ b/bin/commands/init/vsam/.errors
@@ -1,5 +1,3 @@
-ZWEL0148E|148|%s (%s) is not defined in Zowe YAML configuration file.
-ZWEL0149E|149|%s (%s) is not defined in Zowe YAML configuration file.
 ZWEL0157E|157|%s (%s) is not defined in Zowe YAML configuration file.
 ZWEL0300W||%s already exists. This data set member will be overwritten during configuration.
 ZWEL0301W||%s already exists and will not be overwritten. For upgrades, you must use --allow-overwrite.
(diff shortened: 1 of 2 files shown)


bin/commands/init/vsam/index.sh
```diff
@@ -44,7 +44,7 @@ vsam_volume=
 if [ "${vsam_mode}" = "NONRLS" ]; then
   vsam_volume=$(read_yaml "${ZWE_CLI_PARAMETER_CONFIG}" ".zowe.setup.vsam.volume")
   if [ -z "${vsam_volume}" ]; then
-    print_error_and_exit "Error ZWEL0157E: Zowe Caching Service VSAM data set Non-RLS volume (zowe.setup.vsam.volume) is not defined in Zowe YAML configuration file." "" 157
+    print_error_and_exit "Error ZWEL0148E: Zowe Caching Service VSAM data set Non-RLS attribute (zowe.setup.vsam.volume) is not defined in Zowe YAML configuration file." "" 148
```
dkelosky:
```suggestion
    print_error_and_exit "Error ZWEL0148E: Zowe Caching Service VSAM data set volume (zowe.setup.vsam.volume) is not defined in Zowe YAML configuration file." "" 148
```

bin/commands/init/vsam/.errors
```diff
@@ -1,3 +1,5 @@
+ZWEL0148E|148|%s (%s) is not defined in Zowe YAML configuration file.
```
dkelosky: it's interesting to see the error message template implemented this way... given that it is the exact same as 149 & 157, it makes me wonder if they should indeed be the exact same message. what do you think?

taban03: Yes, I was also concerned about that, and maybe you're right, they could be merged into a single message. Although is also true that the causes are different for all the 3 messages.

taban03: But the failure cause is different, isn't it? I guess that it makes sense to keep them under different msg ID, but maybe I miss something

MarkAckert: I agree with Dan, if we have any "property is missing in zowe.yaml" error, I'd expect the message to share the same ID, with specific error details in the output. If it's a post-processing or derived property error I think there's an argument for that case to have it's own ID.

Martin-Zeithaml: 1) Agree with @dkelosky - this is not a good practice, there is already an error message 157, which is generic and can be used for different problems.
2) This pull request [#3718](https://github.com/zowe/zowe-install-packaging/pull/3718) will remove the shell version  of this command anyway.

taban03: Thanks @Martin-Zeithaml  for the notice about the other PR. Do you know how these messaged will be handled by these changes? Cause I will revert now the changes, but I'm not sure how these kind of messages like:
"Error ZWEL0157E: Zowe Caching Service VSAM data set Non-RLS volume (zowe.setup.vsam.volume) is not defined in Zowe YAML configuration file.
will be then displayed

Martin-Zeithaml: [Like this](https://github.com/zowe/zowe-install-packaging/blob/2188d7db55adde994b5677da01615e81f771001c/bin/commands/init/vsam/index.ts#L45-L51).

MarkAckert (approved):

MarkAckert:
Approved and merging in knowing #3718 will address this longer term.

Issue #3123, opened 2022-10-24
Support adding Pull Request Labels to run additional z/OS builds


MarkAckert:
Zowe PAX build pipelines set the SMP/e and PSWI build options to 'false' by default to save time on CI builds which run as part of pull request reviews and requirements. Certain pull requests may impact the packaging/install of Zowe, which requires testing the SMP/e and PSWI builds. Right now this is a manual effort - we should be able to label pull requests with some value such as "ci-extended", "ci-smpe", or "ci-pswi" - which then causes the longer SMP/e and PSWI builds to run as part of CI for that particular pull request.

Done criteria:

- PR labels can be added to zowe-install-packaging builds which modify the following:
  * Build (None, PAX [default], SMP/e, PSWI)
  * Test (None, Integration [default], SMP/e, Extended)
  * If Build=None, then Test=None by default.
- Github Actions are updated to reflect the current workflow criteria. If we open a PR, then set label Build=None, the build workflow check and call-integration workflow check should both be removed or set to passing.
  * Ideally, this should happen without requiring an empty commit or other event to re-trigger the pipelines on the PR.
 
- If possible, core functionality should be generalized in our shared-actions pipelines for use by other squads:
  * https://github.com/zowe-actions/shared-actions 
     * Primary set of zowe's workflow libraries
  * https://github.com/zowe-actions/octorelease
     * CLI squad's additional workflow libraries. Check with the squad for more info
 
- Some ideas on setting this up were explored here: https://github.com/zowe/zowe-chat/pull/182

PR #3766, opened 2024-03-19
Use labels to adjust CI behavior

MarkAckert:
Addresses #3123 

This PR updates existing build and packaging automation to modify their behavior based on labels add to the PR. The labels are only effective for builds triggered from pull requests (manually triggered builds will ignore all labels). This PR does not protect you from invalid combinations of build and test; i.e. Build: PAX and Test: SMP/e will fail.

 The following labels and features are added:

Build Labels:
* Build: None
* __Build: PAX__ (default)
* Build: SMPE
* Build: PSWI
* Build: Kubernetes
* Build: Debug-Remote

`Build: None` will supersede other build labels; the rest of the build labels act by composition. `Build: PSWI` auto-includes SMP/e, as SMP/e is required to build the PSWI. Using both `Build: SMPE` and `Build: PSWI` will create just the one PSWI. `Build: Debug-Remote` is a special label that can retain some server-side build information for additional debugging. Please check with the Systems squad before adding this label.

Test Labels:
* Test: None
* __Test: Basic__ (default)
* Test: SMPE
* Test: Extended
* Test: Silly (Release Tests; about 5 hours to complete)

Tests do _not_ work by composition, and instead will default to the smallest test suite by specified label. Test composition requires far more refactoring, so the work on that is deferred pending requirement from other squads. The decision to default to the least amount of testing is to avoid locking up our systems unnecessarily. 

Test triggers:

Adding and removing labels _will not trigger a build_, as this can kick off too many long-running builds in sequence. Instead, commenting on the pr with `/ci` will trigger the build. Editing the message in-place will also trigger a build. The expected developer workflow is to open a PR, decide if any non-default labels better apply to their PR, add them, and then trigger a build with `/ci`.

### DIFF
--- a/.github/workflows/build-packaging.yml
+++ b/.github/workflows/build-packaging.yml
@@ -92,7 +92,7 @@ jobs:
   
   regular-build:
 
-    if: ${{ github.event_name != 'issue_comment' || needs.pr-comment-check.outputs.run_build == 'true' }}
+    if: ${{ github.event_name != 'issue_comment' }} || steps.pr-comment-check.outputs.run_build 
     runs-on: ubuntu-latest
     needs: check-permission
     steps: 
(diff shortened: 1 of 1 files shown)


MarkAckert:
@zowe-robot ci

MarkAckert:
@zowe-robot ci

struga0258 (approved):

PR #3773, opened 2024-03-25
Fix PSWI build trigger

MarkAckert:
Corrects build logic so PSWI/SMPE selected with empty artifactory references will build everything from scratch.


struga0258 (approved):

MarkAckert:
Merging in; build passed manually

PR #3812, opened 2024-04-25
Reduce shell count by one when configmgr used by using pipe directly

1000TurquoisePogs:
When inspecting for optimizations and reduction of shell usage for TMPDIR troubleshooting, I noticed that there's some unneeded use of shell wrappers for running each server.

In the code today, each time a start.sh of a server is run we spawn a shell to run that in, and then pipe its contents to ANOTHER shell so that we can put `wait` at the end.

We could have just spawned one shell with that pipe, instead of a shell-within-a-shell.
This looks to be largely from the old non-configmgr code, and configmgr can just further optimize.

The process tree goes from 7 processes in the case of zss:
before

To 6:
after

An effort here can further reduce it to 5 https://github.com/zowe/launcher/issues/109
With both efforts combined, the tree of a given Zowe server would just be
* Launcher
    * configmgr
        * server's own start script 
            * server itself
            
But today it is

* Launcher
    * zwe
        * configmgr
            * shell wrapper being removed in this PR 
                * server's own start script 
                    * server itself

How to test this PR?
If the servers come up the same as they did before, success.
If a server is crashed (intentionally or unintentionally), launcher should have the same restart behavior as before.

### DIFF
--- a/bin/commands/internal/start/component/index.ts
+++ b/bin/commands/internal/start/component/index.ts
@@ -12,10 +12,12 @@
 import * as std from 'cm_std';
 import * as os from 'cm_os';
 import * as zos from 'zos';
+import * as xplatform from 'xplatform';
 import * as common from '../../../../libs/common';
 import * as config from '../../../../libs/config';
 import * as shell from '../../../../libs/shell';
 import * as varlib from '../../../../libs/var';
+import * as stringlib from '../../../../libs/string';
 import * as java from '../../../../libs/java';
 import * as fs from '../../../../libs/fs';
 import * as component from '../../../../libs/component';
@@ -93,8 +95,27 @@ export function execute(componentId: string, runInBackground: boolean=false) {
         } else {
           // wait for all background subprocesses created by bin/start.sh exit
           // re-source libs is necessary to reclaim shell functions since this will be executed in a new shell
-          //TODO does this do the same as the shell script before it?
-          shell.execSync('sh', '-c', `cd ${COMPONENT_DIR} ; cat "${fullPath}" | { echo ". \"${ZOWE_CONFIG.zowe.runtimeDirectory}/bin/libs/configmgr-index.sh\"" ; cat ; echo; echo wait; } | /bin/sh`);
+          const startScriptContents = `cd ${COMPONENT_DIR} ; . "${ZOWE_CONFIG.zowe.runtimeDirectory}/bin/libs/configmgr-index.sh" ; ${xplatform.loadFileUTF8(fullPath, xplatform.AUTO_DETECT)} ; wait;`;
+          const pipeArray = os.pipe();
+          if (!pipeArray) {
+            //TODO error message
+            common.printFormattedError("ZWELS", "zwe-internal-start-component", `Error ZWEL???E: Could not create pipe.`);    
+            return;
+          }
+          //TODO this will not work with unicode codepoints longer than a byte
+          const buf = new ArrayBuffer(startScriptContents.length);
+          const view = new Uint8Array(startScriptContents.length);
+          const ebcdicString = stringlib.asciiToEbcdic(startScriptContents);
+          for (let i = 0; i < startScriptContents.length; i++) {
+            view[i] = ebcdicString.charCodeAt(i);
+          }
+
+          os.write(pipeArray[1], buf, 0, startScriptContents.length);
+          os.close(pipeArray[1]);
+          os.exec(['/bin/sh'],
+                  {block: true, usePath: true, stdin: pipeArray[0]});
+          os.close(pipeArray[0]);
+          
         }
       } else {
         common.printFormattedError("ZWELS", "zwe-internal-start-component", `Error ZWEL0172E: Component ${componentId} has commands.start defined but the file is missing.`);
--- a/bin/commands/internal/start/component/index.ts
+++ b/bin/commands/internal/start/component/index.ts
@@ -104,7 +104,7 @@ export function execute(componentId: string, runInBackground: boolean=false) {
           }
           //TODO this will not work with unicode codepoints longer than a byte
           const buf = new ArrayBuffer(startScriptContents.length);
-          const view = new Uint8Array(startScriptContents.length);
+          const view = new Uint8Array(buf);
           const ebcdicString = stringlib.asciiToEbcdic(startScriptContents);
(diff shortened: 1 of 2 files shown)


bin/commands/internal/start/component/index.ts
```diff
@@ -93,8 +95,27 @@ export function execute(componentId: string, runInBackground: boolean=false) {
...
-          //TODO does this do the same as the shell script before it?
-          shell.execSync('sh', '-c', `cd ${COMPONENT_DIR} ; cat "${fullPath}" | { echo ". \"${ZOWE_CONFIG.zowe.runtimeDirectory}/bin/libs/configmgr-index.sh\"" ; cat ; echo; echo wait; } | /bin/sh`);
+          const startScriptContents = `cd ${COMPONENT_DIR} ; . "${ZOWE_CONFIG.zowe.runtimeDirectory}/bin/libs/configmgr-index.sh" ; ${xplatform.loadFileUTF8(fullPath, xplatform.AUTO_DETECT)} ; wait;`;
+          const pipeArray = os.pipe();
+          if (!pipeArray) {
+            //TODO error message
+            common.printFormattedError("ZWELS", "zwe-internal-start-component", `Error ZWEL???E: Could not create pipe.`);    
+            return;
+          }
+          //TODO this will not work with unicode codepoints longer than a byte
```
1000TurquoisePogs, line 104: Since the buffer contains just paths & script content, unless the script is containing things not found in ascii (really unusual aside from .... copyright symbols??) then I guess the only possibility for this to be an issue is if someone has pathnames not found in ASCII.

So I tested how we do with pathnames containing Japanese today, without this PR
PKCS12 gen fails for some reason, but assuming you already have a keyring...
Init passes, prepare passes, and app-server/zss start up.
But all the java servers fail to load with `Error: Could not find or load main class org.springframework.boot.loader.PropertiesLauncher`
So, we have work to do elsewhere too but yeah it looks like the majority of zwe actually does handle unicode alright, so should fix this before merging...

Martin-Zeithaml: IBM recommends to use [POSIX portable character set](https://www.ibm.com/docs/en/zos/3.1.0?topic=system-path-path-name) for paths, otherwise the result is unpredictable. I would not be worried about this.

If the scope of `zwe internal` is still internal, we as a zowe community have under control, what will be in the start-up scripts. I would not be worried about non-ascii chars too.

1000TurquoisePogs: Thats a good point. I guess it's good to go as-is then.

1000TurquoisePogs:
I tested and confirmed if i kill a server process the launcher understands it needs to restart it. so, seems like this code causes no regression.

bin/commands/internal/start/component/index.ts
```diff
@@ -93,8 +95,27 @@ export function execute(componentId: string, runInBackground: boolean=false) {
         } else {
           // wait for all background subprocesses created by bin/start.sh exit
           // re-source libs is necessary to reclaim shell functions since this will be executed in a new shell
-          //TODO does this do the same as the shell script before it?
-          shell.execSync('sh', '-c', `cd ${COMPONENT_DIR} ; cat "${fullPath}" | { echo ". \"${ZOWE_CONFIG.zowe.runtimeDirectory}/bin/libs/configmgr-index.sh\"" ; cat ; echo; echo wait; } | /bin/sh`);
+          const startScriptContents = `cd ${COMPONENT_DIR} ; . "${ZOWE_CONFIG.zowe.runtimeDirectory}/bin/libs/configmgr-index.sh" ; ${xplatform.loadFileUTF8(fullPath, xplatform.AUTO_DETECT)} ; wait;`;
+          const pipeArray = os.pipe();
+          if (!pipeArray) {
+            //TODO error message
+            common.printFormattedError("ZWELS", "zwe-internal-start-component", `Error ZWEL???E: Could not create pipe.`);
```
Martin-Zeithaml: New error message or use existing. [ZWEL0064E](https://github.com/zowe/launcher/blob/80184b29056af9b3b70c5c7ddef317562e917fc7/src/msg.h#L80) is generic, but referring to command.

Martin-Zeithaml:
Working fine on our system. I have noticed this `node` message:
```
...
ZWESVUSR INFO ZWEL0069I Configuration is valid
ZWESVUSR INFO ZWEL0058I WORKSPACE_DIR is '/zowe/workspace'
FSUM7422 node is not found 
Node found in NODE_HOME
ZWESVUSR INFO (zwe-internal-start-prepare) Zowe version: v2.16.0       
ZWESVUSR INFO (zwe-internal-start-prepare) build and hash: PR-3812#5131 (008872531f2a6ba28a8264e6f05266422d246d7b)
...
```

1000TurquoisePogs:
I think the "node not found" came from the PR where node was removed as a prereq. Some code somewhere must be checking for it and then complaining, yet continuing anyway.

1000TurquoisePogs:
Error message assigned.
Node warning found and corrected here https://github.com/zowe/zlux-app-server/pull/301

bin/commands/internal/start/component/index.ts
```diff
@@ -93,8 +95,26 @@ export function execute(componentId: string, runInBackground: boolean=false) {
         } else {
           // wait for all background subprocesses created by bin/start.sh exit
           // re-source libs is necessary to reclaim shell functions since this will be executed in a new shell
-          //TODO does this do the same as the shell script before it?
-          shell.execSync('sh', '-c', `cd ${COMPONENT_DIR} ; cat "${fullPath}" | { echo ". \"${ZOWE_CONFIG.zowe.runtimeDirectory}/bin/libs/configmgr-index.sh\"" ; cat ; echo; echo wait; } | /bin/sh`);
+          const startScriptContents = `cd ${COMPONENT_DIR} ; . "${ZOWE_CONFIG.zowe.runtimeDirectory}/bin/libs/configmgr-index.sh" ; ${xplatform.loadFileUTF8(fullPath, xplatform.AUTO_DETECT)} ; wait;`;
+          const pipeArray = os.pipe();
+          if (!pipeArray) {
+            common.printFormattedError("ZWELS", "zwe-internal-start-component", `Error ZWEL064E: failed to run command os.pipe - Connot start component ${componentId}`);
```
Martin-Zeithaml: Typo `Connot`

Martin-Zeithaml (approved):

PR #3815, opened 2024-05-01
TEST for regression

1000TurquoisePogs:
In the JCL pr, there was some regression on the automated test but it's not clear what from the diffs I saw. So, I'm going to replay some of the changes to isolate it.


1000TurquoisePogs:
Solved: I had confused the ca label with the cert label.

PR #3816, opened 2024-05-01
Changes to sed command to use pr 107 of launcher

1000TurquoisePogs:
PR https://github.com/zowe/launcher/pull/107 which is primarily for the JCL PR, can be merged prior to it as long as this PR is merged which makes the current work compatible with it.

The incompatibility fixed here is just that the DISP=SHR argument is on a different line, so the sed command needs adjustment to that.


JoeNemo (approved):
Approved.

1000TurquoisePogs:
The test at https://github.com/zowe/zowe-install-packaging/pull/3817 passed so I will merge this instead of running the same test twice.

PR #3826, opened 2024-05-08
Node v20 Tests

MarkAckert:
Adds node v20 to backend systems and creates new tests for Node v20. Adds Node v20 to the nightly/release test suite.

Node 20 passing test:
https://github.com/zowe/zowe-install-packaging/actions/runs/9007882831/job/24748714570


1000TurquoisePogs (approved):
I note in the log files that node 20 is used at runtime. Great!
But, also the logs have it so that node v16 is specified in the beginning, and then overridden to 20.

That's not an issue right now. But by september that should be switched to v18 default because v16 will be out of support in september I believe.

PR #3829, opened 2024-05-13
Test - launcher without zwe

1000TurquoisePogs:
Test build which uses launcher proof of concept where it calls configmgr binary directly to shorten startup time.


1000TurquoisePogs:
Closing as test complete.

PR #3873, opened 2024-06-25
ZWEGEN00 split CONFIG entry over multiple lines

1000TurquoisePogs:
Current JCL branch can leave CONFIG entry of ZWESLSTC without template resolution if the configuration length exceeds jcl line length.

This PR attempts to correct it by using some rexx to process ZWESLSTC deliberately for splitting the config the way we want.

CONFIG= works like this

one entry:
CONFIG=FILE(/foo) 

multiple entries:
CONFIG=FILE(/foo):\
FILE(/foo2):\
FILE(/foo3)

Additionally, I saw the opportunity to force "defaults.yaml" into the config reading.
It should not end up in the CONFIG= output, but should be considered during the template resolution.
To add it, I detect runtimeDirectory via finding that in the first line of the schema entries, since they must be in the runtime directory.


files/SZWEEXEC/ZWEGEN00
```diff
@@ -381,6 +388,55 @@ if card.0 > 0 then do
...
+  ExitWithRC(8)
+end
+if ReadFromDataSet('zwejcl') > 0 then do
+  ExitWithRC(8)
+end
+if FreeByDD('zwejcl') > 0 then do
+  ExitWithRC(8)
+end
+do j = 1 to !contentToRead.0
+  if COMPARE(contentToRead.j, 'CONFIG={ZWE_CLI_PARAMETER_CONFIG}') = 0 then do
```
Martin-Zeithaml: Variable name is `!contentToRead.j`

files/SZWEEXEC/ZWEGEN00
```diff
@@ -381,6 +388,55 @@ if card.0 > 0 then do
...
+    say 'Defining CONFIG entry for ZWESLSTC'
+    currentline = j
+    ReadFromDataSet('myconfig')
```
Martin-Zeithaml: Cannot be standalone, it is returning a value, which is interpreted. 
It has to be `if ReadFromDataSet('myconfig') = 0 then do...`

files/SZWEEXEC/ZWEGEN00
```diff
@@ -381,6 +388,55 @@ if card.0 > 0 then do
...
+      element = type'('location')'
+      if COMPARE(type, 'PARMLIB') = 0 then do
+        element = 'PARMLIB('location'(ZWEYAML))'
+      end
+      if j = 1 then do
+        cmd = 'LINE 'currentline' = CONFIG='element
+        if !contentToRead.0 > 1 then do
+          cmd = cmd':\'
+        end
+        address ISREDIT cmd
```
Martin-Zeithaml: This `ISREDIT` must be invoked in similar way as `ZWECHG`: when editing specific member.

1000TurquoisePogs:
Merge back into main branch

Issue #3872, opened 2024-06-21
Certificate issue after update to version 2.16


rudatp:
We are using SAF keyrings, yaml section:

certificate:
  keystore:
    type: JCERACFKS
    file: "safkeyring:////STCZOWE/ZOWE"
    password: "password"
    alias: "ZOWE25"
  truststore:
    type: JCERACFKS
    file: "safkeyring:////STCZOWE/ZOWE"
    password: "password"
  pem:
    key:
    certificate:
    certificateAuthorities:
    - "safkeyring:////STCZOWE/CA1"
    - "safkeyring:////STCZOWE/CA2"

I updated ZOWE today to version 2.16 and startup fails with:

Validity Exceptions(s) with object at
  Validity Exceptions(s) with object at /zowe
    Schema at '/zowe/certificate' invalid
      not oneOf schemas at '/zowe/certificate' are valid, 0 are
        Validity Exceptions(s) with object at /zowe/certificate
          Validity Exceptions(s) with object at /zowe/certificate/keystore
            unequal constant value at /zowe/certificate/keystore/type; expecting value 'PKCS12' of type 'string'
          Validity Exceptions(s) with object at /zowe/certificate/truststore
            unequal constant value at /zowe/certificate/truststore/type; expecting value 'PKCS12' of type 'string'
          Validity Exceptions(s) with object at /zowe/certificate/pem
            type 'null' not permitted at /zowe/certificate/pem/key; expecting type 'string'
            type 'null' not permitted at /zowe/certificate/pem/certificate; expecting type 'string'
        Validity Exceptions(s) with object at /zowe/certificate
          Validity Exceptions(s) with object at /zowe/certificate/pem
            type 'null' not permitted at /zowe/certificate/pem/key; expecting type 'string'
            type 'null' not permitted at /zowe/certificate/pem/certificate; expecting type 'string'

Can someone help?

Martin-Zeithaml:
Hi,
the problem is, there is no value provided for `/zowe/certificate/pem/key` and `zowe/certificate/pem/certificate`. The schema validation is strict and requires a value. To avoid the schema validation problem, remove unused fields:
```
certificate:
  keystore:
    type: JCERACFKS
    file: "safkeyring:////STCZOWE/ZOWE"
    password: "password"
    alias: "ZOWE25"
  truststore:
    type: JCERACFKS
    file: "safkeyring:////STCZOWE/ZOWE"
    password: "password"
  pem:
[...]

PR #3874, opened 2024-06-26
Allow key/certificate to be null for pem

1000TurquoisePogs:
v2.16.0 introduced a regression described in https://github.com/zowe/zowe-install-packaging/issues/3872 and https://github.com/zowe/zowe-common-c/issues/458 in which a bug fix itself caused an issue because there was dependence upon the bug that was fixed.

This should work-around it for the specific situation seen: having a `pem` section when using keyrings.


Martin-Zeithaml (approved):
Ok

PR #3880, opened 2024-07-01
Update typedef of latest configmgr

1000TurquoisePogs:
configmgr currently has 2 new functions that are not represented in the type definition file.


Martin-Zeithaml (approved):
Ok

PR #3884, opened 2024-07-01
Improve 121  and 122 messages to clarify choices

1000TurquoisePogs:
Sometimes users have reported errors when doing zowe setup because they skipped a step in defining the location of java and node.
The error that can print out, ZWEL0122E and ZWEL0121E, is helpful but incomplete.
This PR just improves the message to make it easier to understand how to fix.


Martin-Zeithaml (approved):
OK.

PR #3885, opened 2024-07-01
Updated help output for accuracy and yaml syntax

1000TurquoisePogs:
This PR contains updates to the zwe embedded `--help` documentation, such as typo fixes, missing information, and yaml syntax markdown hinting.


Martin-Zeithaml (approved):
If the zowe example is confusing for users, we will have to change the help command behavior and print only `` ``` `` instead of `` ```yaml ``

PR #3886, opened 2024-07-01
Remove unused convert script

1000TurquoisePogs:
While working on PR 3718, this block of code was spotted as dead.
It was used prior to v2.16.0, but in that release it was replaced with some QJS code that did the same thing.
I have not found any references to this function in the current codebase and testing of PR 3718 has revealed it to be unused.


Martin-Zeithaml (approved):
Ok

PR #3890, opened 2024-07-02
Add JCL samples from 3718

1000TurquoisePogs:
This PR adds all JCL samples and JCL template resolution code present in https://github.com/zowe/zowe-install-packaging/pull/3718

This PR does not invoke these samples in any way.
Rather, since these samples have shown to work for many users, this PR is for shipping these samples as merely samples. Then, they can be used manually or studied, or ignored, without affecting any existing code.


Martin-Zeithaml (approved):
Ok

PR #3909, opened 2024-07-11
ZWEINSTL update

Martin-Zeithaml:
Minor update of ZWEINSTL even it is not used during install & config now.

But if used by customer, hopefully this updated version will provide better experience.


MarkAckert (approved):
ran modified commands manually to verify them, lgtm

PR #3917, opened 2024-07-17
fix defaults for keyring config

skurnevich:

#### PR type
- [x] Bugfix 
- [ ] Feature 

#### Relevant issues

Fixes 

- 

- [ ] Yes
- [ ] No

---

---

---

---
_multi-line description_

Doc issue/PR number:

### DIFF
--- a/files/defaults.yaml
+++ b/files/defaults.yaml
@@ -70,15 +70,15 @@ zowe:
 
       # Distinguished name for Zowe generated certificates.
       dname: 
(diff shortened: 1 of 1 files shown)


MarkAckert (approved):

skurnevich:
After the discussion, we'll probably fix that through the schema

1000TurquoisePogs (approved):
This is an acceptable compromise to me.

What this would do in effect ensure dname and validity sections whenever keyrings are used.
This is more broad than necessary, but will prevent other bugs for now until a more specific solution is made.
The consequence is you'll end up with dname and validity environment variables set in scenarios where they're actually unused.
I do not expect any components/extensions would be reading "zowe.setup.certificate.keyring.dname" for any reason, so I think this is a fine solution for now.

PR #3929, opened 2024-08-07
Use correct parameters for node

Martin-Zeithaml:
1. `validateZosmfHostAndPort` does not work due to extra quotes for address and header:

```
Error: unknown parameter "https://skynet:800/zosmf/info"
Header name must be a valid HTTP token [""X-CSRF-ZOSMF-HEADER"]
```
2. `let` -> `const`

## Questions
* Why the code is using env variables `ZOSMF_HOST` and `ZOSMF_PORT`?
* What's the purpose of [this](https://github.com/zowe/zowe-install-packaging/blob/1f9ad776820f62672d7b3e17d64ef8b9b0569332/bin/commands/internal/start/prepare/index.ts#L175-L181)?

### DIFF
--- a/bin/libs/zosmf.ts
+++ b/bin/libs/zosmf.ts
@@ -31,7 +31,7 @@ export function validateZosmfHostAndPort(zosmfHost: string, zosmfPort: number): 
     common.printError(`Warning: Could not validate if z/OS MF is available on 'https://${zosmfHost}:${zosmfPort}/zosmf/info'. NODE_HOME is not defined.`);
     zosmfCheckPassed=false;
   } else {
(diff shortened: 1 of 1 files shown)


MarkAckert (approved):

MarkAckert:
This change looks good to me. For your second question:

* I agree the logic looks redundant and could be removed, or moved elsewhere. If port and host are set, then check a field which verifies that host and port are set.

MarkAckert:
Is this OK to merge or did you want to address the questions in your pr comment first?

Martin-Zeithaml:
Let's merge it (the bug and fix is obvious) and I will open an [issue](https://github.com/zowe/zowe-install-packaging/issues/3932) for the rest.

PR #3956, opened 2024-08-28
Clean up the code when checking z/OSMF [v3]

### LINKED ISSUE #3932: Review the zOSMF validation logic in `zwe internal start prepare`
The code is:
https://github.com/zowe/zowe-install-packaging/blob/0c2c64e677e168dcfb65adb9ab6b6f59038633e8/bin/commands/internal/start/prepare/index.ts#L168-L182

* Depending on environmental variables - is this from v1?
https://github.com/zowe/zowe-install-packaging/blob/0c2c64e677e168dcfb65adb9ab6b6f59038633e8/bin/commands/internal/start/prepare/index.ts#L37-L38
* The second check L175 (if processed) always returning `true`
https://github.com/zowe/zowe-install-packaging/blob/1f9ad776820f62672d7b3e17d64ef8b9b0569332/bin/libs/zosmf.ts#L52-L60
* This results in printing `z/OSMF is not configured. Using z/OSMF as authentication provider is not supported.`.

Martin-Zeithaml:

#### PR type
- [x] Feature 

#### Relevant issues

Fixes #3932

When checking z/OSMF connection, some messages might be confusing. See testing. 

- [x] No

#### Testing
Always this setting:
```
zowe:
  environments:
    ZOSMF_HOST: example.com
    ZOSMF_PORT: 1234
```

With or without fix, if `discovery.enabled=true`: 
`Successfully checked z/OS MF is available on 'https://example.com:1234/zosmf/info'`

But when this is set:
```
components.discovery.enabled=false
components.zaas.apiml.security.auth.provider=zosmf
```

No fix, confusing error message is printed and the zowe continues:
`ERROR: z/OSMF is not configured. Using z/OSMF as authentication provider is not supported.  `

With fix, error with reason is printed and zowe is terminated:
```log
ERROR: Using z/OSMF as 'components.zaas.apiml.security.auth.provider' is not possible: discovery is disabled.
ERROR: 2024-08-28 09:38:37 <ZWELS:33950026> martin ERROR (zwe-internal-start-prepare,global_validate) Zosmf validation failed
```


MarkAckert (approved):

MarkAckert:
Should we port this to v2 as well?

Martin-Zeithaml:
> Should we port this to v2 as well?

I called this an enhancement so we don't have to do it for v2 😄 
But it should be ported to v2 as well.

MarkAckert (approved):

PR #3966, opened 2024-08-30
Fix base-jdk and base-node builds

MarkAckert:
Base-jdk and base-node container builds should pull from `<majorVersion>-<OS>` rather than `latest-<OS>`. The former still allows us flexibility in building against the latest base without opening us to an arbitrary `latest` that could be created by past or future major versions of `base`. 

A matching PR will be opened for v2.


t1m0thyj (approved):
LGTM, thanks @MarkAckert!

PR #3967, opened 2024-08-30
Update base-jdk and base-node builds

MarkAckert:
Base-jdk and base-node container builds should pull from `<majorVersion>-<OS>` rather than `latest-<OS>`. The former still allows us flexibility in building against the latest base without opening us to an arbitrary `latest` that could be created by past or future major versions of `base`. 

Matches #3966


t1m0thyj (approved):
LGTM, thanks @MarkAckert!

PR #3969, opened 2024-09-02
Deprecate VSAM storage option in v3

achmelo:

#### PR type
- [ ] Bugfix 
- [ ] Feature 

#### Relevant issues

Fixes 

- 

- [ ] Yes
- [ ] No

---

---

---

---
_multi-line description_

Doc issue/PR number:

### DIFF
--- a/workflows/files/ZWECONF.properties
+++ b/workflows/files/ZWECONF.properties
@@ -539,12 +539,20 @@ components_caching_service_storage_vsam_name=
 #  Port for jgroups. This is required if storage mode is infinispan.
 components_caching_service_storage_infinispan_jgroups_port=7600
 
+# components_caching_service_storage_infinispan_jgroups_host
+# Label: Caching service storage jgroups host
+# Abstract: Host for jgroups
+# Category: components
+# Description:
+#  Host for jgroups
+components_caching_service_storage_infinispan_jgroups_host=
+
 # components_caching_service_storage_infinispan_jgroups_keyExchange_port
 # Label: Caching service storage jgroups key exchange port
 # Abstract: Port for jgroups key exchang
 # Category: components
 # Description:
-#  Port for jgroups key exchang. This is required if storage mode is infinispan.
+#  Port for jgroups key exchange. This is required if storage mode is infinispan.
 components_caching_service_storage_infinispan_jgroups_keyExchange_port=7601
 
 # components_app_server_enabled
--- a/workflows/files/ZWECONF.properties
+++ b/workflows/files/ZWECONF.properties
@@ -539,6 +539,14 @@ components_caching_service_storage_vsam_name=
 #  Port for jgroups. This is required if storage mode is infinispan.
 components_caching_service_storage_infinispan_jgroups_port=7600
 
+# components_caching_service_storage_infinispan_jgroups_keyExchange_port
+# Label: Caching service storage jgroups key exchange port
+# Abstract: Port for jgroups key exchang
+# Category: components
+# Description:
+#  Port for jgroups key exchang. This is required if storage mode is infinispan.
+components_caching_service_storage_infinispan_jgroups_keyExchange_port=7601
+
 # components_app_server_enabled
 # Label: Enable app server
 # Abstract: Check this option to enable the app server
--- a/workflows/files/ZWECONF.xml
+++ b/workflows/files/ZWECONF.xml
@@ -702,7 +702,7 @@ How we want to verify SSL certificates of services. Valid values are:
             <choice>redis</choice>
             <choice>infinispan</choice>
             <choice>VSAM</choice>
-            <default>VSAM</default>
+            <default>infinispan</default>
         </string>
     </variable>
     <variable name="components_caching_service_storage_size" scope="instance" visibility="public">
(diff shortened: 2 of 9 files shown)


workflows/files/ZWECONF.xml
```diff
@@ -2104,8 +2128,13 @@ echo '        # this is required if storage mode is infinispan' >> "${instance-z
 echo '        jgroups:' >> "${instance-zowe_runtimeDirectory}/zowe.yaml"
 #if (${instance-components_caching_service_storage_mode} == "infinispan" )
 echo '          port: ${instance-components_caching_service_storage_infinispan_jgroups_port}' >> "${instance-zowe_runtimeDirectory}/zowe.yaml"
+echo '          host: ${instance-components_caching_service_storage_infinispan_jgroups_host}' >> "${instance-zowe_runtimeDirectory}/zowe.yaml"
+echo '          keyExchange: ' >> "${instance-zowe_runtimeDirectory}/zowe.yaml"
+echo '            port: ${instance-components_caching_service_storage_infinispan_jgroups_keyExchange_port}' >> "${instance-zowe_runtimeDirectory}/zowe.yaml"
 #else
 echo '          port: 7600' >> "${instance-zowe_runtimeDirectory}/zowe.yaml"
+echo '          keyExchange: ' >> "${instance-zowe_runtimeDirectory}/zowe.yaml"
```
muzikovam, line 2133: There should be also the host in the else branch

workflows/files/ZWECONF.properties
```diff
@@ -539,6 +539,22 @@ components_caching_service_storage_vsam_name=
 #  Port for jgroups. This is required if storage mode is infinispan.
 components_caching_service_storage_infinispan_jgroups_port=7600
 
+# components_caching_service_storage_infinispan_jgroups_host
+# Label: Caching service storage jgroups host
+# Abstract: Host for jgroups
+# Category: components
+# Description:
+#  Host for jgroups
+components_caching_service_storage_infinispan_jgroups_host=
```
muzikovam, line 548: Please add the default here

muzikovam (changes requested):
Just small changes

Martin-Zeithaml:
And what about `zwe init vsam` command + VSAM samples + schema?

achmelo:
> And what about `zwe init vsam` command + VSAM samples + schema?

It will remain available in v3 so we don't need to change anything in samples and schema. It may be a good idea to print some message in the `zwe init vsam` but I am not sure where would be the best place. Do you have a suggestion?

workflows/files/ZWECONF.xml
achmelo: there is no default value for this parameter

workflows/files/ZWECONF.properties
achmelo: there is no default value

Martin-Zeithaml:
> > And what about `zwe init vsam` command + VSAM samples + schema?
> 
> It will remain available in v3 so we don't need to change anything in samples and schema. It may be a good idea to print some message in the `zwe init vsam` but I am not sure where would be the best place. Do you have a suggestion?

Ok, make sense, the `zwe init vsam` is already skipped if not `VSAM`. The command could stay the same. Maybe we can state some suggestion in [help](https://github.com/zowe/zowe-install-packaging/blob/v3.x/staging/bin/commands/init/vsam/.help): `VSAM` is deprecated, consider using `infinispan` because...

workflows/files/ZWECONF.properties
muzikovam: What's this then?

workflows/files/ZWECONF.xml
muzikovam: It can't be empty? My point is that if the user wouldn't specify it when they are running the workflow, but then they would like to, they can forget that this variable should be set. What do you think?

workflows/files/ZWECONF.properties
muzikovam: Or delete the default here

achmelo: excellent suggestion! thank you very much for your help

muzikovam (approved):

MarkAckert:
There's an error during zowe startup in the integration test:

```
 ZWESVUSR WARN ZWEL0066W command 'ZWE_zowe_runtimeDirectory="/redacted" ZWE_CLI_PARAMETER_HA_INSTANCE="redacted_ha" ZWED_SSH_PORT="00" ZWED_TN3270_PORT="00" _CEE_RUNOPTS="HEAPPOOLS(OFF),HEAPPOOLS64(OFF)" _EDC_UMASK_DFLT="0002" CONFIG="/redacted.yaml" _BPXK_AUTOCVT="ON" ROOT_DIR="/redact" ZWE_CLI_PARAMETER_CONFIG="/redacted.yaml" /redacted/bin/utils/configmgr -script /redacted/bin/commands/internal/start/prepare/cli.js 2>&1' ended with code 256

<ZWELNCH:16842890> ZWESVUSR ERROR ZWEL0030E failed to prepare Zowe instance
```

PR #3982, opened 2024-09-09
User/muzikovam/update zweconf

muzikovam:

#### PR type
- [ ] Bugfix 
- [ ] Feature 

#### Relevant issues

Fixes 

- Updated ZWECONF workflow

- [ ] Yes
- [x] No

---
* Affected function: _workflows_ *

---
* Description: _ Updated ZWECONF workflow_ *

---

---
_multi-line description_

Doc issue/PR number:


JoeNemo:
@1000TurquoisePogs, do you need to review any of this??

achmelo (approved):

workflows/files/ZWECONF.xml
```diff
@@ -2057,22 +2024,20 @@ echo '  zaas:' >> "${instance-zowe_runtimeDirectory}/zowe.yaml"
...
 echo '          zosmf:' >> "${instance-zowe_runtimeDirectory}/zowe.yaml"
-echo '            jwtAutoconfiguration: "auto"' >> "${instance-zowe_runtimeDirectory}/zowe.yaml"
-echo '            serviceId: "zosmf"' >> "${instance-zowe_runtimeDirectory}/zowe.yaml"
+echo '            jwtAutoconfiguration: auto' >> "${instance-zowe_runtimeDirectory}/zowe.yaml"
+echo '            serviceId: ibmzosmf' >> "${instance-zowe_runtimeDirectory}/zowe.yaml"
 echo '        authorization:' >> "${instance-zowe_runtimeDirectory}/zowe.yaml"
 echo '          endpoint:' >> "${instance-zowe_runtimeDirectory}/zowe.yaml"
 echo '            enabled: false' >> "${instance-zowe_runtimeDirectory}/zowe.yaml"
-echo '          provider: ""' >> "${instance-zowe_runtimeDirectory}/zowe.yaml"
+echo '          provider: "native"' >> "${instance-zowe_runtimeDirectory}/zowe.yaml"
```
pavelzlatnik, line 2037: when on 2030 line is provider without "", are you sure there on line 2037 should be provider with ""?

pavelzlatnik (changes requested):
see my comment to provider

workflows/files/ZWECONF.xml
pavelzlatnik: when there is value, "" should be removed (and later should be "" removed also from sample.yaml I guess, where is provider without value).

muzikovam: Where is it without the quotation marks? What do you mean by "on line"?

muzikovam: That's a different provider. I know what you mean but we are simply trying to have the same file as is zowe-example.yaml. But to be fair it does not really matter. Zwe program would handle it with or without the quotation marks and you can't base on the only fact that the end variable is called the same.

pavelzlatnik: image
when there is set value (native), there should not be used quotation marks. And technically it is the same kind of variable

muzikovam: Well, each string here is the same kind of variable🙃 But please, have this conversation with someome who is taking care of zowe-example.yaml and not with me who only cares about the output being the same😊

pavelzlatnik (approved):
ok

PR #3989, opened 2024-09-11
Fix PSWI Build Failures in Nightly

MarkAckert:
The nightly build is failing due to two new changes:

- We always build PSWI on nightly now
- PSWI added tests for workflows. These passed during PR merge but are failing on staging.

This PR removes the failing PSWI test. We will continue building PSWI nightly.


zFernand0 (approved):
it helps to review with hidden whitespaces.
Also, the ``` `cat JCL` -> $(cat JCL)```  makes sense.

Thanks for commenting out the failing tests 🙏🏽

PR #4043, opened 2024-10-08
User/zlatnikp/apimlwf

pavelzlatnik:

- [ x] DCO signoffs have been added to all commits, including this PR

#### PR type

- [ ] Bugfix 
- [ ] Feature 
- [x ] Other... Please describe:
New workflow, specifically designed for APIML. It is backward compatible with full zowe (full yaml generated).
But it is simplified for APIML configuration. Some non-necessary setups removed from workflow (like HA, etc..)
Not all I describe here, but major changes visible by end-user->

There is new option certificate setup, default set to false (unchecked). When unchecked, workflow works almost the same as original zowe workflow.
image
under APIML components, all are set on by default, except caching service which is false (and has new abstract)
image

certificate variables are moved to extra step (and are skipped in case certificate setup was set to true)
image

If certificate setup was set to true, workflow would let generate zowe certificate, certificate variables are skipped, and 
is enabled new step (which would be normally skipped), which handle generation of zowe certificate
image

So quick setup of APIML (with or without zowe generated certificate) should be way easier and faster for user who wants to test APIML on some system.

#### Relevant issues

Fixes 

- 

- [ ] Yes
- [ x] No

yes, there will have to be created update of documentation to explain use-case for this workflow.

---

---

---

---
_multi-line description_

not yet

Doc issue/PR number:


workflows/files/ZWEAMLCF.xml
```diff
@@ -0,0 +1,2151 @@
...
+    <!--variables-->
+    <variable name="zowe_setup_dataset_prefix" scope="instance" visibility="public">
+        <label>Zowe setup MVS prefix</label>
+        <abstract>Where Zowe MVS data sets will be installed</abstract>
+        <description>Where Zowe MVS data sets will be installed</description>
+        <category>zowe</category>
+        <string valueMustBeChoice="false" multiLine="false">
+            <!-- Put validation here -->
+            <!-- Specify choices here -->
+            <default>IBMUSER.ZWEV2</default>
```
Martin-Zeithaml, line 22: All occurrences:
IBMUSER.ZWEV2 -> IBMUSER.ZWEV**3**

workflows/files/ZWEAMLCF.xml
```diff
@@ -0,0 +1,2151 @@
...
+            <choice>JCECCAKS</choice>
+            <choice>JCERACFKS</choice>
+            <choice>JCECCARACFKS</choice>
+            <choice>JCEHYBRIDRACFKS</choice>
+            <default>PKCS12</default>
+        </string>
+    </variable>
+    <variable name="zowe_certificate_truststore_file" scope="instance" visibility="public">
+        <label>Zowe certificate truststore file</label>
+        <abstract>File location for cert truststore. For keyring, the format is "safkeyring://OWNER/KEYRING"</abstract>
```
Martin-Zeithaml, line 330: File location for cert truststore. -> File location for certificate truststore.

workflows/files/ZWEAMLCF.xml
pavelzlatnik: Good catch for both. Those are already coming from V3, so I will fix that in Zowe V3. APIML workflow I have to rework based on Tuesday demo to APIML team, new use case updated, so I will close this pullRequest and open new one when I am ready. It will be slightly different (because of bit different use).

pavelzlatnik:
I have to rework this new workflow based on Tuesday demo to APIML team, new use case updated.

PR #4048, opened 2024-10-10
Users/zlatnikp/f144933 auth zss bypass

pavelzlatnik:

- [X ] DCO signoffs have been added to all commits, including this PR

#### PR type
- [X ] Bugfix 
- [ ] Feature 
- [X ] Other... Please describe:
- CHANGES ONLY FOR V2 workflow (not for V3!)
Based on request: For our work in the feature F144933, and to keep the installation API ML only, we need to have the z/OSMF Workflow Zowe Configuration of Zowe 2.0, handle the following property, which enables X.509 and OIDC authentication to work without the need of ZSS:
components.gateway.apiml.security.useInternalMapper

+small bug fix I have found during testing of v2 workflow

#### Relevant issues

Fixes 

CHANGES ONLY FOR V2 workflow (not for V3, do not retrofit to V3 workflow!)
Only configuration workflow for V2 is updated based on request from APIML team.
- 

- [ ] Yes
- [ X] No

---

---
* V2 configuration workflow update

---

---
_multi-line description_

Doc issue/PR number:

### DIFF
--- a/workflows/files/ZWECONF.xml
+++ b/workflows/files/ZWECONF.xml
@@ -469,6 +469,17 @@ How we want to verify SSL certificates of services. Valid values are:
             <default>7554</default>
         </integer>
     </variable>
+    <variable name="components_gateway_internalMapper" scope="instance" visibility="public">
+      <label>Enable Gateway internal mapper</label>
+      <abstract>Check this option to enable the Gateway internal ESM identity mapper</abstract>
+      <description>Check this option to enable Gateway internal ESM identity mapper</description>
+      <category>components</category>
+      <boolean>
+          <!-- Put validation here -->
+          <!-- Specify choices here -->
+          <default>false</default>
+      </boolean>
+  </variable>
     <variable name="components_gateway_debug" scope="instance" visibility="public">
         <label>Gateway debug</label>
         <abstract>Switch on the debug mode for the gateway</abstract>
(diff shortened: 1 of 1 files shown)


workflows/files/ZWECONF.xml
```diff
@@ -469,6 +469,17 @@ How we want to verify SSL certificates of services. Valid values are:
             <default>7554</default>
         </integer>
     </variable>
+    <variable name="components_gateway_internalMapper" scope="instance" visibility="public">
```
Martin-Zeithaml, line 472: This is not wrong, however I was confused by the name `components_gateway_internalMapper` expecting to find matching property in gateway schema. But there is `component.gateway.apiml.security.useInternalMapper`. The setting then is done at the right place, but consider using corresponding names.

pavelzlatnik: Technically it is ID, I kept "same shorten approach" as other variables from same section, like components_gateway_port, components_gateway_debug, but I could update it to "long name". Only issue with long names is, that it is hard to read direct xml, or do quick update, when variables are used in conditions, especially on greenscreen. This one is just used and injected within shell script, so I could sync it with components_gateway_apiml_security_auth_zosmf_jwtAutoconfiguration. But in case when variable is used in some condition, especially when condition is more complex or contain multiple variables, code is almost unreadable with long names, and in inline-jcls have to be shortened anyway. I personally prefer usage of short names in IDs with 3 words per ID.

Martin-Zeithaml (approved):
Approved.

Issue #4044, opened 2024-10-09
`zwe install` implementation details


Martin-Zeithaml:
As moving to JCL templates, `zwe install` should be rewritten too. All we need for installation is `zowe.runtimeDirectory` and `zowe.setup.dataset.prefix`.

Current implementation supports 2 ways:
* `zwe install --config /path/to/zowe.yaml`
* `zwe install --ds-prefix ZOWE.V3R0M0`
  * No need for `zowe.yaml`, this is handy for quick installation, developers...

There is a problem with `import`, if we use the same approach as for other commands:
```
bin/command/install/index.sh
+-->  "configmgr cli.js"

bin/command/install/cli.js
+--> "index.execute()"

bin/command/install/index.js
+--> "install implementation"
```

At some place we need to use `import * as config` or `import * as configmgr`: this will cause the code will require config. But we need also code path for `--ds-prefix`, which means without config.

Solutions:
* Remove `--ds-prefix` - I would like to keep it, I am using it 😄 
* Dynamic import - do we support this is quickJS?
* Use 2 different "CLIs" as in #3635 

### Update
Problems:
* With `--ds-prefix` user does not have an option to customize JCL header - unless we will add another parameter
* JCL `ZWEINSTL` creates datasets and copies files (SHell in BPXBATCH)
  * At this moment, it will always fail for the second submit (ALLOCATE will fail)
  * When `--allow-overwrite` is used, we need to skip allocating or delete datasets or have 2 JCL, one with allocate+shell and the second with shell only or something better

1000TurquoisePogs:
I think the 2 CLI.js approach is valid because we dont want to dilute the principles of configmgr schema validation, but we also don't want to lose features by forcing things to use a YAML when not required. So, yeah, the command variant that uses the YAML should do configmgr validation, and the one that doesnt could be generic quickJS code.

MarkAckert:
I would like to take a closer look, but at a glance, I prefer option (1) or (2). The split CLI.js approach is valid and makes sense, but I'm a little worried about maintenance over time if we're essentially duplicating the bulk of the implementation code in two places, one with configmgr validation and one without. DRY, code drift, and all that. For option (3), my question is: is the benefit to developers worth the additional code maintenance? Or is there another way to make this developer friendly - like `zwe install --ds-prefix my-prefix --dummy-config` (`-dc`)?

Martin-Zeithaml:
## Split
The split (so far) was done in the minimal way:
[...]

PR #4057, opened 2024-10-18
[TEST] zwe install - tmp config

Martin-Zeithaml:
POC for [`zwe install`](https://github.com/zowe/zowe-install-packaging/issues/4044) using JCL template:
* When `--ds-prefix` is used, temporary config is created

### Pros
* "Classis" code path, `index.sh` -> `cli.js` -> `index.js`

### Cons
* Another tmp file
* If only `--ds-prefix` is used and it is wrong, schema validation error will be displayed, which is confusing


MarkAckert:
@Martin-Zeithaml can we close this?

Martin-Zeithaml:
#4057 and #4056 are POC for `zwe install` using JCL:
* I was very optimistic related to finishing Migrate to JCL and started with `zwe install`
* There is technical challenge - current `zwe install` works without config - that's why there are two POC

**But we can close both.**

PR #4069, opened 2024-11-05
Read yaml by configmgr

Martin-Zeithaml:
Remove nodeJS dependency by removing `read_yaml` - same functionality done with `configmgr`.

0) Quickly remove nodeJS dependency while wating for #3718
1) New `shell` function added
2) Minimal changes to `install` command - this could be used by `init` too

## Test
```
/zowe/bin: ./zwe install -c ./zowe.yaml -vv
===============================================================================
>> INSTALL ZOWE MVS DATA SETS

- read_yaml_configmgr process ./zowe.yaml and extract '.zowe.setup.dataset.prefix -> /zowe/setup/dataset/prefix'
  * Exit code: 0
  * Output:
    ZOWE.TEST
Create MVS data sets if they do not exist
Creating Zowe sample library - ZOWE.TEST.SZWESAMP
- tsocmd ALLOCATE NEW DA('ZOWE.TEST.SZWESAMP') dsntype(library) dsorg(po) recfm(f b) lrecl(80) unit(sysallda) space(15,15) tracks
  * Succeeded
  * Exit code: 0
  * Output:
    ALLOCATE NEW DA('ZOWE.TEST.SZWESAMP') dsntype(library) dsorg(po) recfm(f b) lrecl(80) unit(sysallda) space(15,15) tracks

...

Copy files/SZWESAMP/ZWEIAPF to ZOWE.TEST.SZWESAMP
- cp  -v ./ZWEIAPF //'ZOWE.TEST.SZWESAMP'
  * Succeeded
  * Exit code: 0
  * Output:
    ./ZWEIAPF -> ZOWE.TEST.SZWESAMP(ZWEIAPF): text

...

-------------------------------------------------------------------------------
>> Zowe MVS data sets are installed successfully.

Zowe installation completed. In order to use Zowe, you need to run "zwe init" command to initialize Zowe instance.
- Type "zwe init --help" to get more information.

You can also run individual init sub-commands: mvs, certificate, security, vsam, apfauth, and stc.
- Type "zwe init <sub-command> --help" (for example, "zwe init stc --help") to get more information.
```

## Notes
Old bugs kept:
* No check of e.g. `--ds-prefix 1234`
* `--allow-overwrite` not needed if you delete `SZWEEXEC`
* Special chars
* And probably much more...

PR #4088, opened 2024-11-14
Read yaml by configmgr - `zwe init`

Martin-Zeithaml:
Continuation of #4069.

Simple NodeJS dependency removal for `zwe init` command while waiting for #3718.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,6 @@ All notable changes to the Zowe Installer will be documented in this file.
 
 ## `3.0.1`
 - Enhancement: command `zwe init` does not require NodeJS [#4088](https://github.com/zowe/zowe-install-packaging/pull/4088)
-- Bugfix: When `--log-dir` parameter for `zwe` command is a file, there might be an error "InternalError: stack overflow". [#4071](https://github.com/zowe/zowe-install-packaging/pull/4071)
 - Enhancement: new javascript funtion `getStatvfs()` to obtain information about the file sysytem [#3994](https://github.com/zowe/zowe-install-packaging/pull/3994)
 - Enhancement: command `zwe diagnose` in javascript only [#4061](https://github.com/zowe/zowe-install-packaging/pull/4061)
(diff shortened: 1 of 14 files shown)


1000TurquoisePogs (changes requested):
this fails due to different behavior of configmgr:


```
- read_yaml_configmgr process /zowe/inst/no-nodeinit/zowe.yaml and extract '.zowe.setup.certificate.keyring.zOSMF.user -> /zowe/setup/certificate/keyring/zOSMF/user'
  * Exit code: 0
  * Output:
    error not found, reason=101
...
  * certificate pkcs12 import --keystore /zowe/inst/no-nodeinit/keystore/error not found, reason=101/error not found, reason=101.keystore.p12 --password error not found, reason=101 --alias error not found, reason=101 --source-keystore error not found, reason=101 --source-password error not found, reason=101 --source-alias error not found, reason=101
```

looks like errors need to be >/dev/null

Martin-Zeithaml:
> this fails due to different behavior of configmgr:
> 
> ```
> - read_yaml_configmgr process /zowe/inst/no-nodeinit/zowe.yaml and extract '.zowe.setup.certificate.keyring.zOSMF.user -> /zowe/setup/certificate/keyring/zOSMF/user'
>   * Exit code: 0
>   * Output:
>     error not found, reason=101
> ...
>   * certificate pkcs12 import --keystore /zowe/inst/no-nodeinit/keystore/error not found, reason=101/error not found, reason=101.keystore.p12 --password error not found, reason=101 --alias error not found, reason=101 --source-keystore error not found, reason=101 --source-password error not found, reason=101 --source-alias error not found, reason=101
> ```
> 
> looks like errors need to be >/dev/null

It seems to be more complicated, the error in printed to [stdout](https://github.com/zowe/zowe-common-c/blob/3d7c04c6f4c6acf4cd440ff775fabf8a6add31af/c/configmgr.c#L1834) and return code is always 0 (found/not found). I am afraid, the only shell solution for now is to check if the return value is `error not found, reason=` and hoping it is not real value.

The right solution would be update configmgr to return > 0 for "not found" case.

MarkAckert (approved):

Issue #3828, opened 2024-05-13
bin/libs/common.sh: print_error_and_exit can not use error codes > 255


Martin-Zeithaml:
The function `print_error_and_exit` is ending with [exit](https://github.com/zowe/zowe-install-packaging/blob/2b08d669ff8ecd7162013ae3654b2fafcd128196/bin/libs/common.sh#L126) command, which accepts any number, however only the [least significant 8bits are used](https://www.ibm.com/docs/en/zos/3.1.0?topic=descriptions-exit-return-shells-parent-process-tsoe).

For example: in case of missing `useConfigmgr=true`, [this](https://github.com/zowe/zowe-install-packaging/blob/2b08d669ff8ecd7162013ae3654b2fafcd128196/bin/commands/components/search/index.sh#L22) returns 60:
```
/zowe/bin: ./zwe components search --component-id org.zowe.zlux.sample.angular
Error ZWEL0316E: Command requires zowe.useConfigmgr=true to use.
/zowe/bin: echo $?
60
```

JoeNemo:
Sean suggests returning 8 for all failures and putting the particular status in a environment variable.

Martin-Zeithaml:
And the same for JS:
```
echo "import * as std from 'cm_std'; std.exit(316);" > test_exit.js
/zowe/bin/utils/configmgr -script test_exit.js
echo $?
60
```

JoeNemo:
This is an external breaking changes and should be done within the 3.0 releases.

PR #4107, opened 2024-12-11
Zwe Init Refactor to JCL for v3.x/staging

MarkAckert:
Continuation of work started in #3718


Martin-Zeithaml:
Improvements needed:
## Messages
The message `ZWEL0319E` seems to be overloaded:
```
ZWEL0319E||NodeJS required but not found. Errors such as ZWEL0157E may occur as a result. The value 'node.home' in the Zowe YAML is not correct. 
ZWEL0319E|319|zowe.setup.dataset.jcllib does not exist, cannot run. Run 'zwe init', 'zwe init generate', or submit JCL ${prefix}.SZWESAMP(ZWEGENER) before running this command.
ZWEL0319E|319|Configuration stored in PARMLIB must use member name ZWEYAML when using generate action.
```
## Update of parmlib
`zwe init` updates a members and then it fails:
```
zwe init -c 'ZOWE.PARM(HELLO)' --update-config
Writing temp file for PARMLIB update. Command= cp -v "/tmp/.zweenv-7470/zwe-parmlib-3194" "//'ZOWE.PARM(HELLO)'

ERROR: ZWEL0319E Configuration stored in PARMLIB must use member name ZWEYAML when using generate action.
```
It should fail before any update.

## Parmlib
`zwe init --config 'PARMLIB(ZOWE.PR#4107.PARMLIB(A))'` will cause a problem [here](https://github.com/zowe/zowe-install-packaging/blob/73e64037734a89e5e77a2816bb4f3cf310a07c8e/bin/commands/init/generate/index.ts#L56) as the config path is split based on words `FILE|PARMLIB`. I am afraid, the word `PARMLIB` is used for, ehm, parmlib very often. This must be rewritten.

MarkAckert:
@Martin-Zeithaml do you think this is ready to merge into #3718? After merge, we can change the head on that PR to v3.x/staging

Martin-Zeithaml:
> @Martin-Zeithaml do you think this is ready to merge into #3718? After merge, we can change the head on that PR to v3.x/staging

I thought this will be merged (some day) into `v3.x/staging` and the #3718 will be just deleted.
I am testing this and making changes here only. There are some new [problems](https://github.com/zowe/zowe-install-packaging/pull/4107#issuecomment-2618458066).

MarkAckert:
We should close out #3718 then to avoid confusion with this PR. I can update the title and description. cc @1000TurquoisePogs does that sound OK to you?

MarkAckert:
> > @Martin-Zeithaml do you think this is ready to merge into #3718? After merge, we can change the head on that PR to v3.x/staging
> 
> I thought this will be merged (some day) into `v3.x/staging` and the #3718 will be just deleted. I am testing this and making changes here only. There are some new [problems](https://github.com/zowe/zowe-install-packaging/pull/4107#issuecomment-2618458066).

I took a stab at problems 2 and 3, without a complete test suite pointed at it. I added a new message `ZWEL0320E` which may not help with overloaded or sprawling messages...

Martin-Zeithaml:
## Apfauth
Help:
- `zowe.setup.dataset.authLoadlib` is the user custom APF LOADLIB. This field is
  **optional**. If it's not defined, `SZWEAUTH` from `zowe.setup.dataset.prefix` data
  set will be APF authorized.
- `zowe.setup.dataset.authPluginLib` is the user custom APF PLUGINLIB.
  You can install Zowe ZIS plugins into this load library.

`zowe.setup.dataset.authLoadlib` and `zowe.setup.dataset.authPluginLib` omitted in config and this is generated by `ZWEGEN00`:

```
//STDPARM DD *                                          
SH cd '/zowe/' &&                       
cd bin/utils &&                                         
export LOADLIB='CFG.ZOWE.SETUP.DATASET.AUTHLOADLIB' &&  
export LOADLOC=SMS &&                                   
export PLUGLIB='CFG.ZOWE.SETUP.DATASET.AUTHPLUGINLIB' &&
export PLUGLOC=SMS &&                                   
./opercmd.rex "SETPROG APF,ADD,DSN=$LOADLIB,$LOADLOC" &&
./opercmd.rex "SETPROG APF,ADD,DSN=$PLUGLIB,$PLUGLOC"   
//*                                                     
```

When trying `zwe init apfauth -c ./zowe.yaml`:

`ERROR: Error ZWEL0157E: zowe.setup.dataset.authLoadlib is not defined in Zowe YAML configuration file.`

Expectation:
If `zowe.setup.dataset.authLoadlib` and `zowe.setup.dataset.authPluginLib` omitted, the JCL will authorize only `zowe.setup.dataset.prefix`.`SZWEAUTH`.

MarkAckert:
I started on the Apfauth updates, testing in the other PR. More changes to come - the first commit isn't working yet.

bin/commands/init/generate/index.ts
```diff
@@ -0,0 +1,145 @@
...
+  if (!runtimeDirectory) {
+    common.printErrorAndExit(`Error ZWEL0157E: Zowe runtime directory (zowe.runtimeDirectory) is not defined in Zowe YAML configuration file.`, undefined, 157);
+  }
+  
+  let jclPostProcessing = false;
+  let jclHeaderJoined = '';
+  const jclHeader = ZOWE_CONFIG.zowe?.environments?.jclHeader == null ? '' : ZOWE_CONFIG.zowe.environments.jclHeader;
+  if (Array.isArray(jclHeader)) {
+    jclPostProcessing = true;
+    jclHeaderJoined = jclHeader.join("\n// ");
```
Martin-Zeithaml: This is braking the intended behavior, I wanted to avoid writing JCL parser, it is up to user to define whatever is needed to follow `JOB` keyword.

Help for `init generate`:
JCL JOB statement customization:
- Each JOB statement is defined as `//ZWEnnnnn JOB {zowe.environments.jclHeader}`, where `nnnnn` identifies each job
- Default setting is empty string causing no additional JOB fields will be used - if your site or External Security Manager does not require additional fields, you can skip this
- Otherwise you can use `zwe init generate` to set additional JOB fields
- Note: JCL syntax is not checked by `zwe init generate` command

Accounting information only:
```yaml
zowe:
  environments:
    jclHeader: 123456
```
Additional fields as one string:
```yaml
zowe:
  environments:
    jclHeader: "123456,\n//       'Zowe User',\n//       NOTIFY=&SYSUID"
```
Additional fields as an array:
```yaml
zowe:
  environments:
    jclHeader:
    - "123456,"
    - "//       'Zowe User',"
    - "//       NOTIFY=&SYSUID"
```

schemas/zowe-yaml-schema.json
```diff
@@ -58,8 +58,16 @@
                   "default": "<hlq>.SZWEAUTH"
                 },
                 "authPluginLib": {
-                  "$ref": "/schemas/v2/server-common#zoweDataset",
-                  "description": "APF authorized LOADLIB for Zowe ZIS Plugins"
+                  "oneOf": [{
```
Martin-Zeithaml: I am not sure about this, `authPluginLib` is not defined in defaults and it is not required by schema, user has already the options:
* Either specify dataset - which must be valid from schema (regex) point of view
* Omit this completely as intended not to use it 

Allowing empty dataset name is strange solution.

MarkAckert: Yeah, I don't love this either and I'm going to revert it. An empty value could also imply behaviors other than an effectively unset variable.

I was working from the perspective of a user who starts with an example-zowe.yaml, and as they go through the yaml file field-by-field they blank the authPluginLib since they don't want or need it. I think this user scenario could happen, and with the reverted behavior, the schema validator will tell them they need a dataset value, which they then fill out, which enables authPluginLib when they don't want it. All this assumes they've gone off-doc at least a little bit, which also seems plausible to me :). I would like to revisit the schema for cases like this another time...

files/defaults.yaml
```diff
@@ -27,9 +27,12 @@ zowe:
   #-------------------------------------------------------------------------------
   # These configurations are used by "zwe install" or "zwe init" commands.
   #-------------------------------------------------------------------------------
+  environments:
+    jclHeader: ''
   setup:
     # MVS data set related configurations
-    dataset:      
+    dataset: 
+      authLoadlib: ${{ zowe.setup.dataset.prefix }}.SZWEAUTH
```
Martin-Zeithaml: One of my old test is failing. There is no `zowe.setup.dataset.prefix` defined in the main config, so this line will generate `undefined.SZWEAUTH`. Then `zwe` command fails on schema validation:
```
    Validity Exceptions(s) with object at /zowe/setup
      Validity Exceptions(s) with object at /zowe/setup/dataset
        Validity Exceptions(s) with string at /zowe/setup/dataset/authLoadlib
          string pattern match fail s='undefined.SZWEAUTH', pat='^([A-Z\$\#\@]){1}([A-Z0-9\$\#\@\-]){0,7}(\.([A-Z\$\#\@]){1}([A-Z0-9\$\#\@\-]){0,7}){0,11}$', at /zowe/setup/dataset/authLoadlib
```

Martin-Zeithaml: When I tried this, it works, however `zowe.setup.dataset.prefix` was not in the merged config.
```
    dataset: "${{ function dsProps() {
                    let defaultDSProps = { 'parmlibMembers': { 'zis': 'ZWESIP00' } };
                    if (zowe.setup.dataset.prefix) {
                      defaultDSProps['authLoadlib'] = zowe.setup.dataset.prefix + '.SZWEAUTH';
                    }
                    return defaultDSProps;
                  };
                  dsProps(); }}"
```
`zowe.setup.dataset.prefix` was in config, but missing in the merged output:
```
"setup": {
    "dataset": {
      "parmlibMembers": {
        "zis": "ZWESIP00"
      },
      "authLoadlib": "ZOWE.SZWEAUTH"
    },
```

It could be done with the `dsProps` function, but this function must check all possible `dataset` properties. And after all, it will be conditional default.

It seems the best solution is to delete this default and do the checking in the code only.

MarkAckert: Is the correct outcome: if `zowe.setup.dataset.prefix` is undefined, get a schema validation error like: `/zowe/setup/dataset/prefix must be defined`? 

Or do we want to tolerate an unset prefix so users can enter target datasets as-is?

Martin-Zeithaml: This would be a solution, but making `zowe.setup.dataset.prefix` mandatory:
* it is a braking change
* any other existing `zwe` command will require it even it is not needed for the processing of that command

For example, I am usually deleting `zowe.setup.*` in config to keep yaml file small.

bin/commands/init/mvs/index.ts
```diff
@@ -0,0 +1,118 @@
...
+
+  // check if user passed --generate
+  const forceGen = !!std.getenv('ZWE_CLI_PARAMETER_GENERATE')
+  if (forceGen) {
+    initGenerate.execute();
+  }
+
+  const jcllib = zoslib.verifyGeneratedJcl(ZOWE_CONFIG);
+  if (!jcllib) {
+    common.printErrorAndExit(`Error ZWEL0319E: zowe.setup.dataset.jcllib does not exist, cannot run. Run 'zwe init', 'zwe init generate', or submit JCL ${prefix}.SZWESAMP(ZWEGENER) before running this command.`, undefined, 319);
```
ChongZhou-Broadcom, line 44: The maximum exit code of USS is 255, so 319 will be clamped to 63, which is probably not what you are expecting, although this is really not a big deal for now (...but think about your test framework) If you search for "printErrorAndExit", you might find more similar problems.

Martin-Zeithaml: This is on [todo](https://github.com/zowe/zowe-install-packaging/issues/3828) list.... for version 4, because there was no time to push it to v3.

bin/commands/init/security/index.ts
```diff
@@ -0,0 +1,81 @@
...
+  const jcllib = zoslib.verifyGeneratedJcl(ZOWE_CONFIG);
+  if (!jcllib) {
+    return common.printErrorAndExit(`Error ZWEL0319E: zowe.setup.dataset.jcllib does not exist, cannot run. Run 'zwe init', 'zwe init generate', or submit JCL ${prefix}.SZWESAMP(ZWEGENER) before running this command.`, undefined, 319);
+  }
+
+  let securityProduct = zos.getEsm();
+  if (!securityProduct || securityProduct == 'NONE') {
+    securityProduct = ZOWE_CONFIG.zowe.setup?.security?.product;
+    if (!securityProduct) {
+      common.printErrorAndExit(`Error ZWEL0157E: Zowe dataset prefix (zowe.setup.dataset.prefix) is not defined in Zowe YAML configuration file.`, undefined, 157);
```
ChongZhou-Broadcom: The error message is incorrect.

Martin-Zeithaml: Yes, I know. There is a [PR](https://github.com/zowe/zowe-install-packaging/pull/4244), which removes the error completely.

ChongZhou-Broadcom: I didn't know it. I spent a couple of hours on creating test cases for the 3 forEach checks.

bin/commands/init/vsam/index.ts
```diff
@@ -0,0 +1,76 @@
...
+  const vsamExistence = zosDataset.isVsamDatasetExists(name);
+  if (vsamExistence && allowOverwrite) {
+    zosJes.printAndHandleJcl(`//'${jcllib}(ZWECSRVS)'`, `ZWECSRVS`, jcllib, prefix, false, true);
+  } else if (vsamExistence) {
+    return common.printErrorAndExit(`Error ZWEL0158E: ${name} already exists.`, undefined, 158);
+  }
+
+  zosJes.printAndHandleJcl(`//'${jcllib}(ZWECSVSM)'`, `ZWECSVSM`, jcllib, prefix);
+  if (!dryRun && updateConfig) {
+    json.updateZoweYaml(std.getenv('ZWE_CLI_PARAMETER_CONFIG_ORIG'), '.components.caching-service.storage.vsam.name', name);
```
Martin-Zeithaml: `ZWE_CLI_PARAMETER_CONFIG_ORIG` is never initialized. This will print:
```
- update zowe config undefined, key: ".components.caching-service.storage.vsam.name" with value: ZOWE.PR#4107.$-CACHE-
  * Success
```

Martin-Zeithaml:
After #4285:
`zwe init generate --config "PARMLIB(ZOWE.PR#4107.TMP(DEFAULT)):PARMLIB(ZOWE.PR#4107.TMP(NODE))"`

```jcl
//ZWEGENER
...
//MYCONFIG DD   *,DLM=$$        
PARMLIB ZOWE.PR#4107.TMP
PARMLIB ZOWE.PR#4107.TMP
```

Failing for obvious reason, still expecting `ZWEYAML`:
`ZWEL0318E - failed to read '//'ZOWE.PR#4107.TMP(ZWEYAML)''`

I'll try to fix this.

**Edit:** [Fix](https://github.com/zowe/zowe-install-packaging/pull/4332) is needed for staging too, once it is merged it should be enough to fix `zwe init generate` to include the members, when building `MYCFG`, 1 line change here:
https://github.com/zowe/zowe-install-packaging/blob/0c3ed61adb5d3b1e2dd39eeb38ada79bd5fbac9a/bin/commands/init/generate/index.ts#L100

MarkAckert:
All release tests functionally passed against this PR. Tests which failed are also failing with the same errors on `v3.x/staging`, and as such are not due to changes in this PR.

struga0258 (approved):

PR #4113, opened 2024-12-23
z/OSMF static definition file processing

pablocarle:

#### PR type

- [ ] Feature 

Include logic to determine if connections from API ML components to z/OSMF are made using https (default) or HTTP (when outbound AT-TLS rules are in place)

- [ ] Yes
- [x] No

### DIFF
--- a/bin/libs/component.ts
+++ b/bin/libs/component.ts
@@ -387,11 +387,23 @@ export function findAllLaunchComponents2(): string[] {
   });
 }
 
+function getBooleanEnv(variableName) {
+  const value = std.getenv(`${variableName}`);
+  if (value === "true") {
+      return true
+  }
+  if (value === "false") {
+      return false
+  }
+  return undefined
+}
+
+
 function isClientAttls() {
-  const clientGlobalAttls = std.getenv('ZWE_zowe_network_client_tls_attls') == 'true' ? true : false;
-  const serverGlobalAttls = std.getenv('ZWE_zowe_network_server_tls_attls') == 'true' ? true : false;
-  const clientLocalAttls = std.getenv('ZWE_components_zaas_zowe_network_client_tls_attls') == 'true' ? true : false;
-  const serverLocalAttls = std.getenv('ZWE_components_zaas_zowe_network_server_tls_attls') == 'true' ? true : false;
+  const clientGlobalAttls = getBooleanEnv('ZWE_zowe_network_client_tls_attls');
+  const serverGlobalAttls = getBooleanEnv('ZWE_zowe_network_server_tls_attls');
+  const clientLocalAttls = getBooleanEnv('ZWE_components_zaas_zowe_network_client_tls_attls');
+  const serverLocalAttls = getBooleanEnv('ZWE_components_zaas_zowe_network_server_tls_attls');
   const clientAttls = clientGlobalAttls || clientLocalAttls;
   if ((clientGlobalAttls !== false) && (clientLocalAttls !== false) && (!clientAttls)) {
     // If client attls not explicitly false OR truthy, have client follow server attls variable. it simplifies common case in which users want both.
--- a/bin/libs/component.ts
+++ b/bin/libs/component.ts
@@ -388,11 +388,11 @@ export function findAllLaunchComponents2(): string[] {
 }
 
 function isClientAttls() {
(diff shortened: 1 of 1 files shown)


bin/libs/component.ts
```diff
@@ -418,6 +418,22 @@ export function processComponentApimlStaticDefinitions(componentDir: string): bo
 
         const contents = xplatform.loadFileUTF8(path,xplatform.AUTO_DETECT);
         if (contents) {
+          const zosmfScheme = std.getenv("ZOSMF_SCHEME");
+          const attls = std.getenv("ZWE_zowe_network_client_tls_attls");
```
1000TurquoisePogs: This will not work for configs where people have recommended enabling attls through only the server variable.

It would be better to say if the client variable isn't defined, to use the server variable's value.
Something like we do here:

https://github.com/zowe/zlux-server-framework/blob/v3.x/staging/lib/util.js#L510-L520

bin/libs/component.ts
```diff
@@ -387,6 +387,20 @@ export function findAllLaunchComponents2(): string[] {
   });
 }
 
+function isClientAttls() {
+  const clientGlobalAttls = std.getenv("ZWE_zowe_network_client_tls_attls");
+  const serverGlobalAttls = std.getenv("ZWE_zowe_network_server_tls_attls");
+  let clientLocalAttls = std.getenv("ZWE_components_zaas_zowe_network_client_tls_attls");
+  let serverLocalAttls = std.getenv("ZWE_components_zaas_zowe_network_server_tls_attls");
+  let clientAttls = clientGlobalAttls || clientLocalAttls;
+  if ((clientGlobalAttls !== false) && (clientLocalAttls !== false) && (!clientAttls)) {
```
Martin-Zeithaml, line 408: The [`std.getenv`](https://github.com/JoeNemo/quickjs-portable/blob/291b874c516d10ceb103afe0bece7301a1cebef6/quickjs-libc.c#L655) is returning a string. This `if` is comparing string vs Boolean. You need to get Booleans:
```
const clientGlobalAttls = std.getenv("ZWE_zowe_network_client_tls_attls") == "true" ? true : false;
...
```
or comparing strings
```
if ((clientGlobalAttls !== "false") && (clientLocalAttls !== "false") && (!clientAttls)) {
```

I prefer the first version.

1000TurquoisePogs (approved):
This now looks correct after the resolved conversation.

bin/libs/component.ts
```diff
@@ -387,6 +387,20 @@ export function findAllLaunchComponents2(): string[] {
...
 
+function isClientAttls() {
+  const clientGlobalAttls = std.getenv('ZWE_zowe_network_client_tls_attls') == 'true' ? true : false;
+  const serverGlobalAttls = std.getenv('ZWE_zowe_network_server_tls_attls') == 'true' ? true : false;
+  const clientLocalAttls = std.getenv('ZWE_components_zaas_zowe_network_client_tls_attls') == 'true' ? true : false;
+  const serverLocalAttls = std.getenv('ZWE_components_zaas_zowe_network_server_tls_attls') == 'true' ? true : false;
+  const clientAttls = clientGlobalAttls || clientLocalAttls;
+  if ((clientGlobalAttls !== false) && (clientLocalAttls !== false) && (!clientAttls)) {
+    // If client attls not explicitly false OR truthy, have client follow server attls variable. it simplifies common case in which users want both.
+    return serverGlobalAttls || serverLocalAttls;
```
Martin-Zeithaml, line 410: Sorry, I did not noticed Sean's comment, it has to more complicated. This [code](https://github.com/zowe/zlux-server-framework/blob/v3.x/staging/lib/util.js#L510-L520) is expecting 3-value logic: `true|false|undefined`. To keep the logic the same, we need to get 3 values too, e.g.: 
```
function getBooleanEnv(variableName) {
    const value = std.getenv(`${variableName}`);
    if (value === "true") {
        return true
    }
    if (value === "false") {
        return false
    }
    return undefined
}

function isClientAttls() {
  const clientGlobalAttls = getBooleanEnv('ZWE_zowe_network_client_tls_attls');
...
```

Otherwise this statement is never hit.

pablocarle: I can apply this changes to make it more appropriate, but I think the current one covers the scenario well as it cares only about when it's actually true.

Martin-Zeithaml (approved):

PR #4121, opened 2025-01-09
Print error message on invalid extension install package

1000TurquoisePogs:
Today, if you try to install an extension with a filename ending in something other than .tar, .pax, or .zip you'll get an error like:

TypeError: cannot read property 'rc' of undefined
  at execute (/zowe/bin/commands/components/install/extract/index:83)

This is because there is no "else" case on the file extension types.
So, this PR adds an "else" that just prints the error:

> Error ZWEL0318E File extension invalid. Supported file extensions: .pax, .tar, .zip


Martin-Zeithaml (approved):

Martin-Zeithaml:
Could we use `file` to determine the filetype rather than to force users to use extensions? If you rename `package.pax` to `package`, it is still PAX.

1000TurquoisePogs:
`file` would be nice if it worked well, but on z/OS 3.1 I just tested a few things and... 
* it doesnt understand .zip
* it doesnt understand some kinds of .tar
* it doesnt know the difference between a .pax that's USTAR and a .tar thats USTAR
I think it would cause more issues than just sticking to a fixed list of extensions.

Issue #4180, opened 2025-02-07
`init apf` support authLoadlib only with a reasonable default


MarkAckert:
## Apfauth
Help:
- `zowe.setup.dataset.authLoadlib` is the user custom APF LOADLIB. This field is
  **optional**. If it's not defined, `SZWEAUTH` from `zowe.setup.dataset.prefix` data
  set will be APF authorized.
- `zowe.setup.dataset.authPluginLib` is the user custom APF PLUGINLIB.
  You can install Zowe ZIS plugins into this load library.

`zowe.setup.dataset.authLoadlib` and `zowe.setup.dataset.authPluginLib` omitted in config and this is generated by `ZWEGEN00`:

```
//STDPARM DD *                                          
SH cd '/zowe/' &&                       
cd bin/utils &&                                         
export LOADLIB='CFG.ZOWE.SETUP.DATASET.AUTHLOADLIB' &&  
export LOADLOC=SMS &&                                   
export PLUGLIB='CFG.ZOWE.SETUP.DATASET.AUTHPLUGINLIB' &&
export PLUGLOC=SMS &&                                   
./opercmd.rex "SETPROG APF,ADD,DSN=$LOADLIB,$LOADLOC" &&
./opercmd.rex "SETPROG APF,ADD,DSN=$PLUGLIB,$PLUGLOC"   
//*                                                     
```

When trying `zwe init apfauth -c ./zowe.yaml`:

`ERROR: Error ZWEL0157E: zowe.setup.dataset.authLoadlib is not defined in Zowe YAML configuration file.`

Expectation:
If `zowe.setup.dataset.authLoadlib` and `zowe.setup.dataset.authPluginLib` omitted, the JCL will authorize only `zowe.setup.dataset.prefix`.`SZWEAUTH`.

_Originally posted by @Martin-Zeithaml in https://github.com/zowe/zowe-install-packaging/issues/4107#issuecomment-2643323806_

MarkAckert:
will be addressed in #4107

Martin-Zeithaml:
Similar with `zwe init mvs`, `zowe.setup.dataset.authLoadlib` and `zowe.setup.dataset.authPluginLib` should be both optional:
* `zowe.setup.dataset.authLoadlib` - as we claim in help, `"SZWELOAD"` will be used
* `zowe.setup.dataset.authPluginLib` - in "API ML" only scenario, this is not needed

`zwe init mvs -c zowe.no.authPluginLib.yaml`
[...]

PR #4192, opened 2025-02-12
Init MVS could skip some steps

Martin-Zeithaml:
## Migrate to JCL
An attempt to solve comment in #4180:

### Maximal Init MVS Scenario
All datasets are created and all members were copied.

### Minimal Init MVS Scenario
If all `parmlib`, `authLoadlib`, `authPluginLib` not set, do nothing and ends with success.


files/SZWESAMP/ZWEIMVSP
```diff
@@ -0,0 +1,43 @@
...
+/*
+//MEMBCPY EXEC PGM=IDCAMS
+//  SET OUTPARM={zowe.setup.dataset.parmlibMembers.zis}
+//SYSPRINT DD  SYSOUT=*
+//PARMIN  DD DISP=SHR,
+//  DSN={zowe.setup.dataset.prefix}.SZWESAMP(ZWESIP00)
+//PARMOUT DD DISP=OLD,
+//  DSN={zowe.setup.dataset.parmlib}(&OUTPARM)
+//SYSIN DD *
+  REPRO INFILE(PARMIN) OUTFILE(PARMOUT)
```
Martin-Zeithaml: @1000TurquoisePogs Would this help for `zwe init stc`? `ZWEMCOPY` is using repro and I think (but I am not sure), IDCAMS is after all using TSO commands anyway. If `ZWEMCOPY` is better solution, then the "IDCAMS JCL" might be also better solution.

INSTALLATION.md
```diff
@@ -111,7 +111,7 @@ When the JCL is prepared, the following jobs can be submitted to perform the fol
...
+|Create Instance Datasets|**Purpose:** Create datasets for Zowe's PARMLIB content and non-ZFS extension content for a given Zowe Instance<br><br>**Action:**<br>1) Allocate PDSE FB80 dataset with at least 15 tracks named from Zowe parameter `zowe.setup.dataset.parmlib`<br>2) Allocate PDSE FB80 dataset with at least 30 tracks named from Zowe parameter `zowe.setup.dataset.authPluginLib`<br>3) Copy ZWESIP00 member from `zowe.setup.dataset.prefix`.SZWESAMP into `zowe.setup.dataset.parmlib`|[ZWEIMVSP](https://github.com/zowe/zowe-install-packaging/tree/feature/v3/jcl/files/SZWESAMP/ZWEIMVSP)<br/><br/>[ZWEIMVSZ](https://github.com/zowe/zowe-install-packaging/tree/feature/v3/jcl/files/SZWESAMP/ZWEIMVSZ)|
```
MarkAckert: links should be relative directories? (./files/SZWESAMP/ZWEIMVSP)

MarkAckert: Actually I see this is for every link here...

smpe/bld/SMPMCS.txt
```diff
@@ -50,7 +50,9 @@
 ++SAMP(ZWEIKRT1)     SYSLIB(SZWESAMP) DISTLIB(AZWESAMP) RELFILE(2) .
 ++SAMP(ZWEIKRT2)     SYSLIB(SZWESAMP) DISTLIB(AZWESAMP) RELFILE(2) .
 ++SAMP(ZWEIKRT3)     SYSLIB(SZWESAMP) DISTLIB(AZWESAMP) RELFILE(2) .
-++SAMP(ZWEIMVS)      SYSLIB(SZWESAMP) DISTLIB(AZWESAMP) RELFILE(2) .
```
MarkAckert, line 53: IIRC we need to keep the old members around, or the SMP/e upgrades will fail. We can remove these with a clean fmid

Martin-Zeithaml: OK, I will keep the original names and create only one new.

MarkAckert (approved):

PR #4199, opened 2025-02-18
Update ZWEISTC to use DISP=SHR (V3)

1000TurquoisePogs:
This PR fixes the DISP property of the proclib used in the job ZWEISTC.
When it was DISP=OLD, an exclusive lock was requested which is very likely to fail and hang.
Using DISP=SHR is better for this command, as it is non-disruptive.

This fixes a bug in which using the job ZWEISTC may never complete, and prevent others from accessing the system proclib.

Related to https://github.com/zowe/zowe-install-packaging/pull/4198


Martin-Zeithaml:
Duplicate of #4195.

PR #4208, opened 2025-02-24
`zwe init apfauth` more optional

Martin-Zeithaml:
`zwe init apfauth` update:
* Removing `authLoadlib: ${{ zowe.setup.dataset.prefix }}.SZWEAUTH` from `defaults.yaml`
  * This could be set only if `zowe.setup.dataset.prefix` is mandatory

```jcl
//STDPARM DD *
SH cd '{zowe.runtimeDirectory}' &&
cd bin/utils &&
LOADLIB='{zowe.setup.dataset.authLoadlib}' &&
LOADLOC=SMS &&
./opercmd.rex "SETPROG APF,ADD,DSN=$LOADLIB,$LOADLOC" &&
PLUGLIB='{zowe.setup.dataset.authPluginLib}' &&
PLUGLOC=SMS &&
./opercmd.rex "SETPROG APF,ADD,DSN=$PLUGLIB,$PLUGLOC";
//*
```
If `zowe.setup.dataset.authLoadlib` not set, we will use `zowe.setup.dataset.prefix.SZWEAUTH`.
If `zowe.setup.dataset.authPluginLib` not set, `PLUGLIB=..` + 2 following lines are deleted.
If `zowe.setup.dataset.authPluginLib` same as `zowe.setup.dataset.authLoadlib`, `PLUGLIB=...` + 2 following lines are deleted.

## Update
There was a bug in `ZWEGEN00` causing undefined `yaml` items were propagated as `CFG.` + `"yaml property path"`, e.g. for `zowe.setup.dataset.prefix` -> `CFG.ZOWE.DATASET.PREFIX`. Originally it was keeping undefined template variable as it was and with this fix we will be doing the same again.

For example: if there is no `zowe.runtimeDirectory` in yaml, this will be the same before and after `generate` action: 
* `SH cd '{zowe.runtimeDirectory}' &&` 

## Changes
* `.help` - mentioning `jcllib`
* `index.ts` - when detects need for update, it will delete major part of shell and recreate it again. The goal was to have simple code.
* `ZWEGEN00` - condition for checking uninitialized variable was incorrectly nested 
* `ZWEIAPF2` - `export` not needed, lines reorganized
* `defaults.yaml` - default for `authLoadlib` removed

### DIFF
--- a/files/SZWESAMP/ZWEIAPF2
+++ b/files/SZWESAMP/ZWEIAPF2
@@ -12,12 +12,17 @@
 //*********************************************************************
 //*
 //*  This JCL is used to set APF for the two datasets of Zowe
-//*  Which need it. You can issue this, or use another
-//*  Way to accomplish the task.
+//*  which need it. You can issue this, or use another
+//*  way to accomplish the task.
 //*
 //*  The following variables are derived from the zowe YAML config:
-//*  LOADLIB: the dataset that holds the APF portion of Zowe
-//*  PLUGLIB: The dataset that holds the extensions for ZIS.
+//*  LOADLIB: the dataset that holds the APF portion of Zowe.
+//*    If not set, zowe.setup.dataset.prefix + SZWEAUTH is used.
+//*  PLUGLIB: the dataset that holds the extensions for ZIS.
+//*    If not set, no PLUGLIB authorization is done.
+//*
+//*  Note: if you plan NOT to use Zowe Cross Memory Server at all,
+//*    you can skip this step.
 //*
 //*********************************************************************
 //*
@@ -28,10 +33,15 @@
 //STDPARM DD *
 SH cd '{zowe.runtimeDirectory}' &&
 cd bin/utils &&
-export LOADLIB='{zowe.setup.dataset.authLoadlib}' &&
-export LOADLOC=SMS &&
-export PLUGLIB='{zowe.setup.dataset.authPluginLib}' &&
-export PLUGLOC=SMS &&
+LOADLIB='{zowe.setup.dataset.authLoadlib}' &&
+if [ "${LOADLIB}" = "CFG.ZOWE.SETUP.DATASET.AUTHLOADLIB" ]; then
+  LOADLIB='{zowe.setup.dataset.prefix}.SZWEAUTH';
+fi &&
+LOADLOC=SMS &&
 ./opercmd.rex "SETPROG APF,ADD,DSN=$LOADLIB,$LOADLOC" &&
-./opercmd.rex "SETPROG APF,ADD,DSN=$PLUGLIB,$PLUGLOC"
+PLUGLIB='{zowe.setup.dataset.authPluginLib}' &&
+if [ "${PLUGLIB}" != "CFG.ZOWE.SETUP.DATASET.AUTHPLUGINLIB" ]; then
+  PLUGLOC=SMS &&
+  ./opercmd.rex "SETPROG APF,ADD,DSN=$PLUGLIB,$PLUGLOC";
+fi
 //*
--- a/files/SZWESAMP/ZWEIAPF2
+++ b/files/SZWESAMP/ZWEIAPF2
@@ -34,14 +34,9 @@
 SH cd '{zowe.runtimeDirectory}' &&
 cd bin/utils &&
 LOADLIB='{zowe.setup.dataset.authLoadlib}' &&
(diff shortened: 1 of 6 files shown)


files/SZWESAMP/ZWEIAPF2
```diff
@@ -28,10 +33,15 @@
 //STDPARM DD *
 SH cd '{zowe.runtimeDirectory}' &&
 cd bin/utils &&
-export LOADLIB='{zowe.setup.dataset.authLoadlib}' &&
-export LOADLOC=SMS &&
-export PLUGLIB='{zowe.setup.dataset.authPluginLib}' &&
-export PLUGLOC=SMS &&
+LOADLIB='{zowe.setup.dataset.authLoadlib}' &&
+if [ "$LOADLIB" = "CFG.ZOWE.SETUP.DATASET.AUTHLOADLIB" ]; then
```
MarkAckert: What's the origin of `CFG.ZOWE.SETUP.DATASET.AUTHLOADLIB` ? Is this generated by configmgr when yaml fields are missing?

Martin-Zeithaml: Yes and no. It is generated by `ZWEGEN00`, technically it is uninitialized `rexx` variable, which is the result of missing yaml property not found by `rexx configmgr`.

Martin-Zeithaml: The question I am asking is: should we make all JCL as simple as possible and do the complicated processing in javascript? Or should we include some logic into JCL to keep the code more simple?

This solution is somewhere in between.

MarkAckert: Yup, the solution you went with makes sense to me. As a default position I'd support simpler JCL since system programmers and administrators may review it and we'd want to prioritize their experience during install and config, though it's not a hard rule.

I would prefer if `ZWEGEN00` set the `{zowe.setup.dataset.authLoadlib}` variable to something which conveyed an undefined or unset field rather than dumping the name of the rexx variable itself, but its not a huge deal. This looks good as-is.

Martin-Zeithaml:
## Tests

Description shows the DSN segment used for `authLoadlib` and `authPluginLib`.
In the output, looking if specific string is there or not.

```diff
+  [1/8][expected rc=  0, result rc=  0  ]: /zowe/bin/zwe install --ds-prefix "ZOWE.PR#4107.A" --allow-overwrite

+ [2/8][expected rc=100, result rc=100  ]: /zowe/bin/zwe init apfauth --help

+ [3/8][expected rc=  0, result rc=  0  ]: /zowe/bin/zwe init apfauth -c ./yaml/zowe.apf00.yaml --generate --dry-run
  [   ][Description                     ]: auth(), plug()
+ [   ]['SZWEAUTH' found                ]
+ [   ]['PLUGLOC' not found             ]

+ [4/8][expected rc=  0, result rc=  0  ]: /zowe/bin/zwe init apfauth -c ./yaml/zowe.apf01.yaml --generate --dry-run
  [   ][Description                     ]: auth(), plug(ZWESPLUG)
+ [   ]['SZWEAUTH' found                ]
+ [   ]['ZWESPLUG' found                ]

+ [5/8][expected rc=  0, result rc=  0  ]: /zowe/bin/zwe init apfauth -c ./yaml/zowe.apf10.yaml --generate --dry-run
  [   ][Description                     ]: auth(ZWESALL), plug()
+ [   ]['ZWESALL' found                 ]
+ [   ]['PLUGLOC' not found             ]

+ [6/8][expected rc=  0, result rc=  0  ]: /zowe/bin/zwe init apfauth -c ./yaml/zowe.apf11.yaml --generate --dry-run
  [   ][Description                     ]: auth(ZWESALL), plug(ZWESPLUG)
+ [   ]['ZWESALL' found                 ]
+ [   ]['ZWESPLUG' found                ]

+ [7/8][expected rc=  0, result rc=  0  ]: /zowe/bin/zwe init apfauth -c ./yaml/zowe.apfdn.yaml --generate --dry-run
  [   ][Description                     ]: auth(SZWEAUTH), plug()
+ [   ]['SZWEAUTH' found                ]
+ [   ]['PLUGLOC' not found             ]

+ [8/8][expected rc=  0, result rc=  0  ]: /zowe/bin/zwe init apfauth -c ./yaml/zowe.apfdd.yaml --generate --dry-run
  [   ][Description                     ]: auth(ZWESALL), plug(ZWESALL)
+ [   ]['ZWESALL' found                 ]
+ [   ]['PLUGLOC' not found             ]
```

MarkAckert (approved):

PR #4209, opened 2025-02-26
Better label for variable: Existing zowe runtime directory

### LINKED ISSUE #4073: PSWI: Existing Zowe runtime directory not pre-populated
When going thru a new PSWI install, the z/OSMF workflows created an install location for Zowe.  However, when inputting the "main variables" I have to provide the "Existing Zowe runtime directory":

image

Since this was created in prior steps, can the workflow just "know" where my installation exists?

pavelzlatnik:

#### PR type
- [X] Bugfix 
- [ ] Feature 

#### Relevant issues

Fixes #4073

Better variable label, so it is less confusing for customer (in description is written it is mountpoint to TARGET library, it is allocated as TARGET library via PSWI or SMP/E job, but variable is originally named runtime directory). It serve as a "runtime" just for install launcher).
image

- 

- [ ] Yes
- [X] No

---

---

---

---
_multi-line description_

Doc issue/PR number:

### DIFF
--- a/workflows/files/ZWEAMLCF.xml
+++ b/workflows/files/ZWEAMLCF.xml
@@ -100,11 +100,11 @@
         </string>
     </variable>       
     <variable name="zowe_runtimeDirectory" scope="instance" visibility="public">
-        <label>Existing Zowe runtime directory</label>
-        <abstract>Zowe runtime (root) directory with libraries</abstract>
-        <description>Existing Zowe runtime.&lt;br /&gt;
+        <label>Existing Zowe Launcher runtime directory</label>
+        <abstract>Zowe Launcher runtime (root) directory with libraries</abstract>
+        <description>Existing Zowe Launcher runtime.&lt;br /&gt;
 If you created ZOWE via PSWI, &lt;br /&gt;
-zowe runtime is your Target Mount Point from PSWI deployment. &lt;br /&gt;
+Zowe Launcher runtime is your Target Mount Point from PSWI deployment. &lt;br /&gt;
 It is used by ZWESLSTC or by zwe command</description>
         <category>zowe</category>
         <string valueMustBeChoice="false" multiLine="false">
--- a/workflows/files/ZWECONF.properties
+++ b/workflows/files/ZWECONF.properties
@@ -88,13 +88,13 @@ zowe_setup_vsam_volume=
 zowe_setup_vsam_storageClass=
 
 # zowe_runtimeDirectory
-# Label: Existing Zowe runtime directory
-# Abstract: Zowe runtime (root) directory with libraries
+# Label: Existing Zowe Launcher runtime directory
+# Abstract: Zowe Launcher runtime (root) directory with libraries
 # Category: zowe
 # Description:
-#  Existing Zowe runtime.<br />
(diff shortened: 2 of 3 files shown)


1000TurquoisePogs (changes requested):
The Zowe runtime directory is where almost everything EXCEPT the launcher resides.

The launcher resides in datasets, not unix, and is not involved in the installation process.

The documentation and YAML refer to this as the "Zowe Runtime" currently, so I do not understand needing a change in the workflow.

Can you explain what problem you are solving?

pavelzlatnik:
> The Zowe runtime directory is where almost everything EXCEPT the launcher resides.
> 
> The launcher resides in datasets, not unix, and is not involved in the installation process.
> 
> The documentation and YAML refer to this as the "Zowe Runtime" currently, so I do not understand needing a change in the workflow.
> 
> Can you explain what problem you are solving?

Hi,
Based of my understanding and what zowe guys told to me, they referred as "Launcher" to **zwe**, and calling zwe command (zwe install, zwe init, etc.)
And it is exactly what is done under this mounted directory, from this directory is launched zowe installer, which creates RUNTIME datasets (referred as launcher to me, if there is another thing which is referred as Launcher as well, there is already even more confusion). 

Biggest issue for me is that this zowe directory is labeled as "zowe runtime" from the project beginning, but is mounted on TARGET dataset created within SMP/E via this JCL. 
https://github.com/zowe/zowe-install-packaging/blob/8f9a375837657a6c5e41fdc38804f00ae24c9033/workflows/templates/smpe-install/ZWE4ZFS.vtl#L73

Which lead in past to confusion between TARGET and RUNTIME datasets, and has another issues hidden (I had already experience with product, which was technically similar to how zowe operates and how is installed/configured. There were separated groups which makes products SMP/E installations, and groups which makes product configurations in product runtime.  Person who wanted to "launch" the installer has no permission to change file permissions (like umask 0002;
chmod -R 775) to TARGET dataset based on policy set for production systems, they should always do modification on RUNTIME datasets only, ie. RUNTIME datasets which has been copy of TARGET datasets.) I just wanted to distinguish the label, so installing person has at least some hint that this is not RUNTIME datasets with all its consequences.

1000TurquoisePogs:
Zowe's launcher is a program here https://github.com/zowe/launcher which resides in SZWEAUTH dataset as member ZWELNCH. The ZFS directories are unrelated to it. It calls "zwe" at runtime, but "zwe" has many other uses that are not at runtime, such as install time and configuration.
The "zowe runtime" is called that in the primary configuration file
https://github.com/zowe/zowe-install-packaging/blob/8f9a375837657a6c5e41fdc38804f00ae24c9033/example-zowe.yaml#L281
And throughout the documentation ( https://docs.zowe.org/search?q=zowe%20runtime%20directory )

The runtime directory is not the runtime datasets, in that one is in unix and the other is not, but the thing they both have in common is that they comprise the "runtime", that is, the read-only code that can be used by 1-or-more configurations of zowe "instances".
These concepts and names exist with or without SMPE, with or without z/OSMF workflows, as there is a variety of ways to install and configure Zowe.

pavelzlatnik:
Based on various opinions and feedback, variable kept as it is now.

PR #4213, opened 2025-02-26
jclHeader support for install workflows

MarkAckert:
PR that allows users to define JOB statement parameters used by ZWEGENER, etc. within z/OSMF configuration workflows.

Notes:

* z/OSMF has a way to define a JOB card for workflows run on a system, but I couldn't find any way to access it, either in system or workflow variables. So we have to define our own variable capturing JCL statement parameters.
* The default JOB positional and keyword parameters will be empty, so no JCL headers are applied.
* Syntax is set so users should not input JCL identifier fields, i.e. '//'. The workflow handles this for users when it creates the zowe.yaml file. In the current design, line 1 has no identifier fields since it follows `//<JOBNAME> JOB`, and lines 2...n will prepend `//    `.
* z/OSMF does not support regex validation for multiline strings that I could see. It just won't apply the regex I defined.

Tested:

zwe init generate using zowe.yaml from:
* Workflow with no job statement params
* Workflow with single line job statement
* Workflow with multi-line job statement
* One invalid job statement (error is only uncovered in `init generate`, which happens in step 3 of workflow)

### DIFF
--- a/workflows/files/ZWEAMLCF.xml
+++ b/workflows/files/ZWEAMLCF.xml
@@ -11,6 +11,15 @@
         <General/>
     </workflowInfo>
     <!--variables-->
+    <variable name="zowe_setup_job_stmt_parms" scope="global" visibility="public">
+      <label>JOB statement parameters used by Zowe during installation and configuration.</label>
+      <abstract>JOB statement parameters used by Zowe during installation and configuration.</abstract>
+      <description>These additional JOB statement parameters are optional, and by default empty. These are separate from z/OSMF JOB statements.</description>
+      <category>zowe</category>
+      <string multiLine="true" valueMustBeChoice="false">
+        <default></default>
+      </string>
+    </variable>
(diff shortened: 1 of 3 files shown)


MarkAckert:
cc @muzikovam @pavelzlatnik , this may be of interest to you

pavelzlatnik:
> cc @muzikovam @pavelzlatnik , this may be of interest to you

Hi Mark, you are right. Some time ago, we wanted the same thing in one project. We end up with this solution with extra variable for job header. There is no system variable from which you can read it. (And because IBM did implementation that for workflow registered via workflow panel or z/OSMF API is used one JOB header, but for workflow which is registered via PSWI is used another JOB header, and they are not the same, each is defined in other place and they could be different....., so it could differ if you jump to the workflow from panel, API or PSWI).

muzikovam:
LGTM

PR #4215, opened 2025-02-27
Add variable for apiml registration for attls

1000TurquoisePogs:
This PR adds a zowe yaml variable zowe.network.proxyType
It defaults to 'https', unless either attls is on globally or just for gateway.
This variable is not intended to be edited.
It is intended to be determined by the state of attls, and used by anyone onboarding their servers to apiml, so that no decision making would need to be done by apiml extenders.

### DIFF
--- a/files/defaults.yaml
+++ b/files/defaults.yaml
@@ -84,6 +84,10 @@ zowe:
       # Default to caching service entry as it predates this one
       name: "${{ ()=> { if (components['caching-service']?.storage?.vsam?.name) { return components['caching-service'].storage.vsam.name } else { return '' } }() }}"
 
+  network:
+    # This is a variable derived from gateway's ATTLS state to be used by servers proxied through it in eureka or static definitions.
+    proxyType: "${{ components.gateway?.zowe?.network?.server?.tls?.attls == true ? 'http' : zowe.network?.server?.tls?.attls == true ? 'http' : 'https' }}"
+      
 
   configmgr:
     # STRICT=quit on any error, including missing schema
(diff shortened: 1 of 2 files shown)


files/defaults.yaml
```diff
@@ -84,6 +84,10 @@ zowe:
       # Default to caching service entry as it predates this one
       name: "${{ ()=> { if (components['caching-service']?.storage?.vsam?.name) { return components['caching-service'].storage.vsam.name } else { return '' } }() }}"
 
+  network:
+    # This is a variable derived from gateway's ATTLS state to be used by servers proxied through it in eureka or static definitions.
+    proxyType: "${{ components.gateway?.zowe?.network?.server?.tls?.attls == true ? 'http' : zowe.network?.server?.tls?.attls == true ? 'http' : 'https' }}"
```
Martin-Zeithaml, line 89: Is this correct? `components.gateway?.zowe?.network?.server?.tls?.attls`
The `zowe` property under the `gateway`?

1000TurquoisePogs: yes, this property does exist.
"zowe.network" can be applied globally or per-component.
Prior, "zowe.network.server" and "zowe.network.client" had separate ATTLS properties.
But through discussions recently it appears that having client/server individually customizable for ATTLS is pointlessly complicated and so you should consider "zowe.network.client.tls.attls" to be deprecated and this is why this variable considers just 2 values: the server tls of the gateway config, and the server tls of the global config as a fallback if the gateway one is not set.

Martin-Zeithaml (approved):

PR #4223, opened 2025-03-05
remove pem from keyring output

1000TurquoisePogs:
This PR removes mention of pem from keyring output of `zwe init certificate`.
PEM in keyring  has not been needed for years, but keeps on showing up in user YAMLs either due to this output, or some documentation.

This also removes adding 4 slashes into the keyring output, as this is apparently no longer needed in v2.18.


JoeNemo (approved):
Approved.

PR #4227, opened 2025-03-07
DO NOT MERGE: Enhancment to support adding steplib entries during 'zwe components install' using '-d' option

### LINKED ISSUE #711: Multi-steplib zis proposal
As requested in https://github.com/zowe/zss/issues/703, this ticket is to discuss an implementation proposal

Goal: have `zwe components install -o /the/extension[.pax]` go beyond its current task of just copying unix content into the extensions directory, but also to update the ZIS STC steplib list.

# Idea 1
`zwe components install -o /the/extension[.pax] -d THE.DS1,THE.DS2`
... Have -d serve as a list of datasets to append to the steplib.

Problems: 
* `zwe components upgrade` and `zwe components uninstall` exist. How do we know which steplibs to REMOVE? How can we track which extension added which steplib entry? a file in the workspace directory? something in zowe.yaml?

* I don't like this being manual. How can we detect the datasets so that users do not need to be prompted, and cannot do something wrong? I'd like to query SMPE's knowledge somehow.

# Idea 2
No new -d parameter, just manifest info & SMPE lookup.
... How can I query what datasets on a system are associated with an FMID? I think that's superior. If a Zowe extension manifest could have a field for stating its FMID, like:

```yaml
name: angular-sample
id: org.zowe.zlux.sample.angular
version: 2.16.0
fmid: AZWE999
```

Then, can I get a list of datasets associated with it via something involving https://www.ibm.com/docs/en/zos/3.1.0?topic=command-listing-by-fmid-fmidset ?

And, if there were multiple datasets, how do I know which one goes in the steplib?
Would we need a suffix pattern to follow, such as:

```yaml
name: angular-sample
id: org.zowe.zlux.sample.angular
version: 2.16.0
fmid: AZWE999 # Proposed new attribute. Used with steplibLLQ to find exact dataset.
zisPlugins:
  - id: plugin1
    steplibLLQ: ZISPLUG1 # Proposed new pattern. No copy occurs, but steplib entry added
  - id: plugin2
    steplibLLQ: ZISPLUG2
  - id: plugin3
    path: "/zisServer" # Current zowe pattern. it copies from here to the zis plugin ds
```

Gautham-coder:

#### PR type
 - [x ] Feature 

Fixes: https://github.com/zowe/zss/issues/711, https://github.com/zowe/zss/issues/703

Makes `zwe components install -o /the/extension[.pax`] go beyond its current task of just copying Unix content into the extensions directory, but also to update the ZIS STC and AUX STC steplib list.

### Implementation:

Implementing IDEA 1 from https://github.com/zowe/zss/issues/711, where we supply the the list of dataset through arguement '-d', as shown below:

`zwe components install -o /u/XXXX/file-manager-1.3.0-20241112.201716.pax -c example-zowe.yaml -d DSN1, DSN2...`

--Processing the input list to check for valid DSNs:

- check for duplicates,
- wrong dataset names/length and
- if dataset exists.

--When adding the dataset to the STEPLIB list:

- Check if the dataset exits in the list if so don't add.
- if commented then uncomment it.
- Also if there a empty lines or comments in the JCL to skip them correctly

- [ ] Yes
- [x] No

The description of the this ticket and above explanation should suffice

---
* Affected function: `zwe component install` *

---
* Description: Makes `zwe components install -o /the/extension[.pax`] go beyond its current task of just copying Unix content into the extensions directory, but also to update the ZIS STC and AUX STC steplib list.*

---

---
_multi-line description_

Doc issue/PR number: 

#### Example
Before:
STEPLIB_Before
STEPLIB_Before

After:
STEPLIB_After
STEPLIB_After


bin/commands/components/install/cli.ts
```diff
@@ -16,6 +16,6 @@ import * as configmgr from '../../../libs/configmgr';
 
 const upgrade = false; //would be true for upgrade.ts
 
-index.execute(std.getenv('ZWE_CLI_PARAMETER_COMPONENT_FILE'), std.getenv('ZWE_CLI_PARAMETER_AUTO_ENCODING'), (std.getenv('ZWE_CLI_PARAMETER_SKIP_ENABLE') === 'true'), std.getenv('ZWE_CLI_PARAMETER_HANDLER'), std.getenv('ZWE_CLI_PARAMETER_REGISTRY'), (std.getenv('ZWE_CLI_PARAMETER_DRY_RUN') === 'true'), upgrade);
+index.execute(std.getenv('ZWE_CLI_PARAMETER_COMPONENT_FILE'), std.getenv('ZWE_CLI_PARAMETER_AUTO_ENCODING'), (std.getenv('ZWE_CLI_PARAMETER_SKIP_ENABLE') === 'true'), std.getenv('ZWE_CLI_PARAMETER_HANDLER'), std.getenv('ZWE_CLI_PARAMETER_REGISTRY'), (std.getenv('ZWE_CLI_PARAMETER_DRY_RUN') === 'true'), upgrade, std.getenv('ZWE_CLI_PARAMETER_STEPLIB'));
```
1000TurquoisePogs: replace with

`std.getenv('ZWE_CLI_PARAMETER_STEPLIB') ? std.env('ZWE_CLI_PARAMETER_STEPLIB').split(',') || undefined`

making the input a string array or undefined.

bin/commands/components/install/index.ts
```diff
@@ -18,8 +18,11 @@ import * as fs from '../../../libs/fs';
 import * as config from '../../../libs/config';
 import * as componentlib from '../../../libs/component';
 import { HandlerCaller, getHandler, getRegistry } from '../handlerutils';
+import * as xplatform from 'xplatform';
+import * as zosdataset from '../../../libs/zos-dataset';
+import * as zosfs from '../../../libs/zos-fs';
 
-export function execute(componentFile: string, autoEncoding?:string, skipEnable?:boolean, handler?: string, registry?: string, dryRun?: boolean, upgrade?: boolean) {
+export function execute(componentFile: string, autoEncoding?:string, skipEnable?:boolean, handler?: string, registry?: string, dryRun?: boolean, upgrade?: boolean, stepLibs?: string) {
```
1000TurquoisePogs: change to `stepLibs?: string[]`
if it exists, it should be an array.

bin/commands/components/install/index.ts
```diff
@@ -62,10 +65,137 @@ export function execute(componentFile: string, autoEncoding?:string, skipEnable?
...
+  }
+  else {
+    common.printError(`Error: Could not store updated contents in tmp unix file ${proclibMemberAsUnixFile}.`);
+  }
+  return update;
+}
+
+function processSteplibArgs(inputArgs: string): string[] {
+  return Array.from(
+    new Set(
```
1000TurquoisePogs: this logic is good, but have the input already be a `string[]` to better alert other developers that it is intended to be an array.

bin/commands/components/install/.exclusive-parameters
```diff
@@ -3,3 +3,4 @@ auto-encoding|e|string||auto|||If we want to automatically tagging the module fi
 skip-enable||boolean|||||Install component without enabling it for use.
 registry|r|string|||||Specifies the registry to searh within instead of the default. The registry must be compatible with the manager used.
 handler||string|||||Specifies the registry handler name used with the package registry, instead of the default. The handler must be compatible with the registry used.
+steplib|d|string|||||STEPLIB entries that need to be added to the zis and aux STC's.
```
1000TurquoisePogs: Needs a better description and possibly a better name.

Comma separated list of datasets to reference as ZIS STEPLIB entries for loading plugins. Requires that the datasets exist prior and are APF authorized. If not provided when installing a component containing ZIS plugins, the plugins will be copied into the ZIS APF Plugin dataset.

As a result name could be zis-plugin-datasets?

1000TurquoisePogs (changes requested):
This logic is the alternative to copying a zis plugin from unix into the apf plugin dataset.
But I don't see that action being prevented here, so it looks like both actions are done.

You need to change it so only one is done.

The pre-existing ZIS install code is

zwe components install -> zwe components install process-hook -> [processZisPluginInstall](https://github.com/zowe/zowe-install-packaging/blob/v3.x/staging/bin/libs/component.ts#L597) -> [zisPluginInstall](https://github.com/zowe/zowe-install-packaging/blob/v3.x/staging/bin/libs/component.ts#L649)

zisPluginInstall does 2 actions
1) makes a parmlib entry for the new zis plugin: good, keep this.
2) copies the zis plugin into the apf plugin dataset: alternative to what you're doing, make it conditional


As a result, I think you should take your new functions and move them into [component.ts](https://github.com/zowe/zowe-install-packaging/blob/v3.x/staging/bin/libs/component.ts) so that you can make them a conditional alternative to the code here:
https://github.com/zowe/zowe-install-packaging/blob/v3.x/staging/bin/libs/component.ts#L666-L674

This also means you'll need to pass the plugin list down to component.ts and rename your functions to better describe their purpose.

"updateSTCS" is too generic. `updateZisStcs` is better, for example.
Because when in component.ts, you do not want people to mistake these functions as doing something generic - these are specifically for ZIS.

bin/commands/components/install/cli.ts
Gautham-coder: Replaced the code as suggested.

bin/commands/components/install/index.ts
Gautham-coder: Resolved this as suggested.

bin/commands/components/install/.exclusive-parameters
Gautham-coder: Updated the option name and description as suggested.

bin/commands/components/install/index.ts
Gautham-coder: Resolved this as suggested.

Gautham-coder:
Added in logic to update the steplib entries of zis stcs as an alternative to copying a zis plugin from unix into the apf plugin dataset as suggested.

bin/libs/component.ts
```diff
@@ -1001,6 +1012,136 @@ export function processComponentDiscoverySharedLibs(componentDir: string): boole
...
+  if (zosdataset.isDatasetExists(`${proclib}(${auxMember})`)) {
+    if (!updateStcSteplibEntries(proclib, auxMember, stepLibEntries)) {
+      common.printError(`${proclib}(${zisMember}) could not be updated`);
+      return false;
+    }
+  }
+  return true;
+}
+
+function updateStcSteplibEntries(proclib: string, member: string, stepLibEntries: string[]): boolean {
```
1000TurquoisePogs, line 1225: pass through dry-run parameter all the way to here. when present, stop at the copy-to-tmp area

bin/libs/component.ts
```diff
@@ -1001,6 +1012,136 @@ export function processComponentDiscoverySharedLibs(componentDir: string): boole
...
+
+  // 2. Get the updated content
+  const updatedContent = updateStepLib(proclibMemberAsUnixFile, stepLibEntries);
+
+  // 3. store the updated content in the same temporary Unix file
+  let rc = xplatform.storeFileUTF8(proclibMemberAsUnixFile, xplatform.AUTO_DETECT, updatedContent);
+  if(!rc) {
+    // 4. Copy the contents from the temporary file into a temporary dataset
+    const tmpDataset = zosdataset.createDatasetTmpMember(jclLib);
+    rc = zosdataset.copyToDataset(proclibMemberAsUnixFile, `${jclLib}(${tmpDataset})`, "", true);
```
1000TurquoisePogs, line 1243: this is where you should stop with dry run
it is ok to copy to the temp dataset, but not okay to overwrite the legitimate dataset on dry run mode.

bin/libs/component.ts
```diff
@@ -1001,6 +1012,136 @@ export function processComponentDiscoverySharedLibs(componentDir: string): boole
...
+  const jclLib = ZOWE_CONFIG.zowe.setup.dataset.jcllib;
+  const prefix = ZOWE_CONFIG.zowe.setup.dataset.prefix;
+  let update = false;
+
+  // 1. create temporary Unix file
+  const proclibMemberAsUnixFile = fs.createTmpFile(`${proclib}`);
+  zosfs.copyMvsToUss(`${proclib}(${member})`, proclibMemberAsUnixFile);
+
+  // 2. Get the updated content
+  const updatedContent = updateStepLib(proclibMemberAsUnixFile, stepLibEntries);
```
1000TurquoisePogs, line 1236: print out this content, so users can see what it contains.
you can also print out the original, unedited version, so users can see the difference.

also print out the location of the original, and the location of the temp dataset, so that users with dry run, can find the temp one, and then do whatever they want.

1000TurquoisePogs: this will result in a lot printing, so, let's decide whether or not some of the prints should be DEBUG vs INFO. not sure yet.

PR #4238, opened 2025-03-14
Default to bring your own keyring

1000TurquoisePogs:
This PR changes the default behavior of `zwe init` to no longer run `zwe init certificate`, and moves all `zowe.setup.certificate` content away from the example-zowe.yaml and into a new folder zowe/files/examples/setup/certificate

## With this PR, to create certificates in a new zowe install, `zwe init` is not enough anymore!
Users must use the contents of zowe/files/examples/setup/certificate and then run either `zwe init --create-certificate` or `zwe init certificate`

The purpose is to have the YAML and `zwe init` meet users expectations better.
Many users bring their own keyrings when doing a zowe setup, yet they end up for some reason running certificate setup anyway. They may encounter errors, or at least end up with something they didn't want.

Further, users often get confused between the difference of `zowe.setup.certificate` and `zowe.certificate`. So, if they get confused by the one they do not want to use anyway, why not move it?

This PR is not a breaking change in the sense that it does not disrupt existing installs.
I do not believe a HOLD statement is needed.
However, new installs will no longer get certificates created by default.
So, documentation will need to call out that as an action item for those who do want that, unpopular as it is.

The new example yaml references the location of the setup certificate example yamls, and they contain comments within that explain how to use them.

I have made a slight edit to `zwe init certificate` to make these YAML as small as possible: I found that `zowe.setup.dataset.prefix` was needed for keyring creation, but not for pkcs12 creation, so I moved it into the appropriate block.


MarkAckert:
To avoid users needing to duplicate information, could we adjust the init certifcate command so it takes both a zowe.yaml configuration file and a cert scenario file? 

`zwe init certificate -c zowe.yaml --cert-config path/to/cert/setup/scenario.yaml`

bin/commands/init/index.sh
```diff
@@ -93,7 +93,9 @@ if [ "${ZWE_CLI_PARAMETER_SKIP_SECURITY_SETUP}" != "true" ]; then
   zwecli_inline_execute_command init apfauth
   zwecli_inline_execute_command init security
 fi
-zwecli_inline_execute_command init certificate
+if [ "${ZWE_CLI_PARAMETER_CREATE_CERTIFICATE}" != "true" ]; then
```
Martin-Zeithaml: 1. Should check for "is equal": `if [ "${ZWE_CLI_PARAMETER_CREATE_CERTIFICATE}" = "true" ]; then`
2. This is a braking change as the new default is "Do not process certificates".

bin/commands/init/.parameters
```diff
@@ -3,3 +3,4 @@ skip-security-setup||boolean|||||Whether should skip security related setup.
 security-dry-run||boolean|||||Whether to dry run security related setup.
 ignore-security-failures||boolean|||||Whether to ignore security setup job failures.
 update-config||boolean|||||Whether to update YAML configuration file with initialization result.
+create-certificate||boolean|||||Whether to use the contents of zowe.setup.certificate to create Zowe's TLS certificates.
```
Martin-Zeithaml: `.help` would probably need some update too.

Martin-Zeithaml (dismissed):

MarkAckert:
The changes look good. Before merging this in: 

* Is there a docs-site update?
* Do we need any changes to PSWI configuration workflows?

I can take a look at the second question.

MarkAckert:
I found a few more YAML fields, `verifyCertificates`, `zosmf.*`, and `externalDomains` which are used by the `init certificate` command and not included in the new scenario YAML files. These are probably the last fields users can fill out but I don't know for sure since it's hard to audit w/ defaults. 

Thinking about the UX...is it easier to ask users to always copy over scenario yaml file content to the main zowe.yaml? This lines up with what we already tell them for the `zwe init --create-certificate` case, and helps avoid redundant/missed `zowe.yaml` fields, but comes at the cost of some convenience - `zwe init certificate -c </path/to/scen.yml>` won't work anymore. Imo asking users to copy over the content and append it is worth the tradeoff in this case, but I'd like to hear other opinions. I also considered asking users to use configmgr syntax, which solves redundant or missed fields: `-c FILE(/path/to/scenario.yml):FILE(/path/to/zowe.yaml)`, though my gut says for most users this will be more error-prone than simply copying over content. 

cc @Martin-Zeithaml @1000TurquoisePogs

example-zowe.yaml
```diff
@@ -90,147 +90,6 @@ zowe:
     #     # STC name of Zowe ZIS Auxiliary Server
     #     aux: ZWESASTC
 
-    # >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
```
Martin-Zeithaml, line 93: This will remove the certificate scenario from this example, which is fine. But there is a default definition in `defaults.yaml` file. 

Do we want to keep it there?

MarkAckert: We're relying on the default definitions in combination with the updated scenario YAML files, so I think we want to keep it around (for now anyway). 

Do you have any concerns on keeping the defaults?

Martin-Zeithaml: No special concern, it is part of the defaults from the beginning. Just asking.

bin/commands/init/index.ts
```diff
@@ -123,7 +123,18 @@ export function execute(allowOverwrite?: boolean, dryRun?: boolean, ignoreSecuri
...
+    common.printMessage(`Zowe appears to be missing certificate setup. Complete zowe.certificate in your Zowe YAML before starting Zowe.`);
+    common.printMessage(`Zowe does not create certificates, keystores, or truststores by default, so verify that zowe.certificate in your Zowe YAML is valid.`);
+    common.printMessage(`You can review Zowe certificate requirements at https://docs.zowe.org/stable/user-guide/configure-certificates#zowe-certificate-requirements`);
+    common.printMessage(`---------`);
+    common.printMessage(`If you want Zowe to assist in certificate creation, the command "zwe init certificate" can be run with YAML properties from "zowe.setup.certificate"`);
+    common.printMessage(` Examples of "zowe.setup.certificate" are included in "${std.getenv('ZWE_zowe_runtimeDirectory')}/files/examples/setup/certificate"`);
+    common.printMessage(` After reviewing the included examples and selecting one, append the contents of the example file into your Zowe YAML and run "zwe init certificate".`);
+    common.printMessage(`---------`);
+  }
```
Martin-Zeithaml, line 137: This should be in `index.sh` too.

files/examples/setup/certificate/scenario-1.yaml
```diff
@@ -0,0 +1,60 @@
...
+################################################################################
+
+# >>>> Certificate setup scenario 1
+# PKCS12 (keystore) with Zowe generate certificates.
+#
+# Pre-requisite: your main zowe.yaml file is configured.
+#
+# To use this example, complete the configuration below and run 
+#  `zwe init --create-certificate` or `zwe init certificate` with either:
+#    - The contents of this file appended to your main 'zowe.yaml' file
```
Martin-Zeithaml: This instruction is confusing:
```yaml
# ./yaml/zowe.1.yaml
zowe:
  setup:
    dataset:
      prefix: ZOWE.PR#4238
  externalDomains:
    - one.example.com
    - two.example.com
# ------------------------------------
# Literally appended certificate setup
# ------------------------------------
zowe:
  setup:
    certificate:
      type: PKCS12
      pkcs12:
        # ----------------------
        # The only customization
        # ----------------------
        directory: /zowe/certificate
        # # Lock the keystore directory to only accessible by Zowe runtime user and group.
        # lock: true
        # Certificate alias name. Optional, default value is localhost.
        # Note: please use all lower cases as alias.
        name: localhost
        # Keystore password. Optional, default value is password.
        password: password
        # Alias name of self-signed certificate authority. Optional, default value is local_ca.
        # Note: please use all lower cases as alias.
        caAlias: local_ca
        # Password of keystore stored self-signed certificate authority. Optional, default value is local_ca_password.
        caPassword: local_ca_password
      validity: 3650
```

```
$ /zowe/4238/bin/zwe config get --config ./yaml/zowe.1.yaml --path .zowe.setup.certificate.pkcs12.directory
$ /var/zowe/keystore
```

`/var/zowe/keystore` is taken from `defaults.yaml` (which is not a problem or important). The second `zowe:` section is **ignored**.

MarkAckert (approved):

PR #4242, opened 2025-03-21
Schema change for Infinispan

DivergentEuropeans:
Info taken from: https://github.com/zowe/api-layer/blob/v3.x.x/schemas/caching-schema.json

### DIFF
--- a/schemas/zowe-yaml-schema.json
+++ b/schemas/zowe-yaml-schema.json
@@ -440,6 +440,37 @@
                   ]
                 }
               }
+            },
+            "infinispan": {
+              "type": "object",
+              "additionalProperties": false,
+              "description": "VSAM configurations if you are using VSAM as Caching Service storage",
+              "properties": {
+                "mode": {
+                  "type": "string",
+                  "description": "VSAM data set with Record-Level-Sharing enabled or not",
+                  "enum": ["NONRLS", "RLS"],
+                  "default": "NONRLS"
+                },
+                "volume": {
+                  "type": "string",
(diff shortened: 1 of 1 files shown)


1000TurquoisePogs (changes requested):
Awaiting evidence that there is a zowe.setup.infinispan that needs to be documented.

DivergentEuropeans:
We don't need necessarily a `zowe.setup.infinispan` but we need an infinispan schema to adjust infinispan customization in the Wizard. Traditionally, this Wizard checks this section but you're right that architecturally we don't need it there? in that location? Wizard could be made to point directly to API ML schema but it seems strange to check 1 place for the vsam customization it can do great today, but check a different place for the same task for infinispan: customization of caching service storage method

1000TurquoisePogs:
The zowe schema details things that have an effect to zowe. component schemas detail things that affect just one component.
Nothing to date uses "zowe.setup.infinispan", it does not exist.
Given that infinispan vs vsam vs others are properties unique to the caching-service component,
It seems best to me to design the wizard to have per-component configuration and to consume each components schemas as needed to answer prompts. That may take some enhancement in the wizard to organize it as such but will pay off in future component customization use cases.

1000TurquoisePogs:
I am going to close this for now - reopen if the discussion changes. I think it would be great to see the wizard consume the schema of each component it finds in `zowe.runtimeDirectory/components` by reading their manifest.yaml files, but this is very much an "advanced mode" in that the schemas of components can contain very exotic, do-not-touch sort of stuff.

PR #4244, opened 2025-03-21
Migrate to JCL: `zwe init security`: schema update and code removal

Martin-Zeithaml:
## Schema update
With schema update and defaults we can be sure those values will be always defined:
```yaml
    security:
      product: RACF
      groups:
        admin: ZWEADMIN
        stc: ZWEADMIN
        sysProg: ZWEADMIN
      users:
        zowe: ZWESVUSR
        zis: ZWESIUSR
      stcs:
        zowe: ZWESLSTC
        zis: ZWESISTC
        aux: ZWESASTC
```

`product` is `enum`, the rest is string (no blanks allowed).
User has no way (if validation is used) to provide null, undefined or empty value.

## ZWESECUR
Reverting back `workflows/templates/ZWESECUR.vtl`, this JCL example is still provided for customers and it should contain valid statements. When resolved by ZWEGEN00 now, there are lines like this:
```jcl
//         SET  PRODUCT=#[[RACF]]# 
//         SET ADMINGRP=#[[ZWEADMIN]]#
...
```

## Help

Small help update.

## `zwe init security`
* The code in `zwe init security` is no longer checking `zowe.setup.security.*`
* Wrong message `ZWEL0157E` deleted and replaced with checking of real and configured ESM
  * In case user sets wrong ESM, we will print warning update needed or do the update if `--update-config` was used  
## Defaults
Security product is detected in defaults:
```
 product: "${{ [ 'RACF', 'ACF2', 'TSS' ].includes(zos.getEsm()) ? zos.getEsm() : 'RACF' }}"
```

## ZWEGEN00
Is checking and using detected ESM, it must correctly resolve `ZWESECUR` and `ZWENOSEC`.


MarkAckert (approved):

bin/commands/init/security/index.ts
```diff
@@ -77,5 +67,10 @@ export function execute(dryRun?: boolean, ignoreSecurityFailures?: boolean) {
   common.printMessage(`WARNING: Due to the limitation of the ZWEI${securityPrefix} job, exit with 0 does not mean`);
   common.printMessage(`         the job is fully successful. Please check the job log to determine`);
   common.printMessage(`         if there are any messages indicating a problem.`);
+  if (esmWarning) {
```
MarkAckert, line 71: I know this pre-dates the current PR, but I'm curious - why is this a warning instead of an error? Is there any reason why we shouldn't convert this to an error and quit out of the init command before JCL is submitted?

Martin-Zeithaml: There are couple things, which must be done together:
* When preparing `jcllib`, `ZWEGEN00` is deleting all JCLs, which are NOT for the current ESM
  * `ZWEGEN00` is traversing control blocks to get the used ESM
* This command then must use the current ESM
* Generally there is an approach to do the maximum - like if you set `infinispan` for caching, we skip `VSAM` and reporting success
* As far as I know, there is only one complicated task: for example: if you would like to see resolved TSS JCLs on LPAR with RACF or ACF2
* **The same approach must be done in `zwe init certificate`, the shell script is taking only config value**
* I am fine with current state, but we can discuss this more and change it
* Possibly we can offer to update config when mismatch detected

MarkAckert: I'll open a new issue as a low priority. I am also OK with the current state since it yields basically the same user experience - if the wrong security manager is set, the JCL will run and return with an error. Thank you

bin/commands/init/.help
```diff
@@ -30,11 +30,11 @@ The following are the JCL submitted by each command as needed
     - ZWEIAPF2
 - `zwe init security`
     - (If RACF) ZWEIRAC
-    - (If RACF and z/OS 2.4) ZWIRACZ
+    - (If RACF and z/OS 2.4) ZWEIRACZ
```
MarkAckert: I don't think we support 2.4 anymore? Java 17's minimum supported z/OS is 2.5, right?

Martin-Zeithaml: Yes, you are right:
_z/OS V2.4 reached end of support on 30 September, 2024._

Martin-Zeithaml: It is still [here](https://github.com/zowe/zowe-install-packaging/blob/9863f36824a247377eaeb8c50e77cf3c1af732e9/bin/commands/init/security/index.ts#L63), but I think, we can keep it there for now.

PR #4248, opened 2025-03-25
v3: Add AT-TLS automated test case

MarkAckert:
This is not ready for merge while we finalize a working ATTLS configuration on the backend system.

Related PRs:
https://github.com/zowe/zlux-server-framework/pull/617
https://github.com/zowe/zss/pull/797

This PR adds a new extended test to the nightly test suite which covers AT-TLS + Basic Install. The nightly test is able to run on any of the 3 zzow systems; all are setup with AT-TLS enabled and the exact same keyring + certificate configuration. Certificates will need to be replaced on the backend systems by 2035 or if the system IP+domain changes, whichever happens first.

The AT-TLS test keeps the existing playbook behavior of leaving the system state as-is after a test is run. This has been valuable for debugging errors in test cases by leaving us with the failed system state, so we can poke, prod, and modify as necessary to get to the bottom of a failure. By adding AT-TLS to the system state, the tests _must_ clean out any old AT-TLS settings as part of every test case run. Therefore, these playbooks will no longer run on a system without an AT-TLS configuration file available, even when you leave AT-TLS disabled in the playbook configuration. You could substitute a dummy config file and that would likely pass, but this particular scenario won't be tested.

AT-TLS configuration is stored in the `playbooks` directory.


playbooks/roles/configure/tasks/configure_attls.yml
```diff
@@ -0,0 +1,90 @@
...
+  loop:
+    - ATTLS_Rules.txt
+    - zwe-setup-attls.js
+  loop_control:
+    loop_var: filename_to_upload
+
+# !!! Important
+# This runs the script that sets the at-tls rules for the system.
+# If zowe_attls_enabled is not defined or empty, the default will be false (AT-TLS disabled).
+# If the zowe_attls_policy_file is not the system's active AT-TLS policy, then this fectively does nothing.
```
Martin-Zeithaml: then this fectively does nothing -> then this effectively does nothing

struga0258 (approved):

MarkAckert:
This will be merged in despite the ATTLS test suite failing. We need to wait for the zlux-server-framework PR, but the automation needs to be merged upstream because it proactively manages AT-TLS, ensuring it's disabled on the Marist systems.

Martin-Zeithaml (approved):

PR #4251, opened 2025-03-26
Re-enable VT,TN3270 tests

MarkAckert:
As part of the angular upgrade during the initial 3.0 releases some Zowe Desktop plugin test cases were failing due to compatibility issues with the newer angular. These should all be resolved now and we can re-enable the test cases.

The editor component, `test-09-editor`, is not ready for test yet.


1000TurquoisePogs:
I'm not certain about the editor, but the terminals should work!
The editor as of now works in v2 desktop mode of v3, but not in the v3 desktop of v3.
we've been working on it, so it will be there, but its on the fence whether we make it for 3.2

1000TurquoisePogs (approved):

PR #4254, opened 2025-03-27
feat: API ML modulith transition

pablocarle:

#### PR type

- [ ] Feature

#### Relevant issues

https://github.com/zowe/api-layer/issues/2919

In this first iteration, the new API ML component contains in a single JVM package the functionality of the Gateway Service and the Discovery Service.
This PR makes changes to startup procedures to make this transition without breaking changes.

If the apiml component is enabled (default), it will forcefully disable current gateway and discovery services.

To use the previous way, it's still possible to disable apiml component.

- [ ] No

Doc issue/PR number:


bin/libs/config.ts
```diff
@@ -35,8 +35,54 @@ const cliParameterConfig:string = function() {
     return (value as string);
 }();
 
+function isApimlEnabled(): boolean {
+  const config = configmgr.getZoweConfig();
+  const defaultEnabledComponents = component.getEnabledComponents();
+  const componentEnabled = defaultEnabledComponents.includes('apiml') || config.components['apiml'].enabled;
```
pavel-jares-bcm:
```suggestion
  const componentEnabled = config.components['apiml']?.enabled;
```

bin/libs/config.ts
```diff
@@ -35,8 +35,54 @@ const cliParameterConfig:string = function() {
...
+  return componentEnabled 
+  && config.components['gateway'].enabled == false
+  && config.components['discovery'].enabled == false
+  && config.components['caching-service'].enabled == false
+  && config.components['api-catalog'].enabled == false
+  && config.components['zaas'].enabled == false
+  && config.components['apiml'].port == config.components['gateway'].port
```
pavel-jares-bcm:
```suggestion
  && !config.components['gateway'].enabled
  && !config.components['discovery'].enabled
  && !config.components['caching-service'].enabled
  && !config.components['api-catalog'].enabled
  && !config.components['zaas'].enabled
  && !config.components['apiml'].port == config.components['gateway'].port
```

bin/libs/config.ts
```diff
@@ -35,8 +35,54 @@ const cliParameterConfig:string = function() {
...
+        enabled: false
+      },
+      zaas: {
+        enabled: false
+      },
+      "api-catalog": {
+        enabled: false
+      },
+      apiml: {
+        port: config.components['gateway'].port
```
pavel-jares-bcm:
```suggestion
        enabled: true,
        port: config.components['gateway'].port
```
or it is in default

manifest.json.template
```diff
@@ -56,45 +56,42 @@
       "version": "^3.0.0-SNAPSHOT",
       "artifact": "*.pax"
     },
+    "org.zowe.apiml.apiml-package": {
```
pavel-jares-bcm, line 63: What about the apiml's docker?

bin/libs/config.ts
```diff
@@ -35,8 +35,55 @@ const cliParameterConfig:string = function() {
...
+  const config = configmgr.getZoweConfig();
+  const defaultEnabledComponents = component.getEnabledComponents();
+  const componentEnabled = config.components['apiml']?.enabled;
+  return componentEnabled 
+  && !config.components['gateway'].enabled
+  && !config.components['discovery'].enabled
+  && !config.components['caching-service'].enabled
+  && !config.components['api-catalog'].enabled
+  && !config.components['zaas'].enabled
+  && !config.components['apiml'].port == config.components['gateway'].port
```
1000TurquoisePogs: i'm guessing there's something in your code currently that cannot tolerate `apiml.enabled` being set at the same time as `gateway.enabled` for example, as seen here and in draft documentation.

it's going to cause disruption and i think it can be better.

how about:

1) if apiml.enabled, zowe startup will deliberately skip running the start scripts of gateway, discovery, caching-service, api-catalog, zaas. it will just do apiml start script.

2) at the same time, it will trick all components and extensions into believing all of those components are still enabled. in that way, no changes to anyone else's codebase would really be needed.

workflows/files/ZWECONF.xml
```diff
@@ -509,6 +509,28 @@ How we want to verify SSL certificates of services. Valid values are:
             <default>IZUDFLT</default>
         </string>
     </variable>
+    <variable name="components_apiml_enabled" scope="instance" visibility="public">
```
MarkAckert: Should this be removed? Along with the `ZWECONF.properties` updates? I saw you marked these as missing in the sync rules, so if you want to keep them defaulted in both ZWECONF and ZWEAMLCF, removing the variable definitions would clarify that.

workflows/zowe-yaml-sync-rules.json
```diff
@@ -71,6 +80,8 @@
       "rule_target": "ZWEAMLCF",
       "rule_type": "wf_missing_vars",
       "rule_entry": [
+        "components_apiml_enabled",
```
MarkAckert: redundant with earlier change (line 49) for `rule_target: all` ?

bin/libs/config.ts
```diff
@@ -35,8 +35,55 @@ const cliParameterConfig:string = function() {
     return (value as string);
 }();
 
+function isApimlEnabled(): boolean {
```
MarkAckert: who calls `isApimlEnabled` and `enableApiml`?

pablocarle: Noone currently, the idea was to perform the changes automated in the zowe.yaml but I think the proposal in https://github.com/zowe/launcher/pull/160 is a better solution for this, I will remove this code.

bin/libs/config.ts
pablocarle: I assume this comment happened before the proposed PRs, I agree with the changes, I will remove this code (it was unused anyway). The only change expected to happen now to enable this single-service APIML mode is to have `components.apiml.enabled` as `true`

manifest.json.template
pablocarle: I believe those are still unsupported for modulith, at very least untested. It should be included in the final 3.4.0 version though.
It's part of the limitations listing in this technical preview already, so we can keep track and work on it.

workflows/files/ZWECONF.xml
pablocarle: I assume these will be defaulted differently in Zowe v3.3.0 and it will change in Zowe v3.4.0. I'm not very familiar with these workflow files, so I may need to have some guidance here.
The goal is to have the current workflows have this default to false (same as defaults.yaml), but in new PSWI (Zowe v3.4.0) we need to define it as enabled by default, I'm not sure if we will need a workflow step to give the option to switch it off from the installation.

pavel-jares-bcm (approved):

workflows/files/ZWECONF.xml
MarkAckert: I guess the question to ask is: do we want users to see the modulith is even an option in 3.3? Or is the intent to keep this more hidden and work with users directly who might benefit from trying out the modulith?

pablocarle: I think the answer for this technical preview is that it should be a hidden property and those who want to try it out the need to add it.
I think the idea is that for Zowe 3.4.0 the default is to have it enabled but it may not be a fully transparent change from the system configuration point of view

pablocarle: After discussing with the team we think we are going to add (in a future PR) a step to the main installation workflow to decide whether to disable the single-service mode.

MarkAckert: OK, I'll remove the variable choice from the workflow for this release and make sure the builds are passing. Modulith mode will always default to false (in 3.3*).

MarkAckert:
will merge in after https://github.com/zowe/api-layer/pull/4258

MarkAckert (approved):

PR #4271, opened 2025-04-07
Print versions and ESM for better diagnosis

1000TurquoisePogs:
This PR prints out already gathered node and java versions whenever validateNodeHome() or validateJavaHome() are called.
Additionally, at startup (through zwe internal start prepare), ~zOS version~ and ESM type are printed.
This allows the reader of a normal log to have information they would normally have to turn on tracing, run zwe support, or look up elsewhere to find.

I was going to add zOS version print out, but the value printed did not look human readable, so postponing that.

### DIFF
--- a/bin/commands/internal/start/prepare/index.ts
+++ b/bin/commands/internal/start/prepare/index.ts
@@ -27,6 +27,7 @@ import * as java from '../../../../libs/java';
 import * as javaCI from '../../../../libs/java_ci';
 import * as node from '../../../../libs/node';
 import * as zosmf from '../../../../libs/zosmf';
+import * as zoslib from '../../../../libs/zos';
 
 //# This command prepares everything needed to start Zowe.
 const cliParameterConfig = std.getenv('ZWE_CLI_PARAMETER_CONFIG');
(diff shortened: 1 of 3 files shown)


bin/commands/internal/start/prepare/index.ts
```diff
@@ -440,6 +440,7 @@ if (zoweVersion) {
 export function execute() {
   common.printFormattedInfo("ZWELS", "zwe-internal-start-prepare", `Zowe version: v${zoweVersion}`);
   common.printFormattedInfo("ZWELS", "zwe-internal-start-prepare", `build and hash: ${runtimeManifest.build.branch}#${runtimeManifest.build.number} (${runtimeManifest.build.commitHash})`);
+  common.printFormattedInfo("ZWELS", "zwe-internal-start-prepare", `ESM: ${zos.getEsm()}`);
```
Martin-Zeithaml, line 445: With aditional include of `bin/libs/zos as zoslib`
``common.printFormattedInfo("ZWELS", "zwe-internal-start-prepare", `${zoslib.formatZosVersion('{osname}={major}.{minor}')}`);``

Martin-Zeithaml (approved):

PR #4275, opened 2025-04-09
More ZWE message updates

MarkAckert:
Finding and addressing more inconsistencies between IDs and contents.

~Todo: Add detection for messages in commands which are missing from .errors and vice versa~

The zwe detection is improved and tracks error message definitions relative to their local usage within commands. It cross references this with the shared, root `.errors` under commands. Automation also now scans sources under `bin/libs`, and expects the shared `.errors` file will contain **all** the error messages used within libs.

As a result, this will now complain about error definitions for commands when the error itself doesn't appear within the command but instead appears via call to another module/library
For example, `zwe start` `.errors` definition contains this message:

`ZWEL0064E|64|failed to run command os.pipe - Cannot start component %s`.

This error message does not come from `zwe start` itself, but a utility under `bin/libs`. Moving the message to the common command `.errors` fixes the problem as far as automation is concerned, but creates a change in the output of zwe documentation on docs.zowe.org. Instead of the `ZWEL0064E` being identified as an error specific to the `zwe start` command, it's listed under the common error section. See here (ignore the missing 0064E, not caused by this PR): https://docs.zowe.org/stable/appendix/zwe_server_command_reference/zwe/zwe-start#errors .

I expect users running a `zwe` command who get error output will go to the docs.zowe.org command reference and search for the error. As long as its on the page, they will be OK, so we can settle with 064E appearing in the common / inherited block.


MarkAckert:
@Martin-Zeithaml this is now ready for review. I made more significant changes than I originally planned to and laid out why in the PR description.

bin/commands/init/apfauth/.errors
```diff
@@ -1,4 +1,5 @@
-ZWEL0157E|157|%s (%s) is not defined in Zowe YAML configuration file.
 ZWEL0319E|319|zowe.setup.dataset.jcllib does not exist, cannot run. Run 'zwe init', 'zwe init generate', or submit JCL zowe.setup.dataset.prefix.SZWESAMP(ZWEGENER) before running this command.
+ZWEL0324E|324|The dataset specified in %s does not exist.
+ZWEL0325E|325|Failed to prepare ZWEIAPF2
```
Martin-Zeithaml: Should we define this more general? 
`ZWEL0325E|325|Failed to prepare %s` in case we need this for another JCL/else.

Martin-Zeithaml (approved):

PR #4276, opened 2025-04-09
Pre-flight checks for zowe init and startup

1000TurquoisePogs:
I am prototyping a set of utilities that can be invoked by `zwe validate` by a user or zwe itself in order to check if a Zowe install is likely to run successfully or not prior to starting up all the servers and getting complex errors that users may not understand the meaning of.

At this point nothing is set in stone. It is rather a place to dump ideas to return to later.
If this PR becomes more official, this description will be updated to reflect what it can do.

So far, the following exists:

* zwe validate config: an alias of zwe config validate
* zwe validate port available: runs the same "is port available" code that apiml startup scripts do, but does it for all enabled components - in the future, apiml would not need that script as this command could handle it.
* zwe validate directories: validates that key directories exist and are not within the runtime directory since that is not supported
* zwe validate dependencies: validates that dependencies requested by the environment are met (such as how java is required for apiml, and node is required for app-server, but these are both conditional) - this is what `zwe internal start prepare` already does, so the code could be moved here to be reusable
* warns upon possibly okay, but untested dependency versions (node, java, zos)

It could be helpful to invoke these at the last step of `zwe init` as well as at each startup, to catch as much user error as possible.
I am intending to write a bypass, set in the YAML, so that if any validation fails it can be ignored rather than to prevent some execution, in case our validation logic itself is an error.


bin/commands/validate/port/bind/index.ts
```diff
@@ -0,0 +1,59 @@
...
+  bindUtilPath+='bin/utils/bind-test';
+
+  enabledComponents.forEach((component)=> {
+    let port = ZOWE_CONFIG.components[component].port;
+    if (port) {
+      if (network.isPortAvailable(port)) {
+        common.printFormattedInfo(`ZWELS`, `zwe-validate-port-available`, `${component} port ${port} available.`);
+        let listenAddress = '0.0.0.0';
+        if (ZOWE_CONFIG.components[component].zowe?.network?.server?.listenAddresses) {
+          listenAddress = ZOWE_CONFIG.components[component].zowe.network.server.listenAddresses;
```
1000TurquoisePogs, line 37: [0]

bin/libs/java_ci.ts
```diff
@@ -57,19 +58,29 @@ export function validateJavaHome(javaHome:string|undefined=std.getenv("JAVA_HOME
...
+    }
 
     if (tooLow) {
       common.printError(`Java ${javaVersionShort} is less than the minimum level required of Java ${JAVA_MIN_VERSION}.`);
       return false;
     }
-
-    common.printDebug(`Java ${javaVersionShort} is supported.`);
+    if (tooHigh) {
+      common.printMessage(`WARNING: Java ${javaVersionShort} is higher than the latest known working version for Zowe, v${JAVA_TESTED_VERSION}.`);
```
1000TurquoisePogs, line 81: should probably omit warning when startupcheck for dependency is set false

bin/commands/validate/port/available/index.ts
```diff
@@ -0,0 +1,42 @@
...
+  under the terms of the Eclipse Public License v2.0 which
+  accompanies this distribution, and is available at
+  https://www.eclipse.org/legal/epl-v20.html
+ 
+  SPDX-License-Identifier: EPL-2.0
+ 
+  Copyright Contributors to the Zowe Project.
+*/
+
+import * as common from '../../../../libs/common';
```
1000TurquoisePogs, line 12: unsure if port-available is worth keeping if port-bind does the same and more.

1000TurquoisePogs:
PR 4447 is the template for future PRs of this type, closing this prototype.

PR #4282, opened 2025-04-11
zwe install in JCL

MarkAckert:
**Updated Description**
This PR:
* migrates `zwe install` to configmgr driven scripts, removing the prior shell script execution path
* changes `ZOWE_CONFIG` so it's only loaded once explicitly upon request by calling code rather than loaded by default via code execution during import
* changes `zos-jes` job submission so we always submit jobs via temp files and `cat`. `echo` has a variety of formatting errors depending on the JCL content.
* comes with test cases

**Original Description:**
I pulled commits from #4057 and took another approach to the dual paths on `zwe install`. Instead of splitting CLI or requiring the `zowe.yaml`, this no longer loads `zowe.yaml` by default within configmgr and instead lazily loads it when a caller requires it. Using deferred, in-block `import` or `require` would've been even better than the current solution, but the configmgr quickjs implementation doesn't support it. The latest quickjs release for linux does support it however...I don't know the difficulty in porting that to z/OS and preserving the in-block `import` capabilities. 

For instance, this deferred import worked in linux:
```typescript
import * as std from 'std';

const f = std.tmpfile();

await import('os').then((os) => {
    const dirs = os.homedir;
    console.log(dirs);
});
console.log('done');
```


Martin-Zeithaml:
## Current behavior

When `--ds-prefix`, ignore config completely:

`asdf` does not exist, but Zowe is successfully installed to `ZOWE.TEST`:
```
zwe install --ds-prefix zowe.test -c asdf
...
>> Zowe MVS data sets are installed successfully.
```

## To consider

If possible, prefer config - in the future we might add more "install" parameters (like storage class, volume, ...) and it would be probably best to keep everything in config.

`--ds-prefix` is used only if no `--config`

Possible problem is with `export`, which is recommended in our doc to be used for shortening the commands:

```
export ZWE_CLI_PARAMETER_CONFIG=/path/to/zowe.yaml
zwe install
```

## Suggestion 1

| `ds-prefix` | `config` | `export` | Result / to be used | Note |
| :------------ | :--------- | :------- | :----- | :---- |
| false | false | false | Error | |
| false | false | true  | `export` | |
| false | true  | false | `config` | |
| false | true  | true  | `config` | Current behavior is to use `--config` in this situation |
| true  | false | false | `ds-prefix` |
| true  | false | true  | `ds-prefix` | To be able to override `export`, user can omit `--ds-prefix` to use `export` | 
| true  | true  | false | `config` | Maybe warning `--ds-prefix` is ignored? | 
| true  | true  | true  | `config` |  | 

## Suggestion 2

Remove the `--ds-prefix` parameter
* Simplifies the code
* Most of the customers are staring with `example-zowe.yaml` anyway

## Suggestion 3

Keep `--ds-prefix` and combine it with config:
* Some sites must specify JCL header
* In this case config must be used
* `--ds-prefix` will override the `zowe.setup.dataset.prefix`
* JCL header will be used

Martin-Zeithaml:
# Tests

## (weird) dataset names
All created, no problems with special characters, especially with `$`.
```
ZOWE.$.#.@.$-.#-.@-.SZWEAUTH
ZOWE.$.#.@.$-.#-.@-.SZWEEXEC
ZOWE.$.#.@.$-.#-.@-.SZWELOAD
ZOWE.$.#.@.$-.#-.@-.SZWESAMP

ZOWE.PR#4282.$$$$$$$$.SZWEAUTH
ZOWE.PR#4282.$$$$$$$$.SZWEEXEC
ZOWE.PR#4282.$$$$$$$$.SZWELOAD
ZOWE.PR#4282.$$$$$$$$.SZWESAMP

ZOWE.PR#4282.$1.SZWEAUTH      
ZOWE.PR#4282.$1.SZWEEXEC      
ZOWE.PR#4282.$1.SZWELOAD      
ZOWE.PR#4282.$1.SZWESAMP      

ZOWE.PR#4282.$1-#-@.SZWEAUTH  
ZOWE.PR#4282.$1-#-@.SZWEEXEC  
ZOWE.PR#4282.$1-#-@.SZWELOAD  
ZOWE.PR#4282.$1-#-@.SZWESAMP  
```

## General tests
```
[ 1/14][expected rc=100, result rc=100  ]: /pax/4282/bin/zwe install --help
[     ][Description                     ]: Help only

[ 2/14][expected rc=102, result rc=102  ]: /pax/4282/bin/zwe install --ds-prefix ZOWE.TEST
[     ][Description                     ]: --ds-prefix not supported

[ 3/14][expected rc=108, result rc=108  ]: /pax/4282/bin/zwe install
[     ][Description                     ]: Missing config returns ZWEL0108E

[ 4/14][expected rc=157, result rc=157  ]: /pax/4282/bin/zwe install --config 'FILE(./yaml/zowe.missing_prefix.yaml)'
[     ][Description                     ]: Missing prefix returns ZWEL0157E

[ 5/14][expected rc=  1, result rc=  1  ]: /pax/4282/bin/zwe install --config 'FILE(./yaml/zowe.validation_err1.yaml)'
[     ][Description                     ]: Validation error (regex)

[ 6/14][expected rc=  1, result rc=  1  ]: /pax/4282/bin/zwe install --config 'FILE(./yaml/zowe.validation_err2.yaml)'
[     ][Description                     ]: Validation error (null)

[ 7/14][expected rc=159, result rc=159  ]: /pax/4282/bin/zwe install --config ./yaml/zowe.basic.yaml
[     ][Description                     ]: Sample not found

[ 8/14][expected rc=  0, result rc=  0  ]: /pax/4282/bin/zwe install --config ./yaml/zowe.basic.yaml
[     ][Description                     ]: Successful install

[ 9/14][expected rc=  0, result rc=  0  ]: /pax/4282/bin/zwe install --config ./yaml/zowe.basic.yaml
[     ][Description                     ]: Skip due to missing --allow-overwrite
[     ]['Zowe MVS data sets installation skipped' found]

[10/14][expected rc=  0, result rc=  0  ]: /pax/4282/bin/zwe install -c ./yaml/zowe.dollar.yaml
[     ][Description                     ]: "$" in prefix

[11/14][expected rc=  0, result rc=  0  ]: /pax/4282/bin/zwe install -c ./yaml/zowe.dollar.yaml --allow-overwrite
[     ][Description                     ]: tsoDelete for dataset with "$"

[12/14][expected rc=  0, result rc=  0  ]: /pax/4282/bin/zwe install -c ./yaml/zowe.dollar.yaml --dry-run
[     ][Description                     ]: Prints the JCL
[     ]['//AUTHCPY EXEC PGM=BPXBATCH' found]

[13/14][expected rc=  0, result rc=  0  ]: /pax/4282/bin/zwe install --config 'PARMLIB(ZOWE.PR#4282.YAML(SIMPLE)):FILE(./yaml/zowe.env_only.yaml)' --allow-overwrite
[     ][Description                     ]: Config in DSN

[14/14][expected rc=  0, result rc=  0  ]: /pax/4282/bin/zwe install --allow-overwrite
[     ][Description                     ]: Config exported
```

bin/commands/install/index.ts
```diff
@@ -0,0 +1,123 @@
...
+                common.printMessage(`Warning ZWEL0301W: ${prefix}.${DATASETS[ds]} already exists and will not be overwritten. For upgrades, you must use --allow-overwrite.`);
+                skipJCL = true;
+            } else {
+                common.printMessage(`Warning ZWEL0300W: ${prefix}.${DATASETS[ds]} already exists. Members in this data set will be overwritten.`);
+                existingDatasets.push(`${prefix}.${DATASETS[ds]}`);
+            }
+        }
+    }
+
+    if (skipJCL) {
```
Martin-Zeithaml: Should we consider this?
`if (skipJCL && !dryRun) {`

With `--dry-run`, we can continue and print the JCL for user.

MarkAckert: 👍

Martin-Zeithaml (approved):

PR #4288, opened 2025-04-16
Debug 3.2 catalog error

MarkAckert:
Running automation to investigate the issue


MarkAckert:
The issue was found and addressed, closing.

PR #4294, opened 2025-04-23
Updated the schema to support new configuration parameter to trim syslog messages

Gautham-coder:
Updated the schema "zowe-yaml-schema.json" with new parameter "zowe.sysMessageTrim" which will used to trim syslog messages and print only from the sys-message-id. By default it will is disabled to be non-disruptive.

- [] Changelog updated (not updated yet due to automatic format conversion, will do it once this approach is confirmed)


1000TurquoisePogs (commented):
Can you also put this into defaults.yaml

https://github.com/zowe/zowe-install-packaging/blob/v3.x/staging/files/defaults.yaml

you say default=false, so just code it into that file.
Thanks!

Gautham-coder:
Sure @1000TurquoisePogs will do that. 
Thanks

MarkAckert:
Please normalize your parameter - either `sysTrimMessage` or `sysMessageTrim`.  The defaults.yaml has one, and the schema has the other, causing the build failure (schema validation failure for zowe.yaml).

Gautham-coder:
@MarkAckert Sure sorry, will correct it. Thanks

1000TurquoisePogs (approved):

Issue #3444, opened 2023-06-02
Handle IPCS queue full gracefully


pablocarle:
**Describe the bug**

Possibly related to #326 
Related to #3443 
We've been receiving reports, and we are able to reproduce in our systems. When IPCS queues get filled up, Zowe fails to prepare the instance and does not start.

https://www.ibm.com/docs/en/sdk-nodejs-zos/14.0?topic=zos-troubleshooting#d3724e264

This topic does not seem to be covered in the public documentation, either as troubleshooting or properly sizing the queues.

**To Reproduce**
1. Either make sure to stop Zowe incorrectly so the IPC queue will not clean up or reduce the size.
2. Start Zowe
3. Depending on the version used, logs could appear of the error or not. The error and solution is described in the IBM link above.

**Expected behavior**
Stop execution with appropriate message in logs.

1000TurquoisePogs:
I think we do cleanup of the queue at "internal start prepare" via this script https://github.com/search?q=repo%3Azowe%2Fzowe-install-packaging%20cleanup-ipc-mq&type=code but the problem is that this script is not called at "start component" time, so, if a component crashes post-startup, the launcher restarts it, skipping the "prepare" step where this cleanup gets run. So, we could call this script at the start phase too, so that continued restarts does not exhaust the finite queue resources.

MarkAckert:
addressed by https://github.com/zowe/zlux-app-server/pull/340 and https://github.com/zowe/zlux-app-server/pull/341, will be in v3.3.0

PR #4305, opened 2025-04-29
Remove cleanup-ipc-mq from zowe-install-packaging

MarkAckert:
## Updated description
Removes the IPCS cleanup script from zowe-install-packaging and `zwe internal start prepare`. #3444 will now be addressed by https://github.com/zowe/zlux-app-server/pull/340 , which includes the fixes to the cleanup-ipc-mq script. The cleanup will be invoked on every app-server startup.

## Old description
Addresses #3444 in combination with a launcher update. This PR:

* Creates a new zwe command dedicated to running the existing `bin/utils/cleanup-ipc-mq.sh` script.
* Modifies the `internal start prepare` to call this new zwe command instead of calling the `cleanup-ipc-mq.sh` script directly.
* Fixes the cleanup-ipc script. This was broken and we didn't realize it.

This new zwe command will be called by the launcher on component start, which lets us cleanup the IPC MQ after component crashes or intentional start/stop via MODIFY command. 

Manually tested and working on zzow11:
* first-time start causes multiple runs of cleanup-ipc-mq (one for each component)
* running`F APPL=STOP(component)`, then `F APPL=START(component)` invoked the cleanup script
* manually killing a component by pid triggered launcher's automatic restart, which invoked the cleanup script


Martin-Zeithaml:
The changes are fine, maybe one or two sentences to `.help` file.

But it seems, we have some legacy code not working as expected:

https://github.com/zowe/zowe-install-packaging/blob/fc9292e6065b82b5e3bf125fe0f4d524a1cda1df/bin/utils/cleanup-ipc-mq.sh#L30

The original idea was probably to take columns 1, 2 and 16. But on our dev system it is broken due to the format of the date, for example:
```
# Raw data
m      73738 0x61751d12 --rw------- ZWESVUSR ZWEADMIN ZWESVUSR ZWEADMIN        5       300  1M    314572800 Apr 27 00:46:06 Apr 28 02:43:17 Apr 27 00:44:48     394304   50726275

# Result of awk
id='ZWESVUSR'
for s in $(ipcs -a | awk 'match($1,"q|m|s") && $5 == "'${id}'" {print $1","$2":"$16}'); do
    echo "${s}"
fi
q,1162784:30
...
m,23738:Apr
...
s,200760:
```

And there are 11 columns for `s` (semaphores) option.

bin/commands/internal/utils/cleanup-ipcmq/index.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+_CEE_RUNOPTS="XPLINK(ON),HEAPPOOLS(OFF),HEAPPOOLS64(OFF)" ${ZWE_zowe_runtimeDirectory}/bin/utils/configmgr -script "${ZWE_zowe_runtimeDirectory}/bin/commands/internal/utils/cleanup-ipcmq/cli.js"
```
Martin-Zeithaml: [HEAPPOOLS/64](https://github.com/zowe/zowe-common-c/pull/497) could be omitted.

Martin-Zeithaml (approved):

1000TurquoisePogs (approved):

Issue #3543, opened 2023-08-02
Tracking issue for Removal of node.js dependency


Martin-Zeithaml:
Create an Epic or use this issue to record what was done, what is in progress and what is missing related to **Removal of Node.js dependency**.

This whole task is little bit disorganized, spread across several people and squads and there are popping the questions like "_Will it be this Friday done?_" over and over.

Martin-Zeithaml:
Config-converter: https://github.com/zowe/zowe-install-packaging/issues/3544
In progress: https://github.com/zowe/zowe-install-packaging/pull/3135

1000TurquoisePogs:
replacing curl.js is tracked in https://github.com/zowe/zowe-install-packaging/issues/3569

Martin-Zeithaml:
### Update 2023/10/02
* `fconv`
  * More research to be done
* [config-converter](https://github.com/zowe/zowe-install-packaging/issues/3544)
  * Seems not to be used during installation
* `jq`
  * Possibly replaceable by`fakejq` (already part of `zowe/bin/libs`)
  * Or `microjq`, which is implemented in `zowe-common-c`
  * More research to be done
* [curl](https://github.com/zowe/zowe-install-packaging/issues/3569)
   * Joe has some C (or Java?) implementation, which needs to be tested 
* [zwe init](https://github.com/zowe/zowe-install-packaging/pull/3135)
  *  Needs to be merged

Martin-Zeithaml:
### Update 2023/10/27
* `fconv`
  * More research to be done
* [config-converter](https://github.com/zowe/zowe-install-packaging/issues/3544)
  * Seems not to be used during installation
* `jq`
  * Possibly replaceable by`fakejq` (already part of `zowe/bin/libs`)
  * Or `microjq`, which is implemented in `zowe-common-c`
    * **Does not support arrays to be returned**
  * More research to be done
* [curl](https://github.com/zowe/zowe-install-packaging/issues/3569)
   * Joe has some C (or Java?) implementation, which needs to be tested 
* [zwe init](https://github.com/zowe/zowe-install-packaging/pull/3135)
  *  Needs to be merged
* [zwe install](https://github.com/zowe/zowe-install-packaging/pull/3635)
  * Because of unrelated bug, proper testing is required

Martin-Zeithaml:
### Update 2024/01/08

### Why Node.js removal?
* `node.js` is required for zowe installation
* Not every customer has `node.js`
[...]

PR #4307, opened 2025-05-01
Node Removal Tasks - v3.x

MarkAckert:
Addressing other node.js removal tasks as described in #3543. This PR is built on the JCL Init refactor to take advantage of improvements there and the zwe integration test framework.

Requires these 2 PRs to be merged:

As far as I know the below list is all remaining work to address node.js

Complete:
* Removing fconv+njq by changing occurrences of `read_yaml` to `read_yaml_configmgr`. **Note:** later changed so all `*_zowe_yaml_configmgr` variants are instead `*_zowe_yaml`, replacing the old code paths completely.
* Replacing config-converter by changing `*_zowe_yaml` functions in `json.sh` to `*_zowe_yaml_configmgr` variants.
* Replacing two config-convert commands (`convert` and `to-env`) with configmgr variants
* Replace ncert with enhancements to `keyring-utility`
* `zowe-install-packaging-tools` no longer a binary dependency.
* `org.zopencommunity.curl` introduced as a binary dependency.
* Replace curl.js with pre-built zopentools port of curl.
* Update test framework code to remove zowe-install-packaging-tools
* Test changes end-to-end.
* Remove many traces of `require_node`. 
* Support `curl` and `yq` transparently for container environments - no reliance on configmgr code.
* z/OSMF curl checks now always run, since `curl` is always available.

In progress:
* Test cases

To do:
* Done!

Changed Behaviors:
* `zwe migrate for kubernetes` outputs the `zowe.yaml |` block in the k8s ConfigMap **without** comments. In my mind, this is acceptable - users should not be extensively modifying this ConfigMap directly.
* note: keyring-utilites was changed to support `zwe`'s use cases. notably, the command syntax has changed and keyring private keys are output in `.p12` format. see: https://github.com/zowe/keyring-utilities/pull/13
* note: ncert was being used to generate csr signed certs with wildcard SANs. Java 8 keytool did not support wildcards, but 17+ does, and we require 17 for v3. Therefore, we went back to using Java keytool completely instead of branching to ncert for wildcard SANs.
* The `config.ts:generateInstanceEnvFromYamlConfig` method was updated so environment variables with spaces are surrounded by quotes. 
  - Before: `some_spaced_field=some spaced value`
  - After: `some_spaced_field="some spaced value"`
* The `components.ts` configmgr library was modified to lazily load config.
  - This changes where some log messages related to creating `/tmp/zweenv-yyyy` appear in the console output
  - This was necessary for updating zowe yaml regardless of validity. Ansible automation creates invalid zowe.yaml and runs sequential updates to fix it. The default `components.ts` load would always validate against schema and fail the update command, preventing the change that would fix the YAML.

Other: 
* `zwe install` - noted as a medium priority in the issue, this looks complete between `read_yaml_configmgr` updates and #4282
* Containers will require additional testing. Efforts were made to ensure compatibility remained in-place for them, however, the lack of test suites leaves room for errors to be introduced.


bin/libs/zosmf.ts
```diff
@@ -27,20 +27,16 @@ export function validateZosmfHostAndPort(zosmfHost: string, zosmfPort: number):
...
-    const execReturn = shell.execOutSync(`${std.getenv('NODE_HOME')}/bin/node`, `${std.getenv('ZWE_zowe_runtimeDirectory')}/bin/utils/curl.js`, `https://${zosmfHost}:${zosmfPort}/zosmf/info`, `-k`, `-H`, `X-CSRF-ZOSMF-HEADER: true`, `--response-type`, `status`);
-    if (execReturn.rc || !execReturn.out) {
-      common.printError(`Warning: Could not validate if z/OS MF is available on 'https://${zosmfHost}:${zosmfPort}/zosmf/info'. No response code from z/OSMF server.`);
-      zosmfCheckPassed=false
-    } else if (execReturn.out != '200') {
-      common.printError(`Could not contact z/OS MF on 'https://${zosmfHost}:${zosmfPort}/zosmf/info' - ${execReturn.out}`);
-      zosmfCheckPassed=false
-      return false;
-    }
+
```
Martin-Zeithaml, line 27: We should include #4324 into this. For some reason, the z/OSMF setting was taken from environment variables instead of configuration.

MarkAckert: 👍 merged that PR into this one

bin/utils/ModifyZoweYaml.ts
```diff
@@ -0,0 +1,79 @@
+/*
```
Martin-Zeithaml, line 1: This test script
```shell
#!/bin/sh

YAML='./zowe.yaml'
printf "zowe:\n  setup:\n    dataset:\n      prefix: TEST\n" > "${YAML}"

# Update ZWE_zowe_runtimeDirectory
export ZWE_zowe_runtimeDirectory='/zowe/pr4307'
export ZWE_CLI_PARAMETER_CONFIG="${YAML}"
export ZWE_PRIVATE_LOG_LEVEL_ZWELS="TRACE"
. "${ZWE_zowe_runtimeDirectory}/bin/libs/index.sh"

security_update="zowe.setup.security.product TSS"
update_zowe_yaml "${ZWE_CLI_PARAMETER_CONFIG}" $security_update
update_zowe_yaml "${ZWE_CLI_PARAMETER_CONFIG}" "zowe.setup.jcl.header" "123456,\n//  NOTIFY=ME"
update_zowe_yaml "${ZWE_CLI_PARAMETER_CONFIG}" "zowe.environments.var1" "~"
update_zowe_yaml "${ZWE_CLI_PARAMETER_CONFIG}" "zowe.environments.var1b" '~'
update_zowe_yaml "${ZWE_CLI_PARAMETER_CONFIG}" "zowe.environments.var2" "null"
update_zowe_yaml "${ZWE_CLI_PARAMETER_CONFIG}" "zowe.environments.var2b" 'null'
update_zowe_yaml "${ZWE_CLI_PARAMETER_CONFIG}" "zowe.environments.var3" "true"
update_zowe_yaml "${ZWE_CLI_PARAMETER_CONFIG}" "zowe.environments.var4" "false"
update_zowe_yaml "${ZWE_CLI_PARAMETER_CONFIG}" "zowe.environments.var5" ""
update_zowe_yaml "${ZWE_CLI_PARAMETER_CONFIG}" "zowe.environments.var6" ''
update_zowe_yaml "${ZWE_CLI_PARAMETER_CONFIG}" "node.home" 'C:\Program Files\Node'
update_zowe_yaml "${ZWE_CLI_PARAMETER_CONFIG}" "zOSMF.port" "11111"
update_zowe_yaml "${ZWE_CLI_PARAMETER_CONFIG}" "zOSMF.host" "zosmf.example.com"
update_zowe_yaml "${ZWE_CLI_PARAMETER_CONFIG}" "components.app-server.enabled" "true"

cat "${YAML}"

```
will produce this yaml:
```yaml
---
zowe:
  setup:
    dataset:
      prefix: "TEST"
    security:
      product: "TSS"
    jcl:
      header: "123456,\\n//  NOTIFY=ME"
  environments:
    var1: "~"
    var1b: "~"
    var2: "null"
    var2b: "null"
    var3: true
    var4: false
    var5: "\"\""
    var6: "\"\""
node:
  home: "C:\\Program Files\\Node"
zOSMF:
  port: 11111
  host: "zosmf.example.com"
components:
  app-server:
    enabled: true
...
```

There are some limitation:
* Schema undefined keys (e.g. `zowe.environments.*`) is treated as string, null character `~` is updated with `"~"`
* New line character is escaped
* Empty characters are also escaped

This routine is used by shell code and the number of calls of `update_zowe_yaml` is quite limited, the only concern is the empty string behavior. Otherwise the rest could be just mentioned in comment.

bin/libs/config.ts
```diff
@@ -135,44 +152,44 @@ export function generateInstanceEnvFromYamlConfig(haInstance: string) {
...
     });
 
-    componentFileArray = componentFileArray.map((row)=> { return row.endsWith('=null') ? row.substring(0, row.length-5)+'=' : row });
+    componentFileArray = componentFileArray.map((row) => { return row.endsWith('=null') ? row.substring(0, row.length - 5) + '=' : row });
     const componentFileContent = componentFileArray.join('\n');
     rc = xplatform.storeFileUTF8(`${folderName}/.instance-${haInstance}.env`, xplatform.AUTO_DETECT, componentFileContent);
-    if (rc) { 
-      common.printFormattedError("ZWELS", "bin/libs/config.ts,generateInstanceEnvFromYamlConfig", `ZWEL0140E: Failed to translate Zowe configuration (${cliParameterConfig}).`);
+    if (rc) {
+      common.printFormattedError("ZWELS", "bin/libs/config.ts,generateInstanceEnvFromYamlConfig", `ZWEL0140E: 1Failed to translate Zowe configuration (${cliParameterConfig}).`);
```
Martin-Zeithaml: `1Failed` -> `Failed`
Looks like the number was used to differentiate the same error message for various condition.
L230 and L238.
It would be better to include explanation into the error message.

bin/libs/zosmf.sh
```diff
@@ -29,21 +29,16 @@ validate_zosmf_host_and_port() {
 
   zosmf_check_passed=true
 
-  # SH: Note - if node is not available then will continue with a warning
-  if [ -z "${NODE_HOME}" ]; then
-    print_error "Warning: Could not validate if z/OS MF is available on 'https://${zosmf_host}:${zosmf_port}/zosmf/info'. NODE_HOME is not defined."
+  http_response_code=$("${ZWE_zowe_runtimeDirectory}/bin/utils/curl" "https://${zosmf_host}:${zosmf_port}/zosmf/info" -k -H "X-CSRF-ZOSMF-HEADER: true" --response-type status)
```
Martin-Zeithaml: `--response-type status` was an option for `curl.js`

This will return 
```
curl: option --response-type: is unknown
curl: try 'curl --help' or 'curl --manual' for more information
```

Martin-Zeithaml: `-I` could be used:

```
/zowe/bin/utils/curl "https://example.com:1234/zosmf/info" -k -H "X-CSRF-ZOSMF-HEADER: true" -I
HTTP/1.1 200 OK
X-Powered-By: Servlet/3.1
X-Content-Type-Options: nosniff
X-XSS-Protection: 1; mode=block
Content-Type: application/json
Content-Length: 239
Content-Language: en
Strict-Transport-Security: max-age=31536000; includeSubDomains
Date: Mon, 03 Nov 2025 15:39:25 GMT
```

bin/libs/zosmf.ts
```diff
@@ -27,20 +27,16 @@ export function validateZosmfHostAndPort(zosmfHost: string, zosmfPort: number):
...
+  const execReturn = shell.execOutSync(`${std.getenv('ZWE_zowe_runtimeDirectory')}/bin/utils/curl`, `https://${zosmfHost}:${zosmfPort}/zosmf/info`, `-k`, `-H`, `X-CSRF-ZOSMF-HEADER: true`, `--response-type`, `status`);
```
Martin-Zeithaml: Same as above, `--response-type` does not exist.

bin/libs/config.ts
MarkAckert: good catch, i removed the debug numbering. can think about adding additional context to the messages another time

bin/libs/zosmf.sh
MarkAckert: address in commit, using `-w %{http_code}` will output the response status code in the console's stdout

bin/commands/init/index.sh
```diff
@@ -57,7 +57,7 @@ fi
 # zowe.runtimeDirectory
```
Martin-Zeithaml, line 57: This is probably overthinking, but:
* I don't have `node` installed (other user with sufficient permissions will do it)
* I would like install and config full Zowe (Desktop included)
* I can't run `zwe init` as the node is required due to app-server is enabled

bin/utils/ModifyZoweYaml.ts
MarkAckert: The empty string behavior is fixed in `ModifyZoweYaml`. Special characters and strings which do not want to be escaped cannot use `update_zowe_yaml`, and we'll doc that.

bin/commands/init/index.sh
MarkAckert: I'll open an issue where we can discuss this further. For now, I think it's OK to fail `zwe init` in this scenario even if we don't need to fail - the user is just going to hit the same node problem later.

Martin-Zeithaml:
```shell
    security_update="zowe.setup.security.product TSS"
    update_zowe_yaml "${ZWE_CLI_PARAMETER_CONFIG}" $security_update
    update_zowe_yaml "${ZWE_CLI_PARAMETER_CONFIG}" "zowe.setup.jcl.header" "123456,\n//  NOTIFY=ME"
    update_zowe_yaml "${ZWE_CLI_PARAMETER_CONFIG}" "zowe.environments.var1" "~"
    update_zowe_yaml "${ZWE_CLI_PARAMETER_CONFIG}" "zowe.environments.var1b" '~'
    update_zowe_yaml "${ZWE_CLI_PARAMETER_CONFIG}" "zowe.environments.var2" "null"
    update_zowe_yaml "${ZWE_CLI_PARAMETER_CONFIG}" "zowe.environments.var2b" 'null'
    update_zowe_yaml "${ZWE_CLI_PARAMETER_CONFIG}" "zowe.environments.var3" "true"
    update_zowe_yaml "${ZWE_CLI_PARAMETER_CONFIG}" "zowe.environments.var4" "false"
    update_zowe_yaml "${ZWE_CLI_PARAMETER_CONFIG}" "zowe.environments.var5" ""
    update_zowe_yaml "${ZWE_CLI_PARAMETER_CONFIG}" "zowe.environments.var6" ''
    update_zowe_yaml "${ZWE_CLI_PARAMETER_CONFIG}" "node.home" 'C:\Program Files\Node'
    update_zowe_yaml "${ZWE_CLI_PARAMETER_CONFIG}" "zOSMF.port" "11111"
    update_zowe_yaml "${ZWE_CLI_PARAMETER_CONFIG}" "zOSMF.host" "zosmf.example.com"
    update_zowe_yaml "${ZWE_CLI_PARAMETER_CONFIG}" "components.app-server.enabled" "true"
```
```yaml
---
# <-- Initial yaml config
zowe:
  setup:
    dataset:
      prefix: "TEST"
# -->
    security:
      product: "TSS"
    jcl:
      header: "123456,\\n//  NOTIFY=ME"
  environments:
    var1: "~"
    var1b: "~"
    var2: "null"
    var2b: "null"
    var3: true
    var4: false
    var5: ""
    var6: ""
node:
  home: "C:\\Program Files\\Node"
zOSMF:
  port: 11111
  host: "zosmf.example.com"
components:
  app-server:
    enabled: true
...
```

bin/commands/internal/start/prepare/index.ts
```diff
@@ -459,7 +458,7 @@ export function execute() {
     // other extensions need to specify `require_java` in their validate.sh
     java.requireJava();
   }
-  if (stringlib.itemInList('app-server', std.getenv('ZWE_CLI_PARAMETER_COMPONENT'))) {
+  if (stringlib.itemInList(std.getenv('ZWE_PRIVATE_CORE_COMPONENTS_REQUIRE_NODE'), std.getenv('ZWE_CLI_PARAMETER_COMPONENT'))) {
```
Martin-Zeithaml, line 461: This will work only because there is just `app-server` defined. If the list of the components is extended, `itemInList` uses the first parameter to search, but the second parameter is split by `,`.

MarkAckert: Double checked, I'm not sure I follow. The first parameter is the list (previously hard-coded), second is the search term, and the function splits the list by `,`. This should work if we extend `ZWE_PRIVATE_CORE_COMPONENTS_REQUIRE_NODE`.

Martin-Zeithaml: I probably did not read the code correctly, but anyway, I will wrote a small unit test to be sure.

Martin-Zeithaml: Working fine, sorry for confusion. The `ZWE_CLI_PARAMETER_COMPONENT` is always one component. At least it was double checked.

Martin-Zeithaml:
Following configuration set with `zowe.setup.jcl.enable=true`.
Looks like the update is always expecting a file.
```
$ /4307/bin/zwe init security --config "PARMLIB(ZOWE.PR#4307.YAML(YAML)) --update-config"
...
ZWEL0318E - failed to read 'PARMLIB(ZOWE.PR#4307.YAML(YAML))' - EDC5129I No such file or directory. (errno2=0x05620062)
- update zowe config PARMLIB(ZOWE.PR#4307.YAML(YAML)), key: ".zowe.setup.security.product" with value: TSS, and validate: true
  * Success
Error: Could not load config for FILE(PARMLIB(ZOWE.PR#4307.YAML(YAML))), status=7
```

Martin-Zeithaml (approved):
## `curl` replacement
There were some issues with incompatible flags, all resolved and tested.

## `read_yaml`
Using configmgr to read the value of a key. Understands the embedded JS and comments. No issues found.

##  `bin/utils/GenerateInstanceEnv.ts`
Just a small test written, seems to be working fine.

## `bin/utils/ModifyZoweYaml.ts`
Functionality to modify the config. Tested via small shell script, there are some limitations, however should not be exploitable by users.

## Update PARMLIB entry only
A problem was found, when only PARMLIB config was used. Fixed and seems to be working fine now.

## General behavior
User is not able to continue without `node`, if `app-server` is enabled. Discussed and will be resolve in other PR (low priority).

Approved.

MarkAckert:
Build failure is unrelated to these changes, merging in.

PR #4312, opened 2025-05-05
WIP: Removing excessive outputs for zwe init certificate command

skurnevich:
Fixes [4308](https://github.com/zowe/zowe-install-packaging/issues/4308)

WIP: Review for more optimal solution

Issue #4308, opened 2025-05-02
zwe init certificate fails if user has no access to workspace directory


skurnevich:
**Describe the bug**
zwe init certificate command fails if user has no access to the workspace directory

**To Reproduce**
Tested on scenario 1
- Run zwe init certificate command by user who has no access to the Zowe workspace directory
- See `directoryExists path=<zowe-workspace-path>/.env, err=111` in the output
- `>>>> Sign the CSR using the Certificate Authority "local_ca"` step will fail because the directoryExists error becomes a value in SAN
`SAN=dns:>>>,dns:tmpdir=undefined,dns:error:,dns:directoryExists,dns:path=<zowe-workspace-path>/.env,dns:err=111,dns:localhost.localdomain,dns:localhost,ip:127.0.0.1 `

Shortened log:
```
Zowe server command: zwe init certificate
- timestamp: 2025-04-29 11:16:52
- parameters:
  * config: /a/workdir/zowe-v3.yaml
  * trace: true
  * log-dir: /a/workdir/log

<...>
    
- read_yaml_configmgr process /a/workdir/zowe-v3.yaml and extract '.zowe.setup.certificate.pkcs12.import.alias -> /zowe/setup/certificate/pkcs12/import/alias'
  * Exit code: 16
  * Output:
    
directoryExists path=/a/workdir/zowe-workspace/.env, err=111

directoryExists path=/a/workdir/zowe-workspace/.env, err=111

<...>

>>>> Sign the CSR using the Certificate Authority "local_ca":
- Calling keytool -J-Dkeystore.pkcs12.legacy -gencert -v -infile /a/workdir/keystore/localhost/localhost.csr -outfile /a/workdir/keystore/localhost/localhost.signed.cer -keystore /a/workdir/keystore/local_ca/local_ca.keystore.p12 -alias local_ca -keypass local_ca_password -storepass local_ca_password -storetype PKCS12 -ext SAN=dns:>>>,dns:tmpdir=undefined,dns:error:,dns:directoryExists,dns:path=/a/workdir/zowe-workspace/.env,dns:err=111,dns:localhost.localdomain,dns:localhost,ip:127.0.0.1 -ext KeyUsage:critical=keyEncipherment,digitalSignature,nonRepudiation,dataEncipherment -ext ExtendedKeyUsage=clientAuth,serverAuth -rfc -validity 3650
  * keytool failed
  * Exit code: 1
  * Output:
[...]

PR #4319, opened 2025-05-06
WIP: Switch to read_yaml_configmgr for cert CAs and domains extraction

skurnevich:
Addressing [4308](https://github.com/zowe/zowe-install-packaging/issues/4308) 
Alternative solution to #4312, instead of using internal zwe command that requires workspace directory access, use read_yaml_configmgr that now can handle arrays, thanks to : https://github.com/zowe/zowe-common-c/pull/502


Martin-Zeithaml (approved):

1000TurquoisePogs (approved):
This looks like a cleaner, faster approach!

PR #4322, opened 2025-05-09
move zwe version to configmgr logic

1000TurquoisePogs:
This PR rewrites the `zwe version` command as a configmgr command, removing the shell code.
The shell code was also looking for `manifest.json.template`, but that seems like unused code, so i removed it and only check for `manifest.json`.

## Update

`zwe version`
* Without `--config`: prints the Zowe version, where the current `zwe` command is located
* With `--config`: reads the `zowe.runtimeDirectory` and prints the Zowe version from there
  * `--config` is Unix file or `FILE() | PARMLIB()` entries.

### Additional checks
New error `ZWEL0327E` - error parsing (e.g. wrong encoding) or reading (e.g. missing properties) JSON file.


bin/commands/version/index.ts
```diff
@@ -0,0 +1,30 @@
...
+ 
+  Copyright Contributors to the Zowe Project.
+*/
+
+import * as xplatform from 'xplatform';
+import * as common from '../../libs/common';
+import * as config from '../../libs/config';
+
+export function execute() {
+  common.requireZoweYaml();
```
Martin-Zeithaml: The current `zwe version` is working without `--config` and prints the version of Zowe where the `zwe` is located.

This new version requires `--config` (which is a breaking change, but acceptable for this type of command), because it can print any Zowe version - it will use `zowe.runtimeDirectory` to locate manifest/version.

If possible, I would like to have this behavior:
* If no config, return `ZWE_zowe_runtimeDirectory/manifest.json -> version`
  * This is very handy, if you (as a developer) have multiple versions and you can easily get the version without config
* If config, return whatever is defined at the end of `zowe.runtimeDirectory/manifest.json -> version`

`.examples` - new example with config
`.help` - if we implement config yes/no behavior, couple words here

1000TurquoisePogs: oh - so there was a reason we didnt do this :)

Martin-Zeithaml: This is ugly, but works with/without config:

```javascript
import * as std from 'cm_std';
import * as common from '../../libs/common';
import * as fs from '../../libs/fs';
import * as shell from '../../libs/shell';
import * as xplatform from 'xplatform';

export function execute() {

    const cliConfig = std.getenv('ZWE_CLI_PARAMETER_CONFIG');
    const currentRuntimeDirectory = std.getenv('ZWE_zowe_runtimeDirectory');
    let runtimeDirectory = currentRuntimeDirectory;

    if (cliConfig) {
      const configmgr = `${currentRuntimeDirectory}/bin/utils/configmgr`;
      const schema = `${currentRuntimeDirectory}/schemas/server-common.json:${currentRuntimeDirectory}/schemas/zowe-yaml-schema.json`;
      let file = cliConfig;
      if (fs.fileExists(file)) {
        // FILE is required for configmgr extract for a single file
        file = `FILE(${cliConfig})`;
      }
      const result = shell.execOutSync('sh', '-c', `_CEE_RUNOPTS="XPLINK(ON)" "${configmgr}" -s "$schema" -p "${file}" extract "/zowe/runtimeDirectory" 2>&1`);
      if (result.rc == 0 && result.out) {
        runtimeDirectory = result.out;
      } else {
        common.printErrorAndExit("Error ZWEL0150E: Failed to find Zowe manifest.json. Zowe runtimeDirectory is invalid.", undefined, 150);
      }
    }

    const manifestPath = `${runtimeDirectory}/manifest.json`;
    if (!fs.fileExists(manifestPath)) {
      common.printErrorAndExit("Error ZWEL0150E: Failed to find Zowe manifest.json. Zowe runtimeDirectory is invalid.", undefined, 150);
    }

    const contents = xplatform.loadFileUTF8(manifestPath, xplatform.AUTO_DETECT);
    const manifest = JSON.parse(contents);
    common.printMessage("Zowe v" + manifest.version);
    common.printDebug(`build and hash: ${manifest.build.branch}#${manifest.build.number} (${manifest.build.commitHash})`);
    common.printTrace(`Zowe directory: ${runtimeDirectory}`);

}

```

MarkAckert (approved):

PR #4332, opened 2025-05-16
Update ZWEGENER to support different members

Martin-Zeithaml:
Missing in #4285:

* `ZWEGEN00`
  * Do not add `ZWEYAML` when creating `CONFIG` for `ZWESLSTC`
  * No need to set `ZWEYAML` as a member name
* `ZWEGENER`
  * Small comment update

`ZWEGEN00` is then set with the names as needed, for example:
```jcl
//MYCONFIG DD   *,DLM=$$                  
PARMLIB ZOWE.PR#4107.TMP(DEFAULT) 
PARMLIB ZOWE.PR#4107.TMP(NODE)    
$$                                        
```

Started task `ZWESLSTC` with generated `CONFIG` then:
```jcl
//STDENV   DD  *                                    
_CEE_ENVFILE_CONTINUATION=\                         
_CEE_RUNOPTS=HEAPPOOLS(OFF),HEAPPOOLS64(OFF)        
_EDC_UMASK_DFLT=0002                                
CONFIG=PARMLIB(ZOWE.PR#4107.TMP(DEFAULT)):\ 
PARMLIB(ZOWE.PR#4107.TMP(NODE))             
/*
```

PR #4334, opened 2025-05-19
Add CHANGELOG workflow automation

MarkAckert:
In light of PRs like #4332, we should have an automated check which points our attention to the CHANGELOG for all pull requests against staging branches. This automated check will be mandatory moving forward, and it can be bypassed by adding a `skip-changelog` label to the pull request.


Martin-Zeithaml (approved):
Approved. Quick test [here](https://github.com/zowe/zowe-install-packaging/pull/4336).

PR #4337, opened 2025-05-20
ZWEGEN00: check `zowe.setup.dataset.prefix` and `zowe.setup.dataset.jcllib`

Martin-Zeithaml:
Changes:
* Checks `zowe.setup.dataset.prefix` and `zowe.setup.dataset.jcllib`
* When `ZWEGENER` submitted manually, missing `zowe.setup.dataset.prefix` could cause some problems:
    * Attempt to create `CFG.ZOWE.SETUP.DATASET.PREFIX.JCLLIB` dataset, because by default a REXX variable is initialized with own name, which is unfortunately a valid dataset name
    * If user is not authorized to use HLQ `CFG`, there is just another error
    * The `jcllib` must not be set to `zowe.setup.dataset.prefix` + `.SZWESAMP`
* New function `FormatError` and `CheckInput`
* Space increased from 50 to 60

# Tests
## Input parameters

* Leading and trailing spaces removed
* Case sensitive
* No abbreviation
* Check via `CheckInput`, which could be modified to support abbreviation or case sensitivity

### Incorrect `generate` option
```jcl
//SYSTSIN  DD   *
ISPSTART CMD(%ZWEGEN00    Generate   noverbose )
```

```
Error: "Generate" is not a valid operation.
       Valid operations are: generate nogenerate
```

### Incorrect `verbose` option
```jcl
//SYSTSIN  DD   *
ISPSTART CMD(%ZWEGEN00 -
  generate   -
  verb  -
)
```
```
Error: "verb" is not a valid verbosity.
       Valid verbosity levels are: verbose noverbose
```
### Missing `zowe.setup.dataset.prefix`
```
Error: "zowe.setup.dataset.prefix" is not defined in configuration.  
       Update the configuration and submit the JCL again.            
```

### Missing `zowe.setup.dataset.jcllib`
```
Error: "zowe.setup.dataset.jcllib" is not defined in configuration.
       Update the configuration and submit the JCL again.          
```

### `jcllib` same as `prefix` + `.SZWESAMP`
```
Error: "zowe.setup.dataset.prefix" and "zowe.setup.dataset.jcllib" are identical, which is not allowed.
       Update the configuration and submit the JCL again.                                              
```

Otherwise, without fix:
```
Creating a fresh copy of ZOWE.PR4337.SZWESAMP named ZOWE.PR4337.SZWESAMP.         
  ISRLS358Invalid data sets       -/-Same data sets not allowed when processing 'LMCOPY' service. 
```

### DIFF
--- a/files/SZWEEXEC/ZWEGEN00
+++ b/files/SZWEEXEC/ZWEGEN00
@@ -406,7 +406,9 @@ say 'The edit macro was invoked on each member.'
 ================================================================================
     Add the job card to each member if filled out.
 ================================================================================
-*/
+FIXME:
+  1. Card is added to each member (STCs included)
+  2. It overwrites the first line
 
 card.0 = 0
 
@@ -450,7 +452,7 @@ if card.0 > 0 then do
   say 'The job card was added to each member.'
 end
 
-
+*/
 
 say 'Checking the ZWESLSTC CONFIG entry'
 
--- a/files/SZWEEXEC/ZWEGEN00
+++ b/files/SZWEEXEC/ZWEGEN00
@@ -27,10 +27,11 @@ parse arg operation verbosity
 operation = strip(operation)
 verbosity = strip(verbosity)
 
-VALID_OPERATIONS = '"generate" | "nogenerate"'
-VALID_VERBOSITY = '"verbose" | "noverbose"'
+VALID_OPERATIONS = 'generate nogenerate'
+VALID_VERBOSITY = 'verbose noverbose'
 
-if POS(verbosity, VALID_VERBOSITY) = 0 then do
+checkVerbosity = CheckInput(verbosity, VALID_VERBOSITY)
+if length(checkVerbosity) = 0 then do
   queue '"'verbosity'" is not a valid verbosity.'
   queue 'Valid verbosity levels are: '||VALID_VERBOSITY
   FormatError(8)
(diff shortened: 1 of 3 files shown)


files/SZWEEXEC/ZWEGEN00
```diff
@@ -384,7 +406,9 @@ say 'The edit macro was invoked on each member.'
 ================================================================================
     Add the job card to each member if filled out.
 ================================================================================
-*/
+FIXME:
```
MarkAckert, line 409: For a future PR?

Martin-Zeithaml: It depends, if someone has strong opinion, that the current implementation in "Migrate to JCL" is wrong, then we will have to fix this. Otherwise we can postpone this to another PR.

MarkAckert: The jobcard support is required for migrate to jcl, so we can merge this in first and then address it in that PR. It's easier to test there anyway. Does that work?

Martin-Zeithaml: Yes, sounds like a plan. And we should probably reconsider where to place the job card. This code supposes `zowe.setup.jobCard`, which is better than `zowe.environment`. It will require just small schema update.

MarkAckert (approved):

MarkAckert (approved):

PR #4349, opened 2025-05-23
Add default sysmessages

1000TurquoisePogs:
Prior, sysmessages we recommend were only in example-zowe.yaml
This adds them to defaults.yaml so that users of old versions of zowe will benefit from this new default behavior.

Also, this PR adds additional sysmessages that seem useful to see in the syslog, about launcher queries.

### DIFF
--- a/files/defaults.yaml
+++ b/files/defaults.yaml
@@ -142,6 +142,33 @@ zowe:
   # >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
   # To determine whether to trim ths syslog message inorder to print from message-id, disabled by default
   sysMessageTrim: false
+  sysMessages:
+  #   # Zowe starting
+    - "ZWEL0021I"
+  #   # Zowe started
+    - "ZWEL0018I"
+    - "ZWEL0006I"
+  #   # Zowe ready to use
+    - "ZWES1601I"
+  #   # Zowe stopping
+    - "ZWEL0008I"
+  #   # Zowe stopped
+    - "ZWEL0022I"
+  #   # Zowe components starting
+    - "ZWEL0001I"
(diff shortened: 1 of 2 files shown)


files/defaults.yaml
```diff
@@ -142,6 +142,33 @@ zowe:
   # >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
   # To determine whether to trim ths syslog message inorder to print from message-id, disabled by default
   sysMessageTrim: false
+  sysMessages:
```
Martin-Zeithaml, line 145: When creating merged config, the items in **user's config and defaults must be unique**, otherwise there is an error like this:
```yaml
# User defined config
zowe:
  sysMessages:
    - ZWEL0021I
```

```
Error: Validation of FILE(test4349.yaml):FILE(/git/pr/4349/files/defaults.yaml) against schema
/git/pr/4349/schemas/zowe-yaml-schema.json:/git/pr/4349/schemas/server-common.json found invalid JSON Schema data
    Validity Exceptions(s) with array at /zowe/sysMessages
      array uniqueItems violation, duplicate at /zowe/sysMessages/15
```
This is super confusing from the user point of view: just one value was specified and it is not unique.

In other words:
* user can only add new unique messages
* user can't specify any subset of the defaults messages
* user can't remove any message ID defined here (unless the defaults.yaml is modified)

Martin-Zeithaml (changes requested):
I think we need to reconsider this.

1000TurquoisePogs:
You're right. I forgot the reason why we did not do this before.

Martin-Zeithaml:
But the change for `example-zowe.yaml` only is still valid, isn't? We can remove `defaults` and continue.

PR #4359, opened 2025-06-04
Migrate to JCL: jcl header in `ZWEGEN00`

Martin-Zeithaml:
When the JCL `JOB` statement needs to be customized, user can define required keywords via `zowe.setup.jcl.header`. The format of this key is string/number or an array:

One line, for example accounting information only:
```yaml
zowe:
  setup:
    jcl:
      header: 123456
```

Multiple lines as an array:
```yaml
zowe:
  setup:
    jcl:
      header:
        - "123456,"
        - "//       'Zowe User',"
        - "//       NOTIFY=&SYSUID"
```
JCL samples are in format of `//ZWEnnnnn JOB {zowe.setup.jcl.header}`, where `nnnnn` is different for each job.

The `default.yaml` contains `zowe.setup.jcl.header=''` for sites, where no modification is fine.

## Schema changes
New `zowe.setup.jcl.header` could be
* Interger, not null
* String, not null, max length 65 (80 - length(`//ZWEnnnnn JOB `))
* Array, each item string, but not null, at least 2 items (1 line via string)
  * Is it possible to validate `item[0]` for length 65 and the rest for length 80?

## To do list
* zwe remote tests update
* PSWI workflow update

## Done
* `ZWEGEN00` update - always process the header here
  * For the simplification, the `queue` function was used. No need to count lines etc.
  * Variable `jcl` was renamed to `_jcl_`. When used in new compound variable `CFG.zowe.setup.jcl.header`, it was evaluated with the value of `jcllib`.
* Update `zwe init generate` to change job header only
* Schema update - to support new `setup` key
  * `zowe.setup.jcl.header` could be number, string or array
  * String is checked for 65 chars to fit into the `JOB` statement
  * Array is checked for 80 chars (which is not ok for the first line, but ok for the rest)
  * If array is used, at least 2 items must be defined


MarkAckert:
For schema validation on items[0] vs items[1...n], this below block is valid JSON Schema which does just that. I tested this with [ajv 2019](https://github.com/ajv-validator/ajv). Unfortunately, it's not supported by configmgr's schema validation, as zowe-common-c expects `items` to be an object, and not an array. When `items` are an array, each item in the schema is used to validate matching items in the array, with the last schema item being repeated for the `....n` case when the array has more values than schema validator entries. The error thrown by configmgr [happens here](https://github.com/zowe/zowe-common-c/blob/5d4f38341119b5e1406fd4ac110c6c74b7472382/c/jsonschema.c#L2010), because `build` does not support array types. We would need to fix that and update the validator to assume `items` is an array instead of a single object.

```json
{
                      "type": "array",
                      "minItems": 1,
                      "maxItems": 10,
                      "items": [
                        {
                          "type": "string",
                          "not": {
                            "type": "null"
                          },
                          "maxLength": 65
                        },
                        {
                          "type": "string",
                          "not": {
                            "type": "null"
                          },
                          "maxLength": 80
                        }
                      ],
                      "description": "Multiline JCL header options for samples in SZWESAMP: '//ZWEnnnnn JOB {zowe.setup.jcl.header}'."
                    }
```

MarkAckert:
I also updated the `zwe` integration tests - they're looking good on my system. I updated the zowe yaml schema to re-introduce restrains on sysMessages that were lost on merge.

MarkAckert (approved):

MarkAckert:
Manually reviewed the failing test and confirmed its passing; I'll look into automation updates in another PR. The test failure is unrelated to all changes in this PR and the workflow JCL header update PR, where it first popped up.

Issue #4309, opened 2025-05-02
Research - ZWECONF and ZWEAMLCF sync with zowe.yaml


MarkAckert:
Is there any way to simplify the sync between PSWI workflows on zowe.yaml fields? We need to answer:

1. Do we need a duplicate copy of the zowe.yaml in the workflow configuration, or are there other ways to create the working zowe.yaml using `example-zowe.yaml` as a base?

2. How many of the zowe.yaml options need to be configurable via PSWI? Is it the bare minimum to get someone up and running followed by post-workflow customization, or do we want to cover every single possible configuration item?

3. How do we track variables defined in zowe.yaml vs those defined and configurable in the PSWI workflow? With today's sync automation, we can keep the zowe.yaml in check and compatible, but we may not be offering users newer zowe.yaml options that are available, or removing old zowe.yaml options that no longer apply.

cc @muzikovam @jp669844 @1000TurquoisePogs

MarkAckert:
addressed in #4366

PR #4366, opened 2025-06-11
PSWI zowe.yaml sync automation improvements

MarkAckert:
Addresses #4309 , points 2 and 3 keeping workflow variables in sync with `example-zowe.yaml` and `defaults.yaml`. The github action `PSWI workflow sync check` was enhanced to run two more scripts, which run extra validations as laid out below.

***IMPORTANT***: This PR changes variables in both workflows so the variable names now **exactly match** those defined in the zowe.yaml file. The naming structure is used to match definitions in workflows back to zowe.yaml fields.

For both ZWEAMLCF.xml and ZWECONF.xml:
* Check variables defined in the workflows which **do not map back** to any variables present in zowe.yaml
* Check variables defined in example-zowe.yaml and defaults.yaml **which are not defined** in the workflows
* Uses a new file `workflows/zowe-yaml-sync-rules.json` to explicitly list exceptions where missing variables as defined above are expected and allowed.
* Wrote an additional script that covers different rendering possibilities for the workflow based on user input. The schema validation in this script does not use configmgr, so there's room for disparity between test output and real-world use, though in practice this shouldn't happen.

Todo:

Questions and notes for reviewers:
* What, if anything, can we do about point 1 in #4309 ? My impression is not much: at best, there is a generator that copies `example-zowe.yaml` into the workflow body at build-time given workflow-specific considerations, but this too would need to be maintained and may be more work than managing the current duplications. 
* Ask workflow authors to review missing `zowe_configmgr_validation` in `ZWEAMLCF` workflow. We have a default instead, and should the default really be `COMPONENT-COMPAT` instead of `STRICT`?
* Note: the workflow currently fails pending #4359. We could update the `zowe-yaml-sync-rules.json` and merge this in first if preferred, but we will need to change it again for 4359 regardless.


muzikovam (approved):
LGTM

muzikovam:
And to your questions:
1. I agree with you.
2. The new workflow should have as few variables as possible. And the configmgr validation is not essential for the user to fill in. COMPONENT-COMPAT is the default in the big workflow, and people from APIML believe that it can't hurt anything :)
3. I would, too, wait after the merge :)

Issue #4367, opened 2025-06-12
HA config lookup fails due to case sensitivity in haInstance key


struga0258:
In [configmgr.ts](https://github.com/zowe/zowe-install-packaging/blob/e32b89df310b33ce3942b8d17b0f559426622204/bin/libs/configmgr.ts#L505), there's an issue where the code checks config.haInstances[haInstance], but this fails if the haInstance name in the config file uses uppercase letters.

Problem:

Somewhere in the code, haInstance is set to lowercase. But in the config file, the keys under haInstances may be uppercase (e.g HARS27). Because the lookup is case sensitive, it fails and goes into the else block, which uses the wrong hostname from externalDomains[0]. This causes major issues with HA because the wrong server gets picked.

MarkAckert:
Fixes in #4378 and #4370

PR #4370, opened 2025-06-16
Case insensitive ha config lookup

1000TurquoisePogs:
@struga0258  [found](https://github.com/zowe/zowe-install-packaging/issues/4367) that ha instance config lookup is ignoring uppercase ha instance names.
For example, if you had 

```yaml
haInstances:
  FOO:
    sysname: abc
```

Zowe would report your ha instance as "foo", and then not read the "FOO" section.
This PR changes the behavior so that the "FOO" section is read, by doing a lowercase string match.


bin/libs/config.ts
```diff
@@ -183,8 +183,17 @@ export function generateInstanceEnvFromYamlConfig(haInstance: string) {
...
-    haConfig = merger.merge(jsonConfig.haInstances[haInstance], jsonConfig);
+  let haKeys = [];
+  if (jsonConfig.haInstances) { haKeys = Object.keys(jsonConfig.haInstances) }
+
+  if (haInstance && haKeys.length > 0) {
+    for (let i = 0; i < haKeys.length; i++) {
+      // make haInstance lookup case-insensitive by looking for first lowercase match
+      if (haKeys[i].toLowerCase() == haInstance) {
+        haConfig = merger.merge(jsonConfig.haInstances[haKeys[i]], jsonConfig);
+        break;
```
Martin-Zeithaml, line 194: We should probably add some doc note about the case insensitivity limitation, otherwise the `foo` is always inaccessible in this valid yaml example:

```yaml
haInstances:
  FOO:
    sysname: abc
  foo:
    sysname: xyz
```

https://docs.zowe.org/stable/appendix/zowe-yaml-configuration/#yaml-configurations---hainstances

Martin-Zeithaml (approved):

bin/libs/config.ts
MarkAckert: addressed in https://github.com/zowe/docs-site/pull/4551 👍

PR #4375, opened 2025-06-17
Update wokflows for JCL headers

Martin-Zeithaml:
If the JOB statement is 1 line, create
```yaml
  setup:
    jcl:
      header: input_from_user
```

For multiple lines
```yaml
  setup:
    jcl:
      header:
      - input_from_user_line1
      - //      input_from_user_line2
      - //      input_from_user_line3
```


MarkAckert:
PSWI and configuration workflows look good in my tests, merging in. I also added `zwe init generate` to the ZWEAMLCF.xml file which was missing.

PR #4381, opened 2025-06-25
Update the ZSS & ZIS description

Martin-Zeithaml:
ZSS and ZIS description updated according to Zowe Glossary.


JoeNemo (approved):
Approved.

PR #4383, opened 2025-06-30
Update `--dry-run` parameter description

Martin-Zeithaml:
Same as for other commands using `--dry-run` parameter.


JoeNemo (approved):
Approved.

Issue #4310, opened 2025-05-02
PSWI Install bug - zowe.yaml in runtimeDirectory


MarkAckert:
The current PSWI workflow outputs the `zowe.yaml` file to the `instance-zowe_runtimeDirectory` variable. The runtimeDirectory should be read-only, and the customized zowe.yaml file output instead to the `instance-zowe_workspaceDirectory` location. The workspace directory may need to be created by the workflow before the customization of the zowe.yaml file happens.

MarkAckert:
Completed in 3.4.0 with #4397

PR #4397, opened 2025-07-10
PSWI - Move zowe.yaml from `runtimeDirectory` to `workspaceDirectory`

MarkAckert:
Addresses #4310 .

I've manually run both the ZWEAMLCF workflow and ZWECONF workflows, and the file is created as expected in the `workspaceDirectory`.

There is no automated test I can write as far as I know


pavelzlatnik (approved):
Changes are ok.
There should be probably done small doc update as well, to mention that for workflows is now used yaml stored in this new workspace directory.
IN here https://docs.zowe.org/stable/user-guide/configure-zowe-zosmf-workflow#create-configuration   and in here https://docs.zowe.org/stable/user-guide/configure-apiml-zosmf-workflow#create-configuration

MarkAckert:
I think this will be deferred to 3.4 due to higher priority work on `zwe install` and `zwe init`, unless someone strongly argues this should make 3.3.

PR #4401, opened 2025-07-14
Missing JCL header

Martin-Zeithaml:
2 samples are missing `{zowe.setup.jcl.header}`.

### DIFF
--- a/files/SZWESAMP/ZWEKRING
+++ b/files/SZWESAMP/ZWEKRING
@@ -1,4 +1,4 @@
-//ZWEKRING JOB
+//ZWEKRING JOB {zowe.setup.jcl.header}
 //*
 //* This program and the accompanying materials are made available
 //* under the terms of the Eclipse Public License v2.0 which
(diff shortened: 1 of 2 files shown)


joao-krauchenco-mottu (approved):

MarkAckert:
I thought these samples were no longer used or maintained; we've only kept them around due to SMP/e. Maybe adding a comment like `DEPRECATED` in the JCL would be good?

Martin-Zeithaml:
> I thought these samples were no longer used or maintained; we've only kept them around due to SMP/e. Maybe adding a comment like `DEPRECATED` in the JCL would be good?

Both JCL samples are used by `zwe init generate` and placed into the `zowe.setup.dataset.jcllib`.
I think we have to do one of the following:
* Update the JCL header, so it could be used - even more manual customization is needed
* Exclude it from processing like some [other members](https://github.com/zowe/zowe-install-packaging/blob/dfad8d8cead8fb43e78abc79fb341fd9516fd979/files/SZWEEXEC/ZWEGEN00#L184-L190)

Anyway, comment deprecated is a good idea.

**Edit**

`zwe init generate --help`:
_Generate ready to execute JCL samples from zowe.yaml configuration values._

Probably the second option (exclude) is better due to the promising description of `generate` action.

MarkAckert:
Agree with option #2, exclude

PR #4410, opened 2025-07-17
Schema validation: allow null/empty strings as in Zowe 3.2.0

Martin-Zeithaml:
Schema changes in #4107 caused this is not valid configuration:
```yaml
zowe:
  setup:
    dataset:
      prefix: 
      parmlib: 
      jcllib: 
      authLoadlib: 
      authPluginLib: 
    security:
      groups:
        admin:
        stc:
        sysProg: 
      users:
        zowe:
        zis: 
      stcs:
        zowe: 
        zis: 
        aux: 
```

It is a braking change, schema must be updated to accept empty or null values for all keys in this example. The main reason is, that the startup process is checking entire configuration, included `zowe.setup.*`. Before the #4107 that example was fine for starting the launcher.

### DIFF
--- a/bin/commands/init/mvs/index.ts
+++ b/bin/commands/init/mvs/index.ts
@@ -27,9 +27,9 @@ export function execute(allowOverwrite?: boolean) {
     common.printErrorAndExit(`Error ZWEL0157E: Zowe dataset prefix (zowe.setup.dataset.prefix) is not defined in Zowe YAML configuration file.`, undefined, 157);
   }
 
-  const zssEnabled = ZOWE_CONFIG.components.zss.enabled;  // In defaults, this property is always defined
+  // For V4: if components.zss.enabled = false => no need to create parmlib
   const parmlib = ZOWE_CONFIG.zowe.setup?.dataset?.parmlib ? ZOWE_CONFIG.zowe.setup.dataset.parmlib : undefined;
-  if (zssEnabled && parmlib == undefined) {
+  if (!parmlib) {
       common.printErrorAndExit(`Error ZWEL0157E: zowe.setup.dataset.parmlib is not defined in Zowe YAML configuration file.`, undefined, 157);
   }
 
--- a/files/SZWEEXEC/ZWEGEN00
+++ b/files/SZWEEXEC/ZWEGEN00
@@ -279,6 +279,70 @@ say 'The 'output.0' members in 'jclCopy' have been recorded.'
 
 /*
 ================================================================================
+    Set defaults - user defined empty/null values will overwrite the defaults
+      defined in "zowe.runtimeDirectory/files/defaults.yaml".
+      It needs to be restored for such case.
+================================================================================
+*/
+
+say 'Setting the defaults values.'
+
+if isNullEmpty(CFG.zowe.setup.dataset.authLoadlib) then do
+  CFG.zowe.setup.dataset.authLoadlib = CFG.zowe.setup.dataset.prefix'.SZWEAUTH'
+  dbgTxt = 'zowe.setup.dataset.authLoadlib = '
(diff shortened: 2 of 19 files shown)


MarkAckert:
FYI [this commit](https://github.com/zowe/zowe-install-packaging/pull/4410/commits/b6673a3e7c179b950672b1b56897fad2d2799d49) is not intended to be final, I committed it for review - which behaviors are as expected, which aren't. There are still return codes that need to be updated in the tests before they'll pass.

bin/commands/init/mvs/index.ts
```diff
@@ -27,9 +27,9 @@ export function execute(allowOverwrite?: boolean) {
     common.printErrorAndExit(`Error ZWEL0157E: Zowe dataset prefix (zowe.setup.dataset.prefix) is not defined in Zowe YAML configuration file.`, undefined, 157);
   }
 
-  const zssEnabled = ZOWE_CONFIG.components.zss.enabled;  // In defaults, this property is always defined
+  // For V4: if components.zss.enabled = false => no need to create parmlib
   const parmlib = ZOWE_CONFIG.zowe.setup?.dataset?.parmlib ? ZOWE_CONFIG.zowe.setup.dataset.parmlib : undefined;
-  if (zssEnabled && parmlib == undefined) {
+  if (!parmlib) {
```
1000TurquoisePogs, line 32: why change this? i dont think we need parmlib if we dont have zis (which we assume based on zss)

files/SZWEEXEC/ZWEGEN00
```diff
@@ -273,6 +273,70 @@ end
...
+================================================================================
+    Set defaults - user defined empty/null values will overwrite the defaults
+      defined in "zowe.runtimeDirectory/files/defaults.yaml".
+      It needs to be restored for such case.
+================================================================================
+*/
+
+say 'Setting the defaults values.'
+
+if isNullEmpty(CFG.zowe.setup.dataset.authLoadlib) then do
```
1000TurquoisePogs, line 286: is this the only such dataset entry that needs defaulting?
tbh i dont really like SZWEAUTH & SZWELOAD are customizable in the yaml because they arent from an SMPE perspective so SMPE users can just create user error - not that we need to change that now.

bin/commands/init/mvs/index.ts
Martin-Zeithaml: The `parmlib` was required before (Zowe 3.2.0), so it is required now.

files/SZWEEXEC/ZWEGEN00
Martin-Zeithaml: This piece of code is consequence of allowing null/empty values. Without this change and with example from PR description, this would be generated for example for `ZWEISTC`:
```jcl
//STCZOWE DD DISP=SHR,    
// DSN=&PROCLIB.(<NULL>)  
//STCZIS DD DISP=SHR,     
// DSN=&PROCLIB.(<NULL>)  
//STCAUX DD DISP=SHR,     
// DSN=&PROCLIB.(<NULL>)  
```

We have to "restore" all defaults, which could be possibly "overwritten" by `null`/empty string.

Martin-Zeithaml: According to our help, this should be the only dataset with default value.

MarkAckert:
To summarize the changes from the prior behavior with stricter schema, based on tests:

* Certain commands now fail during the `init` command rather than as a schema issue
  - As a result, the return codes for `init` commands have changed in multiple places
* Certain commands / negative cases which were failing are now expected to be passing.
  - `init stc`, `init mvs`, `init apf`, `init vsam` all have cases where hard-coded defaults are used when zowe.yaml entries are blank or null

MarkAckert (approved):

PR #4412, opened 2025-07-18
adjust smp/e install to use a zowe.yaml for zwe install

MarkAckert:
The SMP/e build used `zwe install --dataset-prefix`, which was removed with #4282. This PR adjusts the SMP/e build so it creates a `zowe.yaml` file on the fly instead and removes it after to avoid polluting SMP/e installable media.


struga0258 (approved):

MarkAckert:
SMP/e manually tested and working with expected files. Error is unrelated; merging this in with failures to unblock other builds.

PR #4414, opened 2025-07-18
Add zis-test and bind-test to utilities folder

1000TurquoisePogs:
This PR adds the bind-test and zis-test utilities to the bin/utils folder.
They are not used by zowe-install-packaging directly, but will be in the future such as in https://github.com/zowe/zowe-install-packaging/pull/4276

zis-test is currently used in zss and thus needs to be put into the packaging now.

### DIFF
--- a/.github/workflows/build-packaging.yml
+++ b/.github/workflows/build-packaging.yml
@@ -276,7 +276,7 @@ jobs:
         with:
           manifest-file-path: ${{ github.workspace }}/manifest.json
           default-target-path: .pax/binaryDependencies/
-          expected-count: 28
+          expected-count: 30
       
       # this step is not doing a publish, we are just utilizing this actions to get the PUBLISH_TARGET_PATH, 
       # and it will be used in the next step: [Download 3] Download SMPE build log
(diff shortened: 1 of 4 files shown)


.github/workflows/build-packaging.yml
```diff
@@ -276,7 +276,7 @@ jobs:
         with:
           manifest-file-path: ${{ github.workspace }}/manifest.json
           default-target-path: .pax/binaryDependencies/
-          expected-count: 28
+          expected-count: 30
```
Martin-Zeithaml, line 279: Missing update in https://github.com/zowe/zowe-install-packaging/tree/v3.x/staging/bin#utilities

MarkAckert:
After the change, the build is failing with the below. Investigating.

`ZWESVUSR ERROR (zwe-internal-start-prepare,validate_components) Error: Component zss ended with nonzero rc=8`

Martin-Zeithaml:
As reported by @pablocarle, the night build fails to start, it's probably due to the zss validate script is ahead of this PR and trying to execute a binary, which is not there yet.

Martin-Zeithaml (approved):

Issue #4231, opened 2025-03-12
`bin/lib/certificate.sh`: combination of `cat | sed | >` is returning different data than expected


Martin-Zeithaml:
https://github.com/zowe/zowe-install-packaging/blob/ba6ab4904083c8882da3280be2fdb8ec6d263bde/bin/libs/certificate.sh#L1047-L1075

This script:
* Is using variable `result`, which is **always empty** (output is redirected to a file, possible error goes to `stderr`, not `result` )
* `code` is 
  * **0** even in case `cat`  is failing: it is returning nothing, `sed` successfully changes nothing to nothing, which is saved to the temp file and  `code` is `0`
  * `sed` is correct, it should not fail
  * **1** in case the redirection fails (e.g. writing to read-only filesystem)
* `chmod` before the `-f ` check
* print debug or trace suppose the `result` contains a text, which is not true

saurabh12nxf:
@Martin-Zeithaml can you review the #4314 PR

Martin-Zeithaml:
Closed with #4314

PR #4429, opened 2025-07-25
Add back zwe init shell code path

1000TurquoisePogs:
This PR temporarily adds back zwe init shell code path as an option.
This option is toggled using the `zowe.setup.jcl.enable` option.
It does not enhance the behavior of zwe init when disabled, but just exhibits 3.2-like behavior.


bin/libs/common.sh
```diff
@@ -20,6 +20,22 @@ export _BPXK_AUTOCVT="ON"
 export _EDC_ADD_ERRNO2=1                        # show details on error
 unset ENV             # just in case, as it can cause unexpected output
 
+# Temporarily, the zwe init commands have old shell code and new jcl code paths available.
+# To determine which are used, we check if --generate is given (implies jcl)
+# If not, we check if zowe.setup.jcl.enabled is explicitly false.
+# If so, shell is used, otherwise jcl is used.
+check_jcl_init() {
+  enabled=$(yaml_read_configmgr "${ZWE_CLI_PARAMETER_CONFIG" ".zowe.setup.jcl.enable")
```
Martin-Zeithaml: The old code means one important limitation: config is only uss file.

We have to start this check in `check_configmgr_enabled` style:
https://github.com/zowe/zowe-install-packaging/blob/fe7030e0bdf6cd0615ab5c1934fef5cd21bae68e/bin/libs/common.sh#L31

* `--config /uss/path/zowe.yaml` 
  *  Old or new code depends on `zowe.setup.jcl.enable`
* `--config FILE(zowe.yaml):PARMLIB(ZOWE(YAML))`
  * New code automatically

bin/libs/common.sh
```diff
@@ -20,6 +20,22 @@ export _BPXK_AUTOCVT="ON"
 export _EDC_ADD_ERRNO2=1                        # show details on error
 unset ENV             # just in case, as it can cause unexpected output
 
+# Temporarily, the zwe init commands have old shell code and new jcl code paths available.
+# To determine which are used, we check if --generate is given (implies jcl)
+# If not, we check if zowe.setup.jcl.enabled is explicitly false.
+# If so, shell is used, otherwise jcl is used.
+check_jcl_init() {
```
Martin-Zeithaml: If you code this (will fail on missing prefix on purpose):
```yaml
zowe:
  setup:
    dataset:
      prefix: ~
```

If will fail in javaScript code, but it should fail in shell - by default old code.

Also:
* `ZWE_CLI_PARAMETER_DATASET_PREFIX` is unique for old `zwe install`
* The code should check if `FILE()` or `PARMLIB()` was used -> only new code can handle this.
* And I think, this is opposite of what we discussed: user must use `zowe.setup.jcl.enable=true` to request new code.  


```shell
# Temporarily, the zwe init commands have old shell code and new jcl code paths available.
# To determine the code path:
#   Config is uss file - more checks
#     --generate - implies JCL code
#     --dataset-prefix - parameter unique for zwe install & shell only
#     zowe.setup.jcl.enable=false | undefined -> shell code
#     zowe.setup.jcl.enable=true -> JCL code
#   Not uss file
#     FILE | PARMLIB | wrong input (we don't care here)
check_jcl_init() {
  # If uss file
  if [ -f "${ZWE_CLI_PARAMETER_CONFIG}" ]; then
    enabled=$(read_yaml_configmgr "${ZWE_CLI_PARAMETER_CONFIG}" ".zowe.setup.jcl.enable")
    if [ "${ZWE_CLI_PARAMETER_GENERATE}" = "true" ]; then
      echo "true"
    elif [ ! -z "${ZWE_CLI_PARAMETER_DATASET_PREFIX}" ]; then
      echo "false"
    elif [ "${enabled}" = "true" ]; then
      echo "true"
    else
      echo "false"
    fi
  # Not uss file
  else
    echo "true"
  fi
}
```

bin/commands/init/index.sh
```diff
@@ -11,8 +11,105 @@
 # Copyright Contributors to the Zowe Project.
```
Martin-Zeithaml, line 11: We have to update the help for `zwe init`, it should have two versions too.

bin/commands/init/apfauth/index.sh
```diff
@@ -11,8 +11,72 @@
...
+    # authLoadlib can be empty
+    if [ "${key}" = "authLoadlib" ]; then
+      ds="${prefix}.${ZWE_PRIVATE_DS_SZWEAUTH}"
+    else
+      print_error_and_exit "Error ZWEL0157E: ${name} (zowe.setup.dataset.${key}) is not defined in Zowe YAML configuration file." "" 157
+    fi
+  fi
+
+  print_message "APF authorize ${ds}"
+  apf_authorize_data_set "${ds}"
```
Martin-Zeithaml: Undocumented SDSF requirement: https://docs.zowe.org/stable/user-guide/systemrequirements-zos/#zos

bin/commands/init/mvs/index.sh
```diff
@@ -10,10 +11,120 @@
...
+    if [ $? -ne 0 ]; then
+      print_error_and_exit "Error ZWEL0111E: Command aborts with error." "" 111
+    fi
+  fi
+done <<EOF
+$(echo "${cust_ds_list}")
+EOF
+print_message
+
+if [ "${ds_existence}" = "true" ] &&  [ "${ZWE_CLI_PARAMETER_ALLOW_OVERWRITE}" != "true" ]; then
```
Martin-Zeithaml, line 100: Old bug: combination of
* `ds_existence` is only for the last dataset
* Bug in (extra `//`) https://github.com/zowe/zowe-install-packaging/blob/fe7030e0bdf6cd0615ab5c1934fef5cd21bae68e/bin/libs/zos-dataset.sh#L78


Resulting in message "dataset will not be overwritten" and then it is overwritten.
```
Create data sets if they do not exist
Warning ZWEL0301W: ZOWE.PR#4429.PARMLIB already exists and will not be overwritten. For upgrades, you must use --allow-overwrite.
Warning ZWEL0301W: ZOWE.PR#4429.JCLLIB already exists and will not be overwritten. For upgrades, you must use --allow-overwrite.
Creating ZOWE.PR#4429.PLUGINS

Copy ZOWE.PR#4429.SZWESAMP(ZWESIP00) to ZOWE.PR#4429.PARMLIB(ZWESIP00)
```

bin/commands/init/stc/index.sh
```diff
@@ -11,8 +11,265 @@
...
+
+  # ZWESISTC
+  print_message "Modify ZWESISTC and save as ${jcllib}(${security_stcs_zis})"
+  tmpfile=$(create_tmp_file $(echo "zwe ${ZWE_CLI_COMMANDS_LIST}" | sed "s# #-#g"))
+  print_debug "- Copy ${prefix}.${ZWE_PRIVATE_DS_SZWESAMP}(ZWESISTC) to ${tmpfile}"
+  result=$(cat "//'${prefix}.${ZWE_PRIVATE_DS_SZWESAMP}(ZWESISTC)'" | \
+          sed '/^..STEPLIB/c\
+\//STEPLIB  DD   DSNAME='${authLoadlib}',DISP=SHR\
+\//         DD   DSNAME='${authPluginLib}',DISP=SHR' | \
+          sed "s/^\/\/PARMLIB .*\$/\/\/PARMLIB  DD   DSNAME=${parmlib},DISP=SHR/" \
```
Martin-Zeithaml: Will not fit into the 80 chars:
```
----+----1----+----2----+----3----+----4----+----5----+----6----+----7----+----8
//PARMLIB  DD   DSNAME=----+----1----+----2----+----3----+----4----,DISP=SHR
```
`DISP=SHR` must be on extra line.

bin/commands/init/stc/index.sh
```diff
@@ -11,8 +11,265 @@
...
+  print_message "Modify ZWESLSTC and save as ${jcllib}(${security_stcs_zowe})"
+  tmpfile=$(create_tmp_file $(echo "zwe ${ZWE_CLI_COMMANDS_LIST}" | sed "s# #-#g"))
+  print_debug "- Copy ${prefix}.${ZWE_PRIVATE_DS_SZWESAMP}(ZWESLSTC) to ${tmpfile}"
+  if [[ "$ZWE_CLI_PARAMETER_CONFIG" != /* ]];then
+    print_message "CONFIG path defined in ZWESLSTC is converted into absolute path and may contain SYSNAME."
+    print_message "Please manually verify if this path works for your environment, especially when you are working in Sysplex environment."
+  fi
+  result=$(cat "//'${prefix}.${ZWE_PRIVATE_DS_SZWESAMP}(ZWESLSTC)'" | \
+          sed "s/^\/\/STEPLIB .*\$/\/\/STEPLIB  DD   DSNAME=${authLoadlib},/" | \
+          sed "s#^CONFIG=.*\$#CONFIG=$(convert_to_absolute_path ${ZWE_CLI_PARAMETER_CONFIG})#" \
```
Martin-Zeithaml: CONFIG is limited to 74 chars. 
Either we have to split it or print warning/error message.

bin/commands/init/stc/index.sh
```diff
@@ -11,8 +11,265 @@
...
+  fi
+  print_debug "- ${jcllib}(${security_stcs_zis}) is prepared"
+
+  # ZWESASTC
+  print_message "Modify ZWESASTC and save as ${jcllib}(${security_stcs_aux})"
+  tmpfile=$(create_tmp_file $(echo "zwe ${ZWE_CLI_COMMANDS_LIST}" | sed "s# #-#g"))
+  print_debug "- Copy ${prefix}.${ZWE_PRIVATE_DS_SZWESAMP}(ZWESASTC) to ${tmpfile}"
+  result=$(cat "//'${prefix}.${ZWE_PRIVATE_DS_SZWESAMP}(ZWESASTC)'" | \
+          sed '/^..STEPLIB/c\
+\//STEPLIB  DD   DSNAME='${authLoadlib}',DISP=SHR\
```
Martin-Zeithaml: The sample ZWESASTC was changed, resulting into:
```jcl
//STEPLIB  DD   DSNAME=ZOWE.PR#4429.SZWEAUTH,DISP=SHR 
//         DD   DSNAME=ZOWE.PR#4429.PLUGINS,DISP=SHR  
//             DISP=SHR                                       
//         DD  DSNAME={zowe.setup.dataset.authPluginLib},     
//             DISP=SHR                                       
//SYSPRINT DD  SYSOUT=*                               
```

bin/commands/init/security/index.sh
```diff
@@ -11,8 +11,190 @@
...
+  security_stcs_zis=${ZWE_PRIVATE_DEFAULT_ZIS_STC}
+fi
+security_stcs_aux=$(read_yaml_configmgr "${ZWE_CLI_PARAMETER_CONFIG}" ".zowe.setup.security.stcs.aux")
+if [ -z "${security_stcs_aux}" ]; then
+  security_stcs_aux=${ZWE_PRIVATE_DEFAULT_AUX_STC}
+fi
+
+###############################
+# prepare ZWESECUR JCL
+print_message "Modify ZWESECUR"
```
Martin-Zeithaml, line 91: We have to alter the job header and remove "new" variable:
`//ZWESECUR JOB {zowe.setup.jcl.header}` -> `//ZWESECUR JOB`

bin/commands/init/security/index.sh
```diff
@@ -11,8 +11,190 @@
...
+  print_error_and_exit "Error ZWEL0157E: Zowe dataset prefix (zowe.setup.dataset.prefix) is not defined in Zowe YAML configuration file." "" 157
+fi
+# read JCL library and validate
+jcllib=$(read_yaml_configmgr "${ZWE_CLI_PARAMETER_CONFIG}" ".zowe.setup.dataset.jcllib")
+if [ -z "${jcllib}" ]; then
+  print_error_and_exit "Error ZWEL0157E: Zowe custom JCL library (zowe.setup.dataset.jcllib) is not defined in Zowe YAML configuration file." "" 157
+fi
+security_product=$(read_yaml_configmgr "${ZWE_CLI_PARAMETER_CONFIG}" ".zowe.setup.security.product")
+if [ -z "${security_product}" ]; then
+  security_product=RACF
```
Martin-Zeithaml, line 54: I not sure anymore, if this code is breaking change, bugfix or enhancement:

```shell
  security_product=$("${ZWE_zowe_runtimeDirectory}/bin/utils/getesm")
  if [ -z "${security_product} -o "${security_product}" = "NONE ]; then
    security_product=RACF
  fi
```

bin/commands/init/apfauth/index.sh
MarkAckert: This is also in `ZWEIAPF2` ...

bin/commands/init/mvs/index.sh
Martin-Zeithaml: We can postpone fixing this.

bin/commands/init/stc/index.sh
MarkAckert: I added an error message and exit. I think this existed before so it's not a regression, the error out is reasonable for now. Better solution in `--jcl`

bin/commands/init/stc/index.sh
MarkAckert: modified the sed's to try and accomplish this, worked in a test on the side

bin/commands/init/stc/index.sh
MarkAckert: same comment as zwesistc

Martin-Zeithaml:
## `zwe init certificate` issues
### Config
* 3.2.0 - config is a file
* This - config is a file merged with defaults.yaml
https://github.com/zowe/zowe-install-packaging/blob/64c180d3281073ae6f540a99e467ba47ef383599/bin/commands/init/certificate/index.sh#L31
### JCLLIB
Setup for "old code" is requesting submitting `ZWEGENER`
```
[io.makeYaml]: ./yaml/zowe.cert.yaml:
zowe:
  setup:
    dataset:
      prefix: ZOWE.PR#4429
    certificate:
      type: JCERACFKS
      keyring:
        name: TestRing

[env.exportEnv]: ZWE_CLI_PARAMETER_CONFIG=
[1/1]: /4429/bin/zwe init certificate -c ./yaml/zowe.cert.yaml
Error ZWEL0319E: zowe.setup.dataset.jcllib does not exist, cannot run. Run 'zwe init', 'zwe init generate',
or submit JCL ZOWE.PR#4429.SZWESAMP(ZWEGENER) before running this command.
```

MarkAckert:
I've got the core workaround in place; the shell-only code is now branching to 3.2 versions of keyring and no keyring shell methods, and uses ZWEKRING/ZWENOKYR. 

**Updated 8/4: Tested all 3 keyring options against RACF,ACF2,TSS, with and without `--allow-overwrite`**

I've adjusted the following:
* All the new `KR*#` JCL uses a symbol for the z/OSMF CA label. We were generating invalid JCL if a user didn't set `zowe.setup.certificate.keyring.zOSMF.ca`, and we didn't require the field to be set. There's an `_auto_` function in both the old and new keyring paths, and the new one wasn't respecting it.
* I'm erroring out if the IP address is unset, because again we have invalid JCL. Better message to be written later telling the user to fill out `san` or add an IP to `externalDomains`. This will need a doc update as well.
* `zowex` updated and hack removed
* Changed `example-zowe.yaml` so keyring options 1 and 3 have `label` and `caLabel` uncommented now. These are required fields.

Issues:
* Scenario 3 (import from DS) fails due to schema error - ` (zowe.setup.certificate.keyring.label) is not defined in Zowe YAML configuration file.`
  - Underlying cause: `read_yaml_configmgr` and the `configmgr extract` function do not respect defaults set in the schema! If the field is not present in the `zowe.yaml`, it will not be returned by `extract`. This explains my first note in the below section on `label` and `caLabel`, which are set as defaults in the schema file.
  - This same problem afflicts `init generate` and `ZWEGEN00`; default schema values are not supplied when rendering the JCL templates.
  - opened [an issue request](https://github.com/zowe/zowe-common-c/issues/541). I don't know what other impact changing the behavior of extract might have on `zwe` commands.
  - Worked around this by uncommenting label/caLabel in the `example-zowe.yaml`. There's no reason we should comment this out or hide the defaults. In other cases like `zOSMF.ca`, using JCL symbols and `sed` replacements. Doc needs to be updated to match.  No impact on schema or validation outside `init certificate`
* `KRT1`, `KRT2` and `KRT3` do not always end with RC=0, so the `init certificate` is displaying an error for all three. This appears consistent with prior behavior, `ZWEKRING` always ends with RC=4 on TSS.
  - Ditto for ACF2 JCL
* Top Secret JCL do not support parameterized `keyring.name`. This appears true in the prior implementation as well.
  - Ditto for ACF2 JCL

Other notes:
* `label` and `caLabel` weren't required fields for `init certificate` in 3.2, but they are in this revision. I think we should accept this change. Doc update necessary. 
* It's possible to run into mysterious failures due to user error. The `keyring.name` should disallow `safkerying://` prefixes. (Strangely, the new JCL path seemed fine with this in RACF, the old one is the one that broke. Better to disallow regardless). 
* We might want to indicate somewhere that users should run `init generate` before `init certificate` when they make changes to their zowe.yaml. It's an easy to forget step when you're tweaking certificate information in the YAML and using `--jcl`.
* I used the `--jcl` parameter to indicate new JCL, but it's kind of confusing and ridiculous in context; the old certificate + keyring path also submits JCL. I'm open to a better parameter name...
* RC=0 on every submission isn't ideal, I'd like to move the existing warning to the very end of the certificate's stdout in all cases so it's the first thing a user sees. 

> ## `zwe init certificate` issues
> ### Config
> * 3.2.0 - config is a file
> * This - config is a file merged with defaults.yaml
> ```

Martin-Zeithaml:
This [commit](https://github.com/zowe/zowe-install-packaging/pull/4429/commits/ceafe4115edb963c14925b8f50aaada9b9447114) fixes the #4231.

Martin-Zeithaml:
There is a general issue with `sed` command used in `bin/libs/certificates.sh`, for example:
https://github.com/zowe/zowe-install-packaging/blob/2a3acaad92ceb0d78e34f1687be2dd143ab6dd2e/bin/libs/certificate.sh#L897-L905

* Any value with `/` breaks the `sed`
* Old bug
* This can be fixed by replacing `/` with "safe char" (character which cannot be in the value).

Martin-Zeithaml (approved):
Approved.

MarkAckert (approved):

PR #4431, opened 2025-07-29
Seeing more customers use TLS and a non 23 port, so documenting the defaults for what to override 

Joe-Winchester:
#### PR type

- [X] Feature 

- Add comments to explain how to configure the Zowe Desktp TN3270 app to use an alternative port number and protocol. 

- [ ] Yes
- [X] No

- [ ] Yes
- [X] No


example-zowe.yaml
```diff
@@ -328,9 +328,12 @@ zowe:
   #   # Example of a global environment variable for all components
   #   MY_ENV_VAR: my_env_val
 
-  #   # Another example to customize SSH port for VT Terminal Desktop app
+  #   # To customize SSH port for VT Terminal Desktop app
   #   ZWED_SSH_PORT: 22
-  #   ZWED_TN3270_PORT: 23
+
+  #   # To customize port and protocol for TN3270 Desktop app
```
Martin-Zeithaml, line 334: We have documented those variables here:
https://docs.zowe.org/stable/user-guide/mvd-3270/

Maybe we can update that chapter or provide better examples there.

PR #4433, opened 2025-07-29
Backward compat for apiml modulith by declaring apiml components enabled when they are not

1000TurquoisePogs:
This PR includes https://github.com/zowe/launcher/pull/160 and https://github.com/zowe/zlux-app-server/pull/345 and https://github.com/zowe/zss/pull/787

Possibly zss also needs a change, it needs to be reviewed.

This PR attempts to improve backward compatibility of apiml modulith by doing 2 things

1) pretending individual components like "gateway" and "discovery" are enabled whenever "apiml" is
2) having the startup code (this repo, launcher) be smart enough to know that this is a lie, and skip over doing any actual start operations on those individual components

The app-server PR should therefore not be needed, because it should be captured by this compatibility, but to avoid technical debt the app-server PR has been made at this time anyway.

Ideally, this PR would make it so that users do not need to edit their yaml to mark every individual apiml component as "disabled" to use apiml modulith, because setting them to "enabled" won't really do anything, the startup code will ignore it and the startup code will also just inform extensions everything is enabled.

### DIFF
--- a/manifest.json.template
+++ b/manifest.json.template
@@ -16,7 +16,7 @@
       "artifact": "zowe-server-*.pax.Z"
     },
     "org.zowe.zlux.zlux-core": {
-      "version": "^3.0.0-V3.X-STAGING-ZLUX-CORE",
+      "version": "^3.3.0-zlux-app-server-PR-345",
       "artifact": "*.pax"
     },
     "org.zowe.zlux.sample-angular-app": {
@@ -123,7 +123,8 @@
       "artifact": "*.pax"
     },
     "org.zowe.launcher": {
-      "version": "^3.0.0-STAGING"
+      "version": "^3.3.0-PR-160",
+      "artifact": "*.pax"
     },
     "org.zowe.keyring-utilities": {
       "version": "1.0.4",
--- a/manifest.json.template
+++ b/manifest.json.template
@@ -44,7 +44,7 @@
       "artifact": "*.pax"
     },
     "org.zowe.zss": {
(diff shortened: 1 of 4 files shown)


pablocarle:
Is this PR complementary to #4254 ?

pablocarle:
> Is this PR complementary to #4254 ?

Sorry, didn't notice the base of the PR

1000TurquoisePogs:
I was testing this and so far it appears to work as expected:
- apiml components are started except when modulith is on, then only it is started
- regardless of the above, the env vars are always marked as enabled when modulith is enabled, otherwise they have the values seen in the yaml

this provides some backwards compatibility and less steps to enable the modulith.
the associated zss and app-server PRs are still required because they read the yaml directly instead of env vars, so the trick of marking the env vars as different from the yaml does not affect them.

manifest.json.template
```diff
@@ -16,7 +16,7 @@
       "artifact": "zowe-server-*.pax.Z"
     },
     "org.zowe.zlux.zlux-core": {
-      "version": "^3.0.0-V3.X-STAGING-ZLUX-CORE",
```
pablocarle, line 19: Should these be changed? (I can also roll them back in the base PR ofc)

pablocarle (approved):

manifest.json.template
MarkAckert: Let's merge this in, and then update the manifest when the builds from staging are available (by end of today)

MarkAckert (approved):

PR #4441, opened 2025-08-07
Tweaks to `init security`

MarkAckert:
This PR should be included in 3.3.0 and will require a 3.3.0-RC2. More testing is required for TSS, ACF2.

This PR changes the following:
* `--jcl` parameter added to `init security` 
* One typo in `ZWEIRAC`
* All `ZWEIACF` JCL restored sections adding the resource class definition for `ZOWE`, and `APIML.SERVICES`. 
  - Note: we consulted security admins and confirmed the `ZOWE` resource class comes defined out of the box with TSS and RACF now. It does not come out of the box with ACF2, so we're leaving in the definition-by-default there.

* This PR does not re-introduce commands for setting up `SUPERUSER.FILESYS`, as adding the Zowe STC to the Zowe Admin Group should prevent the need for it.

....details....

~~In ACF and TSS (didn't check RACF yet), the following block was removed~~ The below removal is *accepted*, the ZOWE STC is added to the &ADMINGRP, which has write authority to `zowe.workspaceDirectory`.

```
  TSS ADD(&FACACID.) UNIXPRIV(SUPERUSE)
  TSS WHOHAS UNIXPRIV(SUPERUSER.FILESYS)
  TSS PER(&ZOWEUSER.) UNIXPRIV(SUPERUSER.FILESYS) ACCESS(CONTROL)
```

~~In all ESMs this was removed~~ This section was re-introduced and runs by default in all ESMs. Users who want to use the `APIML.SERVICES` resource class to protect access to system APIs must ensure the `ZOWE` resource is configured.
```
* define ZOWE resource type and class mapping
* skip this section if the ZOWE resource class already exists
SET CONTROL(GSO)
INSERT CLASMAP.ZOWE RESOURCE(ZOWE) RSRCTYPE(ZWE)
F ACF2,REFRESH(CLASMAP),TYPE(GSO)
CHANGE INFODIR TYPES(R-RZWE)
F ACF2,REFRESH(INFODIR)
SET CONTROL(GSO)
```


Martin-Zeithaml (approved):

files/SZWESAMP/ZWEITSS
```diff
@@ -215,9 +215,16 @@ TSS PERMIT({zowe.setup.security.users.zowe}) APPL(OMVSAPPL)
 
 /* DEFINE ZOWE RESOURCE PROTECTION ................................. */
 
+/* - Defines new resource class for Zowe that protects access to     */
+/*   sensitive Zowe resources.                                       */
 /* - Defines resource APIML.SERVICES that controls access to         */
 /*   detailed information about API services to Zowe users.          */
 
+/* define ZOWE resource class                                        */
+/* skip this command if the ZOWE resource class already exists       */
```
1000TurquoisePogs: I asked one of the Rocket admins if this is needed on RACF/TSS/ACF2.
I know its not needed for RACF since zos 2.5 (and so this PR was updated, thanks)
But the admin also sees it present on TSS, but not on ACF2.
So it could be that this block is only needed for ACF2???

They were going to ask Broadcom for more info but, if anyone knows for sure when/if TSS/ACF2 versions have already included the class, it would be good to capture this knowledge. Otherwise, we could be producing an error when running this job?

1000TurquoisePogs (dismissed):
Requesting review on this topic.
Once discussed fully, then we can proceed.

files/SZWESAMP/ZWEITSS
1000TurquoisePogs: OK - they heard back that indeed ACF2 does not include this class.
This leaves uncertainty about TSS. Who can confirm if TSS really needs the class defined by us rather than that it ships it?

1000TurquoisePogs: I have now heard back that TSS DOES include this class.
So, only ACF2 needs this change.

MarkAckert: Updated `ZWEITSS` so the resource is longer defined. There are no longer any changes to `ZWEITSS` in this PR.

PR #4447, opened 2025-08-08
Add "zwe validate port bind" and "zwe validate config"

1000TurquoisePogs:
This re-uses some code from the prototype https://github.com/zowe/zowe-install-packaging/pull/4276 but submits just "zwe validate port bind" and "zwe validate config" for review to keep the PR smaller and easier to test, read, merge.

This PR adds a command, "zwe validate config" which is an alias of "zwe config validate".
It then adds a command "zwe validate port bind" which can either test if a single component can bind to its associated port, or if all enabled components can.

That command is then used every startup, unless opted out via `zowe.launchScript.startupChecks.bypassAll: true` or `zwe.launchScript.startupChecks.ports: false`
You can also use "warn" as a value, which will do the check but not prevent startup.
When the check is done and not in warn-mode, it will print how to put into warn-mode, in case a user runs into a problem and needs to bypass the check.

This code is more robust than just the is_port_available code that apiml was using from the zwe codebase.
It attempts to make a tcp server to see if the permissions are correct, not just whether the port is free, and attempts to do so with the right jobname using _BPX_JOBNAME changing.

It is intended as a replacement of is_port_available and if merged, I would like apiml to stop using is_port_available since it is just then duplicate processing.

This PR also includes a manifest schema addition to have components/extensions tell us their likely jobname.
This could be helpful overall, but is used here to try to get the right jobname to check permissions of the bind operation.


bin/commands/validate/config/.examples
```diff
@@ -0,0 +1,3 @@
+zwe validate config -c /path/to/zowe.yaml
+zwe validate config -c FILE(/customizations/zowe.yaml):FILE(/defaults/zowe.yaml) --all
```
Martin-Zeithaml: Apostrophes are mandatory, otherwise the command ends with `FSUM7332 syntax error: got (, expecting Newline`

files/defaults.yaml
```diff
@@ -130,6 +130,9 @@ zowe:
     logLevel: "info"
     # Set to "exit" if you'd like startup to exit if any component has an error in the configure stage, otherwise zwe will warn but continue.
     onComponentConfigureFail: "warn"
+    startupChecks:
+      bypassAll: false
```
MarkAckert: Should `bypassAll` be respected by all other validation scripts, like the certificate check in zlux-app-server?

MarkAckert:
Two errors:

`bind-test` returned S0C4 when the port was already in use:

```
_BPX_JOBNAME=WHOSV ./bind-test --host 0.0.0.0 --port 7554
Error: Bind failed (rc=0x45b, rsn=0x744c7247)
Error: Port 7554 was already occupied
CEE3204S The system detected a protection exception (System Completion Code=0C4).
         From compile unit ZZOW11:../../deps/zowe-common-c/c/bpxskt.c/ at entry point socketClose at statement 1674 at compile unit offset +000000001A70C594 at entry offset +000000000000004C at address 000000001A70C594.
SIGSEGV
```

`bind-test` could not resolve the hostname for the Marist system:

```
_BPX_JOBNAME=WHOSV ./bind-test --host <zzowhost> --port 7550
Error: Could not resolve hostname from DNS. Is this a valid hostname for this system?
Error: Invalid address given for --host
```
```
ping <zzowhost>
CS 3.1: Pinging host <zzowhost> (<zzowip>)
Ping #1 response took 0.003 seconds. (3.378 milliseconds)
```

files/defaults.yaml
1000TurquoisePogs: Yes I hope people respect that and, that one will probably go away in the future because I have a draft of using certificate-analyser.jar universally instead of the nodejs-bound app-server one.

1000TurquoisePogs:
Great catch - I can't close a socket I didn't open, that must be the 0c4 cause.

1000TurquoisePogs:
About hostname not being able to be resolved - I am guessing zss has this same issue but its not seen because the servers are being instructed to try "0.0.0.0" instead of "<zzowhost>". Seems like an issue but not one unique to this and therefore should be able to have this checker be no worse than zss.

bin/commands/validate/port/bind/index.sh
```diff
@@ -0,0 +1,19 @@
...
+# SPDX-License-Identifier: EPL-2.0
+#
+# Copyright Contributors to the Zowe Project.
+#######################################################################
+
+if [ -z "${ZWE_PRIVATE_TMP_MERGED_YAML_DIR}" ]; then
+  # user-facing command, use tmpdir to not mess up workspace permissions
+  export ZWE_PRIVATE_TMP_MERGED_YAML_DIR=1
+fi
+_CEE_RUNOPTS="XPLINK(ON),HEAPPOOLS(OFF),HEAPPOOLS64(OFF)" ${ZWE_zowe_runtimeDirectory}/bin/utils/configmgr -script "${ZWE_zowe_runtimeDirectory}/bin/commands/validate/port/available/cli.js"
```
Martin-Zeithaml: `/4447/bin/commands/validate/port/available/cli.js: EDC5129I No such file or directory. (errno2=0x0594003D)`
```
`/bin/commands/validate/port/available/cli.js` -> `/bin/commands/validate/port/bind/cli.js`

**Comment by MarkAckert** [comment]
My latest tests look better with the updated bind-test though `localhost` isn't resolvable. I'm still getting some curious output that's a little unintuitive. When I ask `bind-test` if a port is free for a given host, I must specify the exact host value used by the STC binding to that port, otherwise the tool reports that the port is free. I think the tool will work OK in the context of a pre-flight check for Zowe if we rely on the `zowe.yaml` for the host, but we may get inconsistent results for other services like z/OSMF depending on how users setup that STC and configure the `zowe.yaml` file.

e.g. Zowe is running on `0.0.0.0`:

```
./bind-test  --host 0.0.0.0 --port 7554
Error: Bind failed (rc=0x45b, rsn=0x744c7247)
Error: Port 7554 was already occupied
...
./bind-test  --host 127.0.0.1 --port 7554
Resolved IP address as 127.0.0.1
Bind succeeded (pointer=0x1B0FD098, rc=0x0, rsn=0x0)
...
./bind-test  --host `zzowhost`--port 7554
Resolved IP address as `zzowip`
Bind succeeded (pointer=0x1B100098, rc=0x0, rsn=0x0)
```

e.g. z/OSMF is bound to the hostname:
```
./bind-test  --host 0.0.0.0 --port <port>
Bind succeeded (pointer=0x1B0FD098, rc=0x0, rsn=0x0)
...
./bind-test  --host 127.0.0.1 --port <port>
Resolved IP address as 127.0.0.1
Bind succeeded (pointer=0x1B0FD098, rc=0x0, rsn=0x0)
...
./bind-test  --host `zzowhost` --port <port>
Resolved IP address as `zzowip`
Error: Bind failed (rc=0x45b, rsn=0x744c7247)
Error: Port <port> was already occupied
```

Localhost is not resolvable:

```
./bind-test  --host localhost --port 7554
Error: Could not resolve hostname from DNS. Is this a valid hostname for this system?
Error: Invalid address given for --host
```

**Comment by 1000TurquoisePogs** [comment]
@MarkAckert the behavior of 0.0.0.0 is unintuitive, but correct.
I tested this now by running 2 servers.

If I try to run 2 servers both on 0.0.0.0, the second one fails
✅ bind-test can detect that

If I try to run 1 server at 0.0.0.0, and another at 127.0.0.1, they both succeed
✅ this is consistent with what bind-test says

Then I try to curl these 2 servers that appear to run at the same localhost IP
curl always gives the second started one, 127.0.0.1, until I stop it
After I stop it, curl gives the first one.

So - somehow the OS is just having the second program that gets a socket be the winner.
I did not expect this, but bind-test is doing the right thing. It's telling you what would really happen even if the truth is strange.

I tried this on both z/OS and macOS, same behavior! So somehow this is expected behavior but very strange indeed.

**Comment by 1000TurquoisePogs** [comment]
As for localhost, I am guessing the system you're testing on hasn't defined localhost. It's not a guaranteed value, it's usually defined in /etc/hosts
But, on my z/OS, your localhost test succeeds.

**Comment by 1000TurquoisePogs** [comment]
OK - manifest is back on normal branches, fix was merged.

File: bin/commands/validate/port/bind/index.ts [hunk 4]
```diff
@@ -0,0 +1,82 @@
...
+  const ZOWE_CONFIG = config.getZoweConfig();
+  let bindUtilPath = ZOWE_CONFIG.zowe.runtimeDirectory;
+  if (!bindUtilPath.endsWith('/')) {
+    bindUtilPath += '/';
+  }
+  bindUtilPath += 'bin/utils/bind-test';
+
+  let myJobname = std.getenv('_BPX_JOBNAME');
+
+  common.printFormattedInfo(`ZWELS`, `zwe-validate-port-available`, `Checking ports of ${enabledComponents.length} enabled components`);
```
Martin-Zeithaml:
```
zwe validate port bind -c /zowe/zowe.yaml
Bind succeeded (pointer=0x2D601098, rc=0x0, rsn=0x0)
Bind succeeded (pointer=0x2D601098, rc=0x0, rsn=0x0)
Bind succeeded (pointer=0x2D601098, rc=0x0, rsn=0x0)
Bind succeeded (pointer=0x2D601098, rc=0x0, rsn=0x0)
Bind succeeded (pointer=0x2D601098, rc=0x0, rsn=0x0)
Bind succeeded (pointer=0x3D601098, rc=0x0, rsn=0x0)
INFO (zwe-validate-port-available) Checking ports of 9 enabled components
INFO (zwe-validate-port-available) Zowe port bind validation passed.
```

1. Should we move the `Bind succeeded (pointer=0x...., rc=0x0, rsn=0x0)` to trace?
2. Should we state what was check? For example: `components.zaas: Bind succeeded (pointer=0x...., rc=0x0, rsn=0x0)`
3. `Checking ports of 9 enabled components` will check (for this instance) 6 components, because explorers are skipped (no port).  Should we inform about this? For example: `components.explorer-jes: Bind skipped due to no port used`

1000TurquoisePogs: I will put the skip message under debug because it's more likely a user says 'why are things skipped?' than 'why is the number wrong?' - people ask about what they think are scary looking messages all the time.

MarkAckert:
All builds are passing, one failure is known with OIDC and unrelated to these changes. Merging in.

PR #4463, opened 2025-08-27
`zwe install`: ZWEL0105E - do not print `undefined`

Martin-Zeithaml:
* `zwe install` - JCL feature
* Incorrect `zowe.runtimeDirectory` - e.g. typo in path, which causes `convertToAbsolutePath` is returning [`undefined`](https://github.com/zowe/zowe-install-packaging/blob/193b75be0142a347d186a1d6d1390e1eccdea5a7/bin/libs/fs.ts#L256)
* Then following error message `ZWEL0105E` refers to `undefined`

## Input config

```yaml
zowe:
  setup:
    jcl:
      enable: true
      header: '12345'
    dataset:
      prefix: ZOWE
  runtimeDirectory: /typo/in/path/
```

## Without fix

```
ERROR: Could not convert /typo/in/path/ to absolute path, err=129
ERROR: Error ZWEL0105E: The Zowe YAML config file is associated to Zowe runtime "undefined", which is not same as where zwe command is located "/zowe/test/4463".
```

## With fix

```
ERROR: Could not convert /typo/in/path/ to absolute path, err=129
ERROR: Error ZWEL0105E: The Zowe YAML config file is associated to Zowe runtime "/typo/in/path/", which is not same as where zwe command is located "/zowe/test/4463".
```

### DIFF
--- a/bin/commands/install/index.ts
+++ b/bin/commands/install/index.ts
@@ -47,10 +47,11 @@ export function execute(): void {
     runtime = runtimeEnv;
   } else {
     // We need clean path for xplatform.loadFileUTF8, otherwise will fail for e.g. /zowe/./files/SZWESAMP//ZWEINSTL
-    runtime = fs.convertToAbsolutePath(runtime);
-    if (runtime != runtimeEnv) {
+    const runtimeAbs = fs.convertToAbsolutePath(runtime);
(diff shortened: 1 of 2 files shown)


JoeNemo (approved):
Approved.

MarkAckert (approved):

PR #4481, opened 2025-09-09
`defaults.yaml` - no need for `files-api` and `jobs-api` in V3

Martin-Zeithaml:
`jobs-api` and `files-api` were removed with Zowe V3, but still used in defaults - both APIs are disabled, so it should be "safe".

But the port 7558 is used for `zaas` and (disabled) `jobs-api`, which might be confusing.


JoeNemo (approved):
Approved.

PR #4492, opened 2025-09-15
Fix --update-config's bug when zowe.yaml is a soft-link

### LINKED ISSUE #4491: --update-config might destroy the target yaml file if a symbolic link is specified
**Describe the bug**
The `--update-config` switch might destroy the target yaml file if a symbolic link of the yaml file is specified instead of a real file.

**To Reproduce**
Assumptions:
- You have a valid zowe.yaml.
- **It has a valid zowe.setup.certificate section.** (because we need to use the `init certificate --update-config` to reproduce the symptom)
- You have backed up your yaml file. (because it will be messed up)

1. cd /path/to/zowe-server/runtime
2. ln -s $(pwd)/zowe.yaml ./my_zowe.yaml
3. bin/zwe init certificate -c ./my_zowe.yaml --allow-overwrite --update-config -vv

**Expected behavior**
The command successfully finishes and the file is correctly updated (or at least remains valid).

**Screenshots**
You will see the following output:

Below is the last a few lines of the whole output. The "/var/zowe/zowe3.yaml" is a symbolic link to a real file.

```
>> Certificate is generated in keyring successfully.

>> Update certificate configuration to /var/zowe/zowe3.yaml

- update  "zowe.certificate.keystore.type" with value: JCERACFKS
  * Exit code: 0
  * Output:
- Remove encoding tag of /var/zowe/zowe3.yaml.
- update  "zowe.certificate.keystore.file" with value: safkeyring://ZWESVUSR/ZOWEKR
YAMLException: the stream contains non-printable characters at line 2, column 1:
    
    ^
Try --help to get usage information, or use --verbose option to display more details.
  * Exit code: 1
  * Output:
Error ZWEL0138E: Failed to update key zowe.certificate.keystore.file of file /var/zowe/zowe3.yaml.
```

Let's skip this...

**Additional context**
The problem happens at the following call-stack:

```
ls -T /path/to/a/symbolink  <--- it returns empty string due to the symbolic link
zos-fs.sh: get_file_encoding()  <--- so it returns "EBCDIC" by "guessing" even when it is actually ASCII
json.sh: update_zowe_yaml()  <--- it can update the target yaml but cannot change it back to EBCDIC which fails the next call to it.
...
```

**Solution**

```sh
###############################
# Get file encoding from z/OS USS tagging
#
# @param string   file name
# output          USS encoding if exists in upper case
get_file_encoding() {
  file="${1}"
  # Resolve symbolic link to real file
  if [ -L "${file}" ]; then
    file=$(readlink -f "${file}")
  fi
  # m ISO8859-1   T=off <file>
  # - untagged    T=off <file>
  ls -T "${file}" | awk '{print $2;}' | upper_case
  # or we can use "chtag -p" command, same result
}
```

I will create a PR with the above solution shortly.

ChongZhou-Broadcom:
Fixes #4491 

When specifying a symbolic link to `--config|-c`, the `--update-config` switch might fail to update the config. It is because the get_file_encoding() function fails to figure out the file encoding from symbolic link. The fix is easy: find the real file by readlink before using ls -T.

I will port the code change to 2.x when this PR is accepted.


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ All notable changes to the Zowe Installer will be documented in this file.
 
 ## `3.3.0`
 
+- Bugfix: `--update-config` might fail if the specified config file is a symbolic link.
```
Martin-Zeithaml: Add this to 3.4.0, Zowe 3.3.0 was already released. And add link to this PR.

Martin-Zeithaml:
```javascript
import * as zosfs from '/zowe/bin/libs/zos-fs';

console.log(zosfs.getFileEncoding('./helloWorld.js'));
console.log(zosfs.getFileEncoding('./linkToHelloWorld.js'));
```
Returns encoding for both files. Seems no update needed for JS version of `getFileEncoding`.

CHANGELOG.md
ChongZhou-Broadcom: Changed. Thanks.

Martin-Zeithaml (approved):

MarkAckert:
I ran a manual test against the `init certificate` command, and as a side-effect the symlink was replaced with a flat file. Updated config fields were present in the once-symlinked file, and not present in the original flat file.

~~This is likely happening from `update_zowe_yaml`'s call to `config-converter`. Looking into this...we will replace config-converter in 3.4.0, but will still use it in 2.18.3.~~

**Updated: config-converter is not responsible, it's `ensure_file_encoding()`. `iconv` doesn't respect symlinks. `export function ensureFileEncoding` in `zos-fs.ts` will have the same behavior.

**Update 2: made quick updates; passed manual tests for shell scripts. For `zos-fs.ts`, I'm not sure the `ensureFileEncoding` method is used anywhere.

ChongZhou-Broadcom:
@MarkAckert Hi Mark, thank you so much for fixing the side effect in code change. Sorry for the late response as I was struggling with something else in the last a few days.

Yes, you are right. Iconv should be responsible for this, not yargs.

I will update the branch of this PR so it is ready for merge. Then I will port your change back to v2's staging soon.

MarkAckert:
The current build failure is unrelated to this change, but needs to be fixed before merge. Github action runners were updated, and the newer version of ansible included is complaining about our playbooks. As soon as that's done I'll merge this.

PR #4493, opened 2025-09-15
Fix --update-config's bug when zowe.yaml is a soft-link (V2)

### LINKED ISSUE #4491: --update-config might destroy the target yaml file if a symbolic link is specified

ChongZhou-Broadcom:
Fixes #4491 

(This PR is porting code changes from [PR 4492](https://github.com/zowe/zowe-install-packaging/pull/4492) to Zowe V2)

When specifying a symbolic link to `--config|-c`, the `--update-config` switch might fail to update the config. It is because the get_file_encoding() function fails to figure out the file encoding from symbolic link. The fix is easy: find the real file by readlink before using ls -T.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to the Zowe Installer will be documented in this file.
 <!--Add the PR or issue number to the entry if available.-->
+
+## `2.18.3`
+- Bugfix: `--update-config` might fail if the specified config file is a symbolic link. [#4493](https://github.com/zowe/zowe-install-packaging/pull/4493)
+
 ## `2.18.2`
 - Bugfix: JCL sample `ZWEINSTL` did not include the `ZWESLSTC`. [#4382](https://github.com/zowe/zowe-install-packaging/pull/4382)
 - Bugfix: YAML lookup for HA instances within "haInstances" was not working when HA instance names were uppercase. Now, the lookup is case insensitive to allow for any casing of HA instance names. [#4378](https://github.com/zowe/zowe-install-packaging/pull/4378)
--- a/bin/libs/zos-fs.sh
+++ b/bin/libs/zos-fs.sh
@@ -18,6 +18,10 @@
 # output          USS encoding if exists in upper case
 get_file_encoding() {
   file="${1}"
+  # Resolve symbolic link to real file because "ls -T" does not work on symbolic link.
+  if [ -L "${file}" ]; then
+    file=$(readlink -f "${file}")
+  fi
   # m ISO8859-1   T=off <file>
   # - untagged    T=off <file>
   ls -T "${file}" | awk '{print $2;}' | upper_case


Martin-Zeithaml (approved):

ChongZhou-Broadcom:
@MarkAckert Hi Mark, thank you for accepting this PR into the staging branch. It has a [sister PR here](https://github.com/zowe/zowe-install-packaging/pull/4492), having the same code changes, but for V3's staging. I wonder if you could kindly merge that one too? Thanks.

MarkAckert:
> @MarkAckert Hi Mark, thank you for accepting this PR into the staging branch. It has a [sister PR here](https://github.com/zowe/zowe-install-packaging/pull/4492), having the same code changes, but for V3's staging. I wonder if you could kindly merge that one too? Thanks.

I will merge the v3 sister PR after I've written test cases for it. The same test framework was not available in the v2 line, and we have a 2.18.3 release coming soon, so I went ahead and merged this one first.

PR #4512, opened 2025-09-24
fix: set additional at-tls properties for z/OSMF

pablocarle:

#### PR type

- [ ] Bugfix

Include additional properties needed for z/OSMF to handle hybrid AT-TLS scenarios correctly.

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No


MarkAckert:
Is there a reason for defining two variables instead of one to represent `SECURE_PORT_ENABLED`? Is there any case here `SECURE_PORT_ENABLED` and `NON_SECURE_PORT_ENABLED` are true/true or false/false?

pablocarle:
> Is there a reason for defining two variables instead of one to represent `SECURE_PORT_ENABLED`? Is there any case here `SECURE_PORT_ENABLED` and `NON_SECURE_PORT_ENABLED` are true/true or false/false?

The problem is that we need the variables in the z/OSMF static definition, and unfortunately, the way install-packaging does the variable replacements in a template file have limited feature support, so not even simple conditions are possible https://github.com/zowe/api-layer/pull/4327

pavel-jares-bcm (approved):

MarkAckert:
> > Is there a reason for defining two variables instead of one to represent `SECURE_PORT_ENABLED`? Is there any case here `SECURE_PORT_ENABLED` and `NON_SECURE_PORT_ENABLED` are true/true or false/false?
> 
> The problem is that we need the variables in the z/OSMF static definition, and unfortunately, the way install-packaging does the variable replacements in a template file have limited feature support, so not even simple conditions are possible [zowe/api-layer#4327](https://github.com/zowe/api-layer/pull/4327)

I wonder if we could enhance template capabilities by using `configmgr` in the future, but it's not something we need to think about right now. I'll open an issue with that idea and get this PR merged. Thanks!

PR #4526, opened 2025-09-29
Perform cleanup of old apiml static def files from workspace

### LINKED ISSUE #3853: z/OSMF static definitions accumulates if configuration is changed
**Describe the bug**

New zosmf static definition files are created every time Zowe is started. The old files never removed, so in following scenarios they are accumulating. This can create problems like having multiple ibmzosmf services registered with APIML.

1. Zowe started once as a single insnance (non-HA) with non-HA z/OSMF defined in zowe.yaml. Just to try-it-out. But later the configuration is switched to HA with z/OSMF in HA (different hostname). 
2. Zowe started with different haInstances names, then they were defined in zowe.yaml before.
3. Zowe started with different values (could be different from name defined in zowe.yaml) of the HAINST parameter set during start of the instance (`/S ZWESVSTC,HAINST=`).

In all these cases in `api-defs` directory, there will be multiple `discovery.zosmf_static_definition_yaml_template.<system>.yml` with different systems in the name. As a result, all of them will be registered with APIML except those that have same hostname and port. 

Some of our customers was facing scenario 1.

The issue is related to #3328 and #3566

**Expected behavior**
During the startup the existing zosmf static definition files are removed before the creation of new one.

1000TurquoisePogs:
fixes https://github.com/zowe/zowe-install-packaging/issues/3853

If a user moves their zowe config from one system to another, or changes their ha instance name, or switches between apiml modulith on/off, entries like the zosmf static registration file can get duplicated within workspace/api-mediation/api-defs.
This then causes unpredictable behavior - some requests succeed, some fail, if apiml is switching between existing and not existing servers from multiple registration files.

This PR resolves that by using the startup code to checking for files that seem to be generated by the startup code, but are invalid for the current configuration.
Invalid old files would be if 'discovery' was found when modulith is used, or vice-versa, or if a file is found from a component which is from an ha instance which seems to not exist currently.

This does not target 'sideloaded' files. If there's a file in the folder that doesnt match the pattern that this code produces, then it is untouched.

Currently this also means files made by extensions that no longer exist - this code could be expanded to cover that, but does not currently.

Here's a test for the functionality - 
[test.js](https://github.com/user-attachments/files/22598268/test.js)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to the Zowe Installer will be documented in this file.
 
 ## `3.4.0`
 - Bugfix: internal routine `copy_to_data_set` did not correctly check if data set exists. [#4476](https://github.com/zowe/zowe-install-packaging/pull/4476)
+- Bugfix: Zowe startup now removes outdated static definition files from workspace/api-mediation/api-defs, which previously caused wrong or duplicate registrations of servers to APIML. [#4526](https://github.com/zowe/zowe-install-packaging/pull/4526)
 
 ## `3.3.0`
 
--- a/bin/libs/component.ts
+++ b/bin/libs/component.ts
@@ -424,6 +424,11 @@ function isClientAttls() {
   }
 }
 
+// This function both creates static definition files from manifest templates,
+// And cleans up existing ones that seem to be outdated
+// It does not touch existing files that are neither - it continues to permit sideloaded static definition files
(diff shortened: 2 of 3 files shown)


1000TurquoisePogs:
zwe internal start prepare calls this function **per each** enabled component
to fix left behind files from missing (uninstalled or disabled) components, the function needs to be rewritten to act upon **all** seen components, so that it can remove files from components no longer seen.

Martin-Zeithaml (approved):

Issue #4495, opened 2025-09-15
Enable `zwe` to use JCL by default in 3.4.0


MarkAckert:
With Zowe release 3.4.0, we would like to switch the default behavior from opt-in to `zwe` JCL to opt-out.

We can set `zowe.setup.jcl.enable: true` as an explicit default in example-zowe.yaml. 

To be discussed with squad: should a missing variable be interpreted as `true` in zowe v3.4+, or should we wait for zowe v4? By changing example-zowe.yaml and keeping the default interpretation as false, we can avoid changing the behavior for users with an older zowe.yaml. Is that the right outcome?

Martin-Zeithaml:
Original "Migrate to JCL" (without old shell code) was recognized as a breaking change. We keep the old code as default.
By defaulting to "JCL", it will be again the breaking change.

We have some options:
* Do nothing / wait for V4
* Justify why it is not a braking change anymore - see Note
* Others

Note: the shell function `read_yaml_configmgr` was originally supposed to replace `read_yaml` (because of Node.js dependency). Both `read_yaml_configmgr` and `read_yaml` are reading **without** defaults.yaml! We are making a decision in shell, if the JCL option will be used. If we move `zowe.setup.jcl.enable` to the defaults, we need to update `read_yaml_configmgr` to include `defaults.yaml`.

MarkAckert:
Complete with #4534

PR #4534, opened 2025-10-01
JCL as default option in `example-zowe.yaml`

Martin-Zeithaml:
## Changes
* https://github.com/zowe/zowe-install-packaging/issues/4495
* `example-zowe.yaml` sets `zowe.setup.jcl.enable=true`
* Couple help files updated

## Behavior
Is not changed. The default install & init method is shell. Users with old config will run update as usually. Users starting with new `example-zowe.yaml` most likely use JCL.


MarkAckert:
I think we need to omit the `zowe.setup.jcl.enable: true` from `defaults.yaml` for v3. In v4, we can use `defaults.yaml`.

Use case:
- A user got started with Zowe v3.0.0, they have a v3.0.0 `zowe.yaml` they've configured, but never ran through init and install.
- This user upgrades to 3.4.0 and keeps their configured `zowe.yaml` from 3.0.0.
- When they run `zwe init` and `zwe install` with Zowe 3.4.0, given the current approach, it will use JCL by default.

This use case was discussed in TSC, and they required that this user would **not** install with JCL by default. Omitting the default from `defaults.yaml` should solve the TSC requirement. Thoughts / other ideas? Consequences?

Martin-Zeithaml:
> I think we need to omit the `zowe.setup.jcl.enable: true` from `defaults.yaml` for v3. In v4, we can use `defaults.yaml`.
> 
> Use case:
> 
> * A user got started with Zowe v3.0.0, they have a v3.0.0 `zowe.yaml` they've configured, but never ran through init and install.
> * This user upgrades to 3.4.0 and keeps their configured `zowe.yaml` from 3.0.0.
> * When they run `zwe init` and `zwe install` with Zowe 3.4.0, given the current approach, it will use JCL by default.
> 
> This use case was discussed in TSC, and they required that this user would **not** install with JCL by default. Omitting the default from `defaults.yaml` should solve the TSC requirement. Thoughts / other ideas? Consequences?

This is a game changer:
* What I did: JCL is a default install & init method
* What is requested: update `example-zowe.yaml` - the behavior remains the same, shell is the default method, only new installation has higher probability to use new `example-zowe.yaml` & JCL

MarkAckert:
These changes look good, and all tests pass. I captured some unintuitive config combinations in a [new issue](https://github.com/zowe/zowe-install-packaging/issues/4549) we can discuss in squad call. The combos aren't related to this PR.

MarkAckert (approved):

PR #4489, opened 2025-09-12
`copy_to_data_set` bug fix and related changes (V2)

Martin-Zeithaml:
V2 version of #4476.

PR #4542, opened 2025-10-07
Patch PTF Playbooks (again)

MarkAckert:
*Tests from this PR need to run before accepting this*

Nightly PTF Automation passes sometimes for v3, and not at all for v2. 

After a closer look, this appears to be related to Ansible updates in the Github runners. Jinja templating is stricter, and a `null` response from a `regex_search` in the playbooks previously returned the empty string, but now it returns null, which breaks existing conditional checks looping while waiting for job output.  This PR sets the `default('')` empty string after the regex to coerce null values into empty strings, preserving the existing conditional checks.

This will need to be ported to v3. As far as I can tell, PTF playbooks will *sometimes* pass if the `LSTFMID` job completes fast enough, but it shouldn't be reliable at all.

### DIFF
--- a/bin/commands/init/vsam/index.sh
+++ b/bin/commands/init/vsam/index.sh
@@ -72,22 +72,14 @@ if [ "${jcl_existence}" = "true" ]; then
 fi
 
 # VSAM cache cannot be overwritten, must delete manually
-# FIXME: cat cannot be used to test VSAM data set
-vsam_existence=$(is_data_set_exists "${vsam_name}")
-if [ "${vsam_existence}" = "true" ]; then
-  # if [ "${ZWE_CLI_PARAMETER_ALLOW_OVERWRITE}" = "true" ]; then
-  #   result=$(tso_command delete "'${vsam_name}'")
-  # fi
-  # else
-  #   print_error_and_exit "Error ZWEL0158E: ${vsam_name} already exists." "" 158
-  # fi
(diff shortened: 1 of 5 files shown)


Martin-Zeithaml (approved):

MarkAckert:
@Martin-Zeithaml I temporarily undid some changes from https://github.com/zowe/zowe-install-packaging/pull/4489. The `init vsam` command still can't verify if VSAM data sets exist using `cat` from `is_data_set_exists()`, and the failure to overwrite an existing vsam data set is the next cause of PTF test case errors.

I'll likely need to undo this in v3 as well, or we can try to handle VSAM data sets differently. ~I get `EDC5041I An error was detected at the system level when opening a file.` when trying to open VSAM with `cat`, which is a different error when a data set doesn't exist, but idk if it's reliable across all systems. I'd prefer a better way of detecting VSAM from USS, like `tsocmd "listds ..."`. Any reason why we cant use `listds` for VSAM or as a backup to `cat`?~

Can we use `tso_is_data_set_exists` instead?

bin/commands/init/vsam/index.sh
```diff
@@ -75,13 +75,20 @@ fi
 # FIXME: cat cannot be used to test VSAM data set
 vsam_existence=$(is_data_set_exists "${vsam_name}")
```
Martin-Zeithaml: Yes, I agree, `tso_is_data_set_exists` resolves the problem and we can return the code back.

PR #4547, opened 2025-10-09
Prepare 2.18.3-RC2

MarkAckert:
Includes changes from #4542 which fixes a bug in the `init vsam` command group and additionally fixes PTF tests in CICD.


Martin-Zeithaml (approved):
Thanks for checking VSAM existence!

PR #4554, opened 2025-10-16
zwe validate certificate

1000TurquoisePogs:
This PR utilizes apiml's `certificate-analyser.jar` to check zowe's keystore/truststore and report action items for specific situations.

It's run at startup of zowe, but also via the command `zwe validate certificate`.

This command can be run with argument `-e STRICT` to change the lEvel of verification to STRICT, for example. This allows users to see what would happen with different levels of strictness.

The command might not run well when run standalone, because often keyrings aren't readable by the user who runs `zwe`. But, it will have the right conditions to be a startup check.

This command can be disabled with `zowe.launchScript.startupChecks.certificate: false`.

This command does not verify ICSF certificates, or at least it's very likely not to. It seems java 11+ doesnt ship with the right security config that 8 had, so when you request to read these, it just tells you JCECCARACFKS isn't available. It then becomes some administrative task to customize the java 11+ config? Not a problem I can solve.


bin/commands/validate/certificate/index.ts
```diff
@@ -0,0 +1,380 @@
...
+    `-t ${truststoreLocation} -tt ${truststoreType} -tp ${truststorePass} ` +
+    `-d ${ZOWE_CONFIG.zowe.externalDomains.join(',')}`;
+  
+  const result = shell.execOutErrSync('java', ...argsString.split(' '));
+  const rc = result.rc;
+  
+  let configLines = [];
+  let output = '';
+  if (result.out) {
+    configLines.concat(result.out.split('\n'));
```
Martin-Zeithaml: `concat` returns merged array, it is not changing `configLines`. Same on L98.
`configLines = configLines.concat(result.out.split('\n'));`

bin/commands/validate/certificate/index.ts
```diff
@@ -0,0 +1,380 @@
...
+
+  let verifyCertificates = level;
+  if (verifyCertificates) {
+    verifyCertificates.toUpperCase();
+  }
+  if (verifyCertificates != 'NONSTRICT' && verifyCertificates != 'STRICT' && verifyCertificates != 'DISABLED') {
+    verifyCertificates = ZOWE_CONFIG.zowe.verifyCertificates;
+  }
+
+  java.requireJava();
```
Martin-Zeithaml, line 50: Unsupported Java used, detected, but continues to run `certificate-analyser.jar`
```
2025-12-05 11:09:49 <ZWELS:33883644> USERID INFO (validateJavaHome) Java version: 1.8.0_461
ERROR: Java 1.8.0_461 is less than the minimum level required of Java 17.
Error: A JNI error has occurred, please check your installation and try again
Exception in thread "main" java.lang.UnsupportedClassVersionError: JVMCFRE199E bad major version 61.0 of class=org/zowe/apiml/Analyser, the maximum supported major version is 52.0; offset=6
        at java.lang.ClassLoader.defineClassImpl(Native Method)
        at java.lang.ClassLoader.defineClassInternal(ClassLoader.java:399)
        at java.lang.ClassLoader.defineClass(ClassLoader.java:360)
        at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:154)
        at java.net.URLClassLoader.defineClass(URLClassLoader.java:730)
        at java.net.URLClassLoader.access$400(URLClassLoader.java:96)
        ...
```

1000TurquoisePogs: we need to add a `zwe validate java` and overhaul requireJava() to obey the whole bypass/warn/exit behavior

bin/commands/validate/certificate/index.ts
```diff
@@ -0,0 +1,380 @@
...
+        keyringUser = truststoreLocation.split('://')[1].split('/')[0];
+      } catch (e) {
+        common.printFormattedWarn(common.MSG_KEY, COMMAND_NAME, `Unexpected keyring format ${truststoreLocation}`);
+      }
+    }
+  } else if (keystoreType != 'PKCS12') {
+    common.printFormattedWarn(common.MSG_KEY, COMMAND_NAME, `Truststore unknown type ${truststoreType}`);
+  }
+  
+  const argsString = `-Djava.protocol.handler.pkgs=com.ibm.crypto.provider -jar ${ZOWE_CONFIG.zowe.runtimeDirectory}/bin/utils/certificate-analyser.jar `+
```
Martin-Zeithaml: `zowe.runtimeDirectory` is ok when running this as pre-flight check.
But when running as `zwe` command from shell, `zowe.runtimeDirectory` might be `undefined` or points to a Zowe version without `certificate-analyzer`.

`/this/pr/bin/zwe validate certificate -c configForZowe3.0.0.yaml` fails as there is no `certificate-analyzer.jar` yet.

1000TurquoisePogs: I see. OK, I'll see if I can do this by relative path, if not then I will check that runtimeDirectory is defined.

bin/commands/validate/certificate/index.ts
```diff
@@ -0,0 +1,380 @@
...
+
+  let certificateInvalid = VALIDATION_OK;
+  let configInvalid = VALIDATION_OK;
+  
+  if (rc == 0) {
+    common.printDebug(configLines.filter(line => !line.startsWith(SECTION_SEPARATOR)).join('\n'));
+    common.printFormattedInfo(common.MSG_KEY, COMMAND_NAME, "Certificate checks passed.");
+  } else {
+
+
```
Martin-Zeithaml, line 149: This case seems to be unmatched:
```yaml
zowe:
  certificate:
    keystore:
      type: "JCERACFKS"
      file: "safkeyringasdf://USR/RING"
```

```
unknown protocol: safkeyringjceasdf
```
The schema pattern allows anything lowercased after `safkeyring`:
```
"pattern": "^safkeyring[a-z]*:\/\/.*"
```

bin/commands/validate/certificate/index.ts
```diff
@@ -0,0 +1,380 @@
...
+    
+    // From R_datalib (IRRSDL00)
+    // Format:
+    //   1) R_datalib (IRRSDL00) error: profile for ring not found (8, 8, 84)
+    //   2) R_datalib (IRRSDL00) error: not RACF authorized to use the requested service (8, 8, 8)
+    if (output.includes("IRRSDL00")) {
+      configInvalid = VALIDATION_ERROR;
+
+      const irrLine = configLines.filter((line)=> line.includes('IRRSDL00'))[0];
```
Martin-Zeithaml: That index at the end will save char at position 0:
```
configLines = 'R_datalib (IRRSDL00) error: not RACF authorized to use the requested service (8, 8, 8)'
const irrLine = configLines.filter((line)=> line.includes('IRRSDL00'))[0];
console.log(`irrLine = ${irrLine}`)
```
`irrLine = R`

bin/commands/validate/certificate/index.ts
```diff
@@ -0,0 +1,380 @@
...
+      const codes = irrLine.substring(0,irrLine[0].length-1).split(':')[1].split('(')[1].split(', ');
+      const esm = zos.getEsm();
+      common.printFormattedError(common.MSG_KEY, COMMAND_NAME, `Could not load keyring. SAF rc=${codes[0]}, ${esm} rc=${codes[1]}, ${esm} rsn=${codes[2]}`);
+      common.printFormattedError(common.MSG_KEY, COMMAND_NAME, `Verify that keystore and truststore are valid and accessible to the Zowe STC.`);
+      common.printFormattedError(common.MSG_KEY, COMMAND_NAME, `Review Zowe YAML properties:`);
+      common.printFormattedError(common.MSG_KEY, COMMAND_NAME, `  zowe.certificate.keystore.file: ${ZOWE_CONFIG.zowe.certificate.keystore.file}`);
+      common.printFormattedError(common.MSG_KEY, COMMAND_NAME, `  zowe.certificate.truststore.file: ${ZOWE_CONFIG.zowe.certificate.truststore.file}`);
+      if (codes[2] == 8) {
```
Martin-Zeithaml: `codes` is array of strings, not numbers. We have to compare proper types.

bin/commands/validate/certificate/index.ts
```diff
@@ -0,0 +1,380 @@
...
+
+  java.requireJava();
+
+  const keystoreType = ZOWE_CONFIG.zowe.certificate.keystore.type;
+  const keystoreAlias = ZOWE_CONFIG.zowe.certificate.keystore.alias;
+  const keystorePass = ZOWE_CONFIG.zowe.certificate.keystore.password;
+  const truststoreType = ZOWE_CONFIG.zowe.certificate.truststore.type;
+  const truststorePass = ZOWE_CONFIG.zowe.certificate.truststore.password;
+  let keystoreLocation = ZOWE_CONFIG.zowe.certificate.keystore.file;
+  let truststoreLocation = ZOWE_CONFIG.zowe.certificate.truststore.file;
```
Martin-Zeithaml, line 67: Would be probably fine for runtime, but might fail, when used as command from shell for testing the current configuration.

`zowe.externalDomains` is also expected as always defined.

bin/commands/validate/certificate/index.ts
```diff
@@ -0,0 +1,380 @@
...
+
+/*
+  This program uses certificate-analyser.jar from the zowe apiml repository.
+  It doesnt work on java 8.
+  That program is currently capable of verifying that the zowe certificate is valid for zowe,
+  And can verify that a remote server is trusted by zowe's truststore.
+
+  It does not seem like the program has extensive CA chain review capability,
+  And it's unclear if this program can handle ICSF certificates.
+  It's also unclear how to detect something is an ICSF certificate, to avoid checking something we cannot verify anyway.
```
Martin-Zeithaml, line 36: Would the type help?
```
# Type of certificate storage. Valid values are: PKCS12, JCERACFKS. APIML additionally supports: JCEKS, JCECCAKS, JCECCARACFKS, or JCEHYBRIDRACFKS
```
Seems like `*CCA*` might be the ICSF keyring.
If type JCECCAKS or JCECCARACFKS -> ICSF.

bin/commands/validate/certificate/index.ts
Martin-Zeithaml: Relative path will not help. It has to be absolute.

bin/commands/validate/certificate/index.ts
```diff
@@ -0,0 +1,380 @@
...
+    
+    // From R_datalib (IRRSDL00)
+    // Format:
+    //   1) R_datalib (IRRSDL00) error: profile for ring not found (8, 8, 84)
+    //   2) R_datalib (IRRSDL00) error: not RACF authorized to use the requested service (8, 8, 8)
+    if (output.includes("IRRSDL00")) {
+      configInvalid = VALIDATION_ERROR;
+
+      const irrLine = configLines.filter((line)=> line.includes('IRRSDL00'))[0];
+      const codes = irrLine.substring(0,irrLine[0].length-1).split(':')[1].split('(')[1].split(', ');
```
Martin-Zeithaml:
```
console.log(codes);
```
```
8,8,8)
```

bin/commands/validate/certificate/index.ts
```diff
@@ -0,0 +1,391 @@
...
+  let truststoreLocation = ZOWE_CONFIG.zowe.certificate.truststore.file;
+
+  let keyringUser;
+  if (keystoreType.startsWith('JCE') && keystoreType.endsWith('RACFKS')) {
+    keystoreLocation = keystoreLocation.replace('safkeyring', 'safkeyring'+keystoreType.toLowerCase().substring(0, keystoreType.length - 'RACFKS'.length));
+    try {
+      keyringUser = keystoreLocation.split('://')[1].split('/')[0];
+    } catch (e) {
+      common.printFormattedWarn(common.MSG_KEY, COMMAND_NAME, `Unexpected keyring format ${keystoreLocation}`);
+    }
```
Martin-Zeithaml, line 76: We might consider to check old way of `safkeyring:////USER/RING`

The problem is, that `certificate-analyser` accepts `safkeyring://` or `safkeyring:////`.

pavel-jares-bcm: What about use the same regex pattern as APIML? `^(safkeyring[^:]*):/{2,4}([^/]+)/([^/]+)$`

Martin-Zeithaml: From the schema point of view, only `://` is correct:
https://github.com/zowe/zowe-install-packaging/blob/f7603a51cb8bfff8c4965deb156eabd4e9d36839/schemas/zowe-yaml-schema.json#L997

IMHO the `safkeyring:////USER/RING` should be recognized as invalid.

And btw the regex is not valid:

```
Digital ring information for user TEST001:
 
   Ring:
        >///<
   *** No certificates connected ***
```

So this ugly, but valid key ring and should work with Zowe: `safkeyring://TEST001////`.

bin/commands/validate/certificate/index.ts
pavel-jares-bcm: The verification is done base on public key. So it shouldn't be problem. An isue could happen just on starting a webserver (see -l/--local and -r/--remoteurl if we need a client certificate).

Anyway to be working on ICSF it is necessary to configure `jvm.security`. It is possible to follow steps in https://github.com/zowe/api-layer/pull/4354/changes (see `JVM_SECURITY_PROPERTIES_OVERRIDE` and `jvm.security.override.properties`)

bin/commands/validate/certificate/index.ts
```diff
@@ -0,0 +1,391 @@
...
+  } else if (keystoreType != 'PKCS12') {
+    common.printFormattedWarn(common.MSG_KEY, COMMAND_NAME, `Keystore unknown type ${keystoreType}`);
+  }
+  if (truststoreType.startsWith('JCE') && truststoreType.endsWith('RACFKS')) {
```
pavel-jares-bcm, line 85: Truststore and keystore processing should be the same. What about refactoring as one common method?

bin/commands/validate/certificate/index.ts
```diff
@@ -0,0 +1,391 @@
...
+  
+  if (rc == 0) {
+    common.printDebug(configLines.filter(line => !line.startsWith(SECTION_SEPARATOR)).join('\n'));
+    common.printFormattedInfo(common.MSG_KEY, COMMAND_NAME, "Certificate checks passed.");
+  } else {
+
+
+    // From cert-analyser codebase
+    // Only happens if keyring pattern is not matched, should be impossible due with proper schema checking.
+    if (output.includes('Incorrect key ring format')) {
```
pavel-jares-bcm, line 152: If I understand well this is rewriting the analyzer response. What about to update rather in there and the just display it?

Martin-Zeithaml: The goal is to detect an error and provide possible solution. Any update to analyzer and simplification of this code is welcome.

bin/commands/validate/certificate/index.ts
```diff
@@ -0,0 +1,391 @@
...
+      }
+    }
+  } else if (keystoreType != 'PKCS12') {
+    common.printFormattedWarn(common.MSG_KEY, COMMAND_NAME, `Truststore unknown type ${truststoreType}`);
+  }
+  
+  const argsString = `-Djava.protocol.handler.pkgs=com.ibm.crypto.provider -jar ${std.getenv('ZWE_zowe_runtimeDirectory')}/bin/utils/certificate-analyser.jar ` +
+    `-k ${keystoreLocation} -kt ${keystoreType} -kp ${keystorePass} -a ${keystoreAlias} ` +
+    `-t ${truststoreLocation} -tt ${truststoreType} -tp ${truststorePass} ` +
+    `-d ${ZOWE_CONFIG.zowe.externalDomains.join(',')}`;
```
Martin-Zeithaml: https://github.com/zowe/api-layer/issues/4486

When the keystore = truststore, we can use `certificate-analyser`
When the keystore != truststore (unusual, but valid), we can't verify - I did not test this situation yet.

bin/commands/validate/certificate/index.ts
1000TurquoisePogs: Check added, but I'm not going to make a schema change to permit something wild that could result in code bugs later.

bin/commands/validate/certificate/index.ts
1000TurquoisePogs: certificate-analyzer can be general-purpose - people can use it outside of zowe.

zwe validate is specific purpose. it has specific output to help guide users, talking about zowe YAML for example.

bin/commands/validate/certificate/index.ts
1000TurquoisePogs: Hm ok. @pavel-jares-bcm  is this a limitation you are going to fix?

I think the zwe code is doing the right thing so not sure if we should change it here.

bin/commands/validate/certificate/index.ts
```diff
@@ -0,0 +1,415 @@
...
+  } else if (truststoreType != 'PKCS12') {
+    common.printFormattedWarn(common.MSG_KEY, COMMAND_NAME, `Truststore unknown type ${truststoreType}`);
+  }
+  
+  const argsString = `-Djava.protocol.handler.pkgs=com.ibm.crypto.provider -jar ${std.getenv('ZWE_zowe_runtimeDirectory')}/bin/utils/certificate-analyser.jar ` +
+    `-k ${keystoreLocation} -kt ${keystoreType} -kp ${keystorePass} -a ${keystoreAlias} ` +
+    `-t ${truststoreLocation} -tt ${truststoreType} -tp ${truststorePass} ` +
+    `-d ${ZOWE_CONFIG.zowe.externalDomains.join(',')}`;
+  
+  const result = shell.execOutErrSync('java', ...argsString.split(' '));
```
Martin-Zeithaml: See https://github.com/zowe/zowe-install-packaging/pull/4714

Martin-Zeithaml (changes requested):
Missing Changelog

Martin-Zeithaml (approved):

PR #4562, opened 2025-11-03
Check app-server

Martin-Zeithaml:
For #4307.

Seems `ZWE_ENABLED_COMPONENTS` is not initialized when running `zwe` command.

```yaml
# ./yaml/zowe.basic.yaml
zowe:
  setup:
    dataset:
      prefix: ZOWE.PR#4307
      parmlib: ZOWE.PR#4307.PARMLIB
      jcllib: ZOWE.PR#4307.JCLLIB
      authLoadlib: ZOWE.PR#4307.ZWESALL
      authPluginLib: ZOWE.PR#4307.ZWESAPL
  externalDomains:
    - zos.example.com
zOSMF:
  host: zosmf.example.com
  port: 443
components:
  app-server:
    enabled: true
    port: 1234

```

* `node` not set in configuration
* `app-server` enabled
* Environment variable `NODE_HOME` is set

Missing `node` in check:

```
$ zwe init -c ./yaml/zowe.basic.yaml
===============================================================================
>> CONFIGURE ZOWE

>> Check if need to update runtime directory, Java and/or node.js settings in Zowe YAML configuration

These configurations need to be added to your YAML configuration file:

zowe:
  runtimeDirectory: "/zowe/4307"
java:
  home: "/java/current"

>> Please manually update "./yaml/zowe.basic.yaml" before you start Zowe.
```


MarkAckert:
I'm looking into why `ZWE_ENABLED_COMPONENTS` isn't being set.

MarkAckert:
This is a bug separate from the upstream Node PR, and it exists in the shell-only code path.

`ZWE_ENABLED_COMPONENTS` is not set globally on startup in the shell scripts, and if we want to use it in `init / index.sh`, we need to call:

```shell
export ZWE_INSTALLED_COMPONENTS="$(find_all_installed_components)"
export ZWE_ENABLED_COMPONENTS="$(find_all_enabled_components)"
```

However.....`ZWE_ENABLED_COMPONENTS` does not work the way you might think in shell scripts; it first lists components available under the `components` dir (`ZWE_INSTALLED_COMPONENTS`) and then filters them against environment variables to determine enablement status. If a component is enabled in zowe.yaml / env and not present in the installation directory (happens in my test environment), it will not show up as enabled. 

The configmgr code works completely differently: it strictly reads the zowe.yaml to retrieve `ZWE_ENABLED_COMPONENTS` without cross-referencing the installed components under the `components/` dir.

PR #4563, opened 2025-11-03
Change zosmf authentication scheme based on apiml auth provider

js665999:
Set z/OSMF authentication provider in static definition based on authentication provider.

If provider 'zosmf' set scheme to 'zosmf'
If provider 'saf' set scheme to 'httpBasicPassTicket'
Add zowe.yaml variable to allow manual overwrite

This change is in preparation for change of default provider to 'saf' coming in later version and will allow z/OSMF REST APIs to function with provider set to SAF. Pass Tickets needs to be setup.

Depends on https://github.com/zowe/api-layer/pull/4374


dkelosky:
In the past, I think the `apimlAuthenticationToken` wrapped the z/OSMF JWT / LTPA2.  Then, when requests are routed to z/OSMF via API ML, the z/OSMF JWT / LTPA2 was used.  With this new approach, will clients (like Zowe CLI & Zowe Explorer) be able to seamlessly invoke z/OSMF REST APIs via API ML?  

Secondly, does `saf` provider imply that API ML requires any extra privileges?  For example, does API ML rely on ZSS or some other privilege service to authentication of mainframe users?

js665999:
> In the past, I think the `apimlAuthenticationToken` wrapped the z/OSMF JWT / LTPA2. Then, when requests are routed to z/OSMF via API ML, the z/OSMF JWT / LTPA2 was used. With this new approach, will clients (like Zowe CLI & Zowe Explorer) be able to seamlessly invoke z/OSMF REST APIs via API ML?
> 
> Secondly, does `saf` provider imply that API ML requires any extra privileges? For example, does API ML rely on ZSS or some other privilege service to authentication of mainframe users?

If SAF is used as authentication provider, Zowe will act as an auth service and will generate JWT instead of using the z/OSMF JWT. So until now z/OSMF APIs did not work with SAF option. This PR fixes that.
No new privileges needed, SAF provider was always an option and it uses SAF APIs directly.

dkelosky:
Thanks -

> So until now z/OSMF APIs did not work with SAF option. This PR fixes that.

Can you elaborate slightly?  What auth details are sent to z/OSMF?

js665999:
> Thanks -
> 
> > So until now z/OSMF APIs did not work with SAF option. This PR fixes that.
> 
> Can you elaborate slightly? What auth details are sent to z/OSMF?

Zowe would send its own generated JWT with the request. This would not work with z/OSMF. 
 With the new option it will send username and passticket.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ All notable changes to the Zowe Installer will be documented in this file.
 
 ## `3.4.0`
 
+- Enhancement: Configure z/OSMF static API definiton based on authentication provider [#4563] tps://github.com/zowe/zowe-install-packaging/pull/4563
```
Martin-Zeithaml: `http` is malformed.

dkelosky:
Will this require additional setup for users that want to onboard z/OSMF: https://docs.zowe.org/stable/user-guide/api-mediation/configuration-extender-passtickets/#configuring-zowe-to-use-passtickets ?

Secondly, is there any concern on additional overhead in requests?  For example, in the past, when using LTPA/JWT for z/OSMF requests, requests are process faster without needing to go to the SAF layer by validating LTPA.

achmelo:
> Will this require additional setup for users that want to onboard z/OSMF: https://docs.zowe.org/stable/user-guide/api-mediation/configuration-extender-passtickets/#configuring-zowe-to-use-passtickets ?
> 
> Secondly, is there any concern on additional overhead in requests? For example, in the past, when using LTPA/JWT for z/OSMF requests, requests are process faster without needing to go to the SAF layer by validating LTPA.

We can mention this as a note in the documentation, but it is not a new requirement. Generating passtickets for zOSMF APPLID is very common for Zowe installations.

The option with SAF as an authentication provider is mainly for users who don't have or want to use zOSMF. This PR addresses edge cases where users still want to use zOSMF REST APIs, but not for authentication. The only way API ML can authenticate a user against zOSMF is to issue a passticket(basic auth). LTPA tokens are not available in this context, only API ML-issued JWTs, client certificates, or PATs.

dkelosky:
Thank you @js665999 & @achmelo for answering questions

bin/libs/component.ts
```diff
@@ -494,6 +494,24 @@ export function processComponentApimlStaticDefinitions(componentDir: string): bo
...
+            authScheme = zosmfAuthenticationScheme;
+          } else if (AuthenticationschemeEnv) {
+            authScheme = AuthenticationschemeEnv;
+          } else if (
+            apimlAuthProviderEnv === 'saf' || gatewayAuthProviderEnv === 'saf' ) {
+            authScheme = 'httpBasicPassTicket';
+          }
+          std.setenv('ZOSMF_AUTHENTICATION_SCHEME', authScheme);
+
+
```
pavel-jares-bcm:
```suggestion
          const zosmfAuthenticationScheme = std.getenv("ZOSMF_AUTHENTICATION_SCHEME") ?? std.getenv("ZWE_zOSMF_authentication_scheme");
          const authProvider = std.getenv("ZWE_components_apiml_apiml_security_auth_provider") ?? std.getenv("ZWE_components_gateway_apiml_security_auth_provider");

          let authScheme = "zosmf";
          if (!zosmfAuthenticationScheme && (authProvider === 'saf')) {
            authScheme = 'httpBasicPassTicket';
          }
          std.setenv('ZOSMF_AUTHENTICATION_SCHEME', authScheme);

```

schemas/zowe-yaml-schema.json
```diff
@@ -796,6 +796,15 @@
         "applId": {
           "type": "string",
           "description": "Appl ID of your z/OSMF instance."
+        },
+        "authentication": {
+          "type": "object",
+          "properties": {
+            "scheme": {
+              "type": "string",
+              "description": "Authentication scheme for z/OSMF."
```
pavel-jares-bcm:
```suggestion
              "description": "Authentication scheme for z/OSMF.",
              "enum": ["httpBasicPassTicket", "zosmf"]
```

I guess we can validate the values itself. No other string makes sense.

zFernand0:
> > Will this require additional setup for users that want to onboard z/OSMF: https://docs.zowe.org/stable/user-guide/api-mediation/configuration-extender-passtickets/#configuring-zowe-to-use-passtickets ?
> > Secondly, is there any concern on additional overhead in requests? For example, in the past, when using LTPA/JWT for z/OSMF requests, requests are process faster without needing to go to the SAF layer by validating LTPA.
> 
> We can mention this as a note in the documentation, but it is not a new requirement. Generating passtickets for zOSMF APPLID is very common for Zowe installations.
> 
> The option with SAF as an authentication provider is mainly for users who don't have or want to use zOSMF. This PR addresses edge cases where users still want to use zOSMF REST APIs, but not for authentication. The only way API ML can authenticate a user against zOSMF is to issue a passticket(basic auth). LTPA tokens are not available in this context, only API ML-issued JWTs, client certificates, or PATs.

> Thank you @js665999 & @achmelo for answering questions


I guess the overhead/performance question sill stands.

As I understand this PR, the setting will go out with a default value of z/OSMF (to avoid breaking changes). That's awesome! 🙏 

However,  I wonder what happens as more people start to consider using SAF as their authentication provider.
The APIML requests that go to z/OSMF (e.g. to retrieve jobs and data sets) still need to be authenticated against z/OSMF and have a passticket issued. Is that correct?

And if so, does that mean that people using z/OSMF via APIML (and decide to use SAF as their authentication provider) may see a slight delay in their responses due to the roundtrip getting a little longer? 

My assumption could be wrong, but here is how I'm picturing the request/roundtrip...
- Before SAF authentication provider:
  - `client (ZE/CLI) > APIML > z/OSMF > APIML > client`
- After SAF authentication provider:
  - `client (ZE/CLI) > APIML > SAF > APIML > z/OSMF > APIML > client`
  
  Again, apologies if I'm misunderstanding something here. 🙏

achmelo:
> > > Will this require additional setup for users that want to onboard z/OSMF: https://docs.zowe.org/stable/user-guide/api-mediation/configuration-extender-passtickets/#configuring-zowe-to-use-passtickets ?
> > > Secondly, is there any concern on additional overhead in requests? For example, in the past, when using LTPA/JWT for z/OSMF requests, requests are process faster without needing to go to the SAF layer by validating LTPA.
> > 
> > 
> > We can mention this as a note in the documentation, but it is not a new requirement. Generating passtickets for zOSMF APPLID is very common for Zowe installations.
> > The option with SAF as an authentication provider is mainly for users who don't have or want to use zOSMF. This PR addresses edge cases where users still want to use zOSMF REST APIs, but not for authentication. The only way API ML can authenticate a user against zOSMF is to issue a passticket(basic auth). LTPA tokens are not available in this context, only API ML-issued JWTs, client certificates, or PATs.
> 
> > Thank you @js665999 & @achmelo for answering questions
> 
> I guess the overhead/performance question sill stands.
> 
> As I understand this PR, the setting will go out with a default value of z/OSMF (to avoid breaking changes). That's awesome! 🙏
> 
> However, I wonder what happens as more people start to consider using SAF as their authentication provider. The APIML requests that go to z/OSMF (e.g. to retrieve jobs and data sets) still need to be authenticated against z/OSMF and have a passticket issued. Is that correct?
> 
> And if so, does that mean that people using z/OSMF via APIML (and decide to use SAF as their authentication provider) may see a slight delay in their responses due to the roundtrip getting a little longer?
> 
> My assumption could be wrong, but here is how I'm picturing the request/roundtrip...
> 
>     * Before SAF authentication provider:
>       
>       * `client (ZE/CLI) > APIML > z/OSMF > APIML > client`
> 
>     * After SAF authentication provider:
>       
>       * `client (ZE/CLI) > APIML > SAF > APIML > z/OSMF > APIML > client`
>       
>       Again, apologies if I'm misunderstanding something here. 🙏

Although we don't have the data, any delay should be negligible. It may be even faster in some cases, as there is no need to validate JWT with a REST call to zOSMF.
With zOSMF auth provider -> 
`client (ZE/CLI) > APIML > validate JWT with zOSMF > request zOSMF with JWT(or ltpa wrapped in JWT) from client > APIML > client`

With SAF auth provider -> 
`client (ZE/CLI) > APIML > validate JWT locally > generate passticket > request zOSMF with username/passticket > APIML > client`

zFernand0:
> Although we don't have the data, any delay should be negligible. It may be even faster in some cases, as there is no need to validate JWT with a REST call to zOSMF. With zOSMF auth provider -> `client (ZE/CLI) > APIML > validate JWT with zOSMF > request zOSMF with JWT(or ltpa wrapped in JWT) from client > APIML > client`
> 
> With SAF auth provider -> `client (ZE/CLI) > APIML > validate JWT locally > generate passticket > request zOSMF with username/passticket > APIML > client`

I see what you mean that there may be cases where it could potentially be faster since the extra steps happen without z/OSMF 😅

bin/libs/component.ts
```diff
@@ -494,6 +494,15 @@ export function processComponentApimlStaticDefinitions(componentDir: string): bo
           std.setenv('ZOSMF_NON_SECURE_PORT_ENABLED', `${nonSecurePortEnabled}`);
           std.setenv('ZOSMF_SECURE_PORT_ENABLED', `${securePortEnabled}`);
 
+          const zosmfAuthenticationScheme = std.getenv("ZOSMF_AUTHENTICATION_SCHEME") ?? std.getenv("ZWE_zOSMF_authentication_scheme");
```
Martin-Zeithaml: I think, this will not work:
* let's assume `ZOSMF_AUTHENTICATION_SCHEME=` and `ZWE_zOSMF_authentication_scheme=zosmf`
* `"" ?? zosmf` is `""`

There must be a check for empty string.

Martin-Zeithaml (approved):

bin/libs/component.ts
```diff
@@ -494,6 +494,15 @@ export function processComponentApimlStaticDefinitions(componentDir: string): bo
           std.setenv('ZOSMF_NON_SECURE_PORT_ENABLED', `${nonSecurePortEnabled}`);
           std.setenv('ZOSMF_SECURE_PORT_ENABLED', `${securePortEnabled}`);
 
+          const zosmfAuthenticationScheme = ([ "zosmf", "httpBasicPassTicket" ].includes(std.getenv("ZOSMF_AUTHENTICATION_SCHEME"))) ? std.getenv("ZOSMF_AUTHENTICATION_SCHEME") : std.getenv("ZWE_zOSMF_authentication_scheme");
+          const authProvider = std.getenv("ZWE_components_apiml_apiml_security_auth_provider") ?? std.getenv("ZWE_components_gateway_apiml_security_auth_provider");
```
zFernand0: Q: Should the same validation happen here?
Either stronger validation, or use `||` which will evaluate `""` to falsy 🙏 

```suggestion
          const authProvider = std.getenv("ZWE_components_apiml_apiml_security_auth_provider") || std.getenv("ZWE_components_gateway_apiml_security_auth_provider");
```

MarkAckert: It's not described in the schema as far as I can see, but the `ZWE_components_gateway_apiml_security_auth_provider` _must_ be set Zowe config, and to date we've guaranteed it's set through `defaults.yaml`. Still, stronger validation and an enumeration of acceptable values sounds like a good idea to me. I'll add this in if no one objects...

MarkAckert:
Build failure is unrelated to this change; merging in.

PR #4565, opened 2025-11-03
enable apiml by default

achmelo:

#### PR type
- [ ] Bugfix 
- [ ] Feature 

#### Relevant issues

Fixes 

- 

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

---

---
Enable API ML modulith runtime as default.
---

---
_multi-line description_

Doc issue/PR number:


muzikovam (approved):

Martin-Zeithaml:
It's just a detail, but...
Enabled by default means it is enabled in [defaults.yaml](https://github.com/zowe/zowe-install-packaging/blob/2de92bc6c7cf5dba1ac7ac02acf341c9b0314e93/files/defaults.yaml#L157).

This PR enables APIML in example and PSWI examples.

Missing update in CHANGELOG, where it should be described clearly.

muzikovam:
Also I might not understand but it looks like the user doesn't have a chance to change if apiml is enabled or not as the "true" is hardcoded.

achmelo:
> It's just a detail, but... Enabled by default means it is enabled in [defaults.yaml](https://github.com/zowe/zowe-install-packaging/blob/2de92bc6c7cf5dba1ac7ac02acf341c9b0314e93/files/defaults.yaml#L157).
> 
> This PR enables APIML in example and PSWI examples.
> 
> Missing update in CHANGELOG, where it should be described clearly.

The intention was to enable it only for new installations. In case it's "true" in defaults.yaml, would it be enabled also for those who upgrade?

Martin-Zeithaml:
> Also I might not understand but it looks like the user doesn't have a chance to change if apiml is enabled or not as the "true" is hardcoded.

User has always option to configure what ever is needed. The defaults.yaml file is merged with user's configurations, where the left most entries are higher priority:
```
CONFIG=FILE(/zowe/userZowe.yaml):PARMLIB(ZOWE.PARM(YAML))
```

The merged config is result of `FILE(/zowe/userZowe.yaml)` + `PARMLIB(ZOWE.PARM(YAML))` + `/zowe/runtime/files/defaults.yaml`

The default values are use only and only if not used by previous entries.

achmelo:
> > Also I might not understand but it looks like the user doesn't have a chance to change if apiml is enabled or not as the "true" is hardcoded.
> 
> User has always option to configure what ever is needed. The defaults.yaml file is merged with user's configurations, where the left most entries are higher priority:
> 
> ```
> CONFIG=FILE(/zowe/userZowe.yaml):PARMLIB(ZOWE.PARM(YAML))
> ```
> 
> The merged config is result of `FILE(/zowe/userZowe.yaml)` + `PARMLIB(ZOWE.PARM(YAML))` + `/zowe/runtime/files/defaults.yaml`
> 
> The default values are use only and only if not used by previous entries.

In that case, if the user doesn't have apiml.enabled at all in zowe.yaml, this change would enable it. Did I get it right?

Martin-Zeithaml:
> In that case, if the user doesn't have apiml.enabled at all in zowe.yaml, this change would enable it. Did I get it right?

No, let's consider this yaml, no `components` section:
```yaml
zowe:
  setup:
    dataset:
      prefix: TEST
```
But if you ask for `components` section, you will get this output (it continues):
```
$ /4565/bin: ./zwe config get -c ./zowe.yaml -p .components
Temporary directory '/tmp/.zweenv-8046' created.
{
  "apiml": {
    "enabled": false,
    "debug": false
  },
  "gateway": {
    "enabled": true,
    "port": 7554,
    "debug": false,
...
```

What is confusing here:
* `zwe config get -c ./zowe.yaml -p .components` 
* Is actually internally `zwe config get -c FILE(./zowe.yaml):FILE(/4565/files/defaults.yaml) -p .components`

This PR enables APIML, if user starts with `example-zowe.yaml` - if this was intention, this PR is fine.

But generally, the default setting is `components.apiml.enabled=false` - to change it, you have to update `files/defaults.yaml` - but this will be the default behavior for new installation and upgrade too.

MarkAckert:
As I understand this, I agree with making the change in `example-zowe.yaml` for 3.x, and moving this to `defaults.yaml` in 4.x. 

Reasoning:
If a user with Zowe v3.0 or v3.1 upgrades to 3.4+, they should experience the same runtime behavior as before (individual services).

We also need to update the modulith test suite in this PR before merging in, changing it to instead use `apiml.enabled=false`. I'll work on this.

MarkAckert:
The APIML only workflow behaves differently from ZWECONF - the user is not given the choice to customize the modulith enablement or debug settings. Is this intended?

achmelo:
> > In that case, if the user doesn't have apiml.enabled at all in zowe.yaml, this change would enable it. Did I get it right?
> 
> No, let's consider this yaml, no `components` section:
> 
> ```yaml
> zowe:
>   setup:
>     dataset:
>       prefix: TEST
> ```
> 
> But if you ask for `components` section, you will get this output (it continues):
> 
> ```
> $ /4565/bin: ./zwe config get -c ./zowe.yaml -p .components
> Temporary directory '/tmp/.zweenv-8046' created.
> Zowe will remove it on success, but if zwe exits with a non-zero code manual cleanup would be needed.
> {
>   "apiml": {
>     "enabled": false,
>     "debug": false
>   },
>   "gateway": {
>     "enabled": true,
>     "port": 7554,
>     "debug": false,
> ...
> ```
> 
> What is confusing here:
> 
>     * `zwe config get -c ./zowe.yaml -p .components`
> 
>     * Is actually internally `zwe config get -c FILE(./zowe.yaml):FILE(/4565/files/defaults.yaml) -p .components`
> 
> 
> This PR enables APIML, if user starts with `example-zowe.yaml` - if this was intention, this PR is fine.
> 
> But generally, the default setting is `components.apiml.enabled=false` - to change it, you have to update `files/defaults.yaml` - but this will be the default behavior for new installation and upgrade too.

`This PR enables APIML, if user starts with example-zowe.yaml - if this was intention, this PR is fine.` Yes, this was the intention of this PR. We don't wont this to have effect on any exising installation.

achmelo:
> The APIML only workflow behaves differently from ZWECONF - the user is not given the choice to customize the modulith enablement or debug settings. Is this intended?

I think this is correct. This is a simplified workflow where we are giving users recommended configuration. For any other setup, they should use the standrad workflow.

MarkAckert (approved):

PR #4572, opened 2025-11-10
V2: replace `which` with `whence`

Martin-Zeithaml:
There is no `which` command in Unix System Services. Used code always produced no string as result:
```
$ which java
which: FSUM7351 not found
$ which java 2>/dev/null
```

* `bin/libs/java.sh`
  * Comment update
  * `which` to `whence`
  * removed the second `which` (copy & paste bug?)

* `bin/libs/node.sh`
  * `which` to `whence`

For V3, this is fixed in #4307.

### Quick test
```shell
#!/bin/sh
export ZWE_zowe_runtimeDirectory=/zowe/4572
. /zowe/4572/bin/libs/index.sh

whence java
detect_java_home
```

```
/sys/java64bt/v17r0m0/usr/lpp/java/current/bin/java
DEBUG: after whence: /sys/java64bt/v17r0m0/usr/lpp/java/current
/sys/java64bt/v17r0m0/usr/lpp/java/current
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 All notable changes to the Zowe Installer will be documented in this file.
 <!--Add the PR or issue number to the entry if available.-->
 
+## `2.18.4`
+- Bugfix: `detect_java_home` and `detect_node_home` routines used incorrect shell command [#____](https://github.com/zowe/zowe-install-packaging/pull/____)
+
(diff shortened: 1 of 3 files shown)


MarkAckert (approved):

MarkAckert:
Confirmed the build failure is unrelated to these changes. Merging.

PR #4573, opened 2025-11-11
Change default authentication provider to SAF

arxioly:
- [X] Necessary documentation (if appropriate) have been added / updated

#### PR type
- [X] Other... Please describe: Change of defaults.

- Changes default authentication provider in `components.gateway.apiml.security.auth.provider` to saf instead of zosmf. Following related to zosmf provider properies removed from yaml:
```
components.gateway.apiml.security.auth.zosmf.serviceId: ibmzosmf
components.gateway.apiml.security.auth.zosmf.jwtAutoconfiguration: jwt
```

- [ ] Yes
- [X] No

- [X] Yes
- [ ] No

For new users only, no migration required.

Doc issue/PR number:  https://github.com/zowe/docs-site/pull/4780


balhar-jakub (approved):

balhar-jakub:
@Martin-Zeithaml @MarkAckert Any concerns for this change?

example-zowe.yaml
```diff
@@ -487,10 +487,7 @@ components:
     apiml:
       security:
         auth:
-          provider: zosmf
-          zosmf:
-            jwtAutoconfiguration: jwt
-            serviceId: ibmzosmf
+          provider: saf
```
Martin-Zeithaml, line 493: We can add the same note in comment, which was added to doc-site about PassTicket.

files/defaults.yaml
```diff
@@ -166,10 +166,7 @@ components:
     apiml:
       security:
         auth:
-          provider: zosmf
-          zosmf:
-            jwtAutoconfiguration: jwt
-            serviceId: ibmzosmf
+          provider: saf
```
achmelo: this would affect also existing installations, is it correct?

MarkAckert: Yes, IMO this should be moved to example-zowe.yaml for v3, defaults for v4.

MarkAckert: I should clarify: it _might_ impact existing installations. User `zowe.yaml` files take precedence and by default explicitly set the auth provider to zosmf before, so those should remain intact. However, for changes like this which can modify runtime behavior (given users may not have Passtickets setup for z/OSMF), my preference is to keep defaults clean to avoid surprises.

arxioly: Ok, so remove from defaults, but keep it in example-zowe.yaml, is that correct? Also does anyone aware about the impact of removing this part? Becasue I see, that "test PSWI" is faling, because of differences between zowe.yaml and ZWECONF.xml, but I'm don't know if this is becasue of defaults or example-zowe.yaml.
```
       zosmf:
            jwtAutoconfiguration: jwt
            serviceId: ibmzosmf
```

MarkAckert: I can help out with the PSWI part. Change in `example-zowe.yaml` are the ones that get flagged by PSWI automation.

MarkAckert (approved):

MarkAckert:
Build failure is expected and unrelated to these changes. This can be merged in.

arxioly:
> Build failure is expected and unrelated to these changes. This can be merged in.

Can somebody merge it then? I don't have permissions for that.

PR #4581, opened 2025-11-13
cleanup small bugs in validate port bind and `generateInstanceEnvFrom…

MarkAckert:
Fixes some small bugs in `validate port bind`, and a bigger bug in `generateInstanceEnv...` which could be triggered by a zowe.yaml configured with `zowe.setup.certificate` set to `keyring`.`defaults.yaml` will populate fields with `null` when PKCS12 is missing from `zowe.setup.certificate`, causing the error.

Instead of writing out envs like below, this fix will exclude them from the final output. There is one place I found in `migrate-for-kubernetes` which checks if the env exists, and writing out the null value would lead to incorrect behavior. 

`ZWE_zowe_setup_certificate_pkcs12_directory=null`

To be included in 3.4.0.


1000TurquoisePogs (approved):
config and validate changes look good to me.

bin/commands/validate/port/bind/index.ts
```diff
@@ -19,7 +19,26 @@ const COMMAND_NAME = 'zwe-validate-port-available';
...
+        return 1;
+      }
+    }
+  } else {
+    checkedComponents = enabledComponents;
+  }
+
   let hasErrors = false;
   const ZOWE_CONFIG = config.getZoweConfig();
   let bindUtilPath = ZOWE_CONFIG.zowe.runtimeDirectory;
```
Martin-Zeithaml: Fails if `zowe.runtimeDirectory` not defined.
```
TypeError: cannot read property 'endsWith' of undefined
    at execute (/zowe/4581/bin/commands/validate/port/bind/index:43)
    at <anonymous> (/zowe/4581/bin/commands/validate/port/bind/cli.js:14)
```
But even if defined, we are not sure, `zowe.runtimeDirectory` contains the binary `bind-test`. It could reference (by mistake) Zowe V2. This should be `runtimeDirectory` of `zwe` command.
```javascript
const bindUtilPath = std.getenv('ZWE_zowe_runtimeDirectory') + '/bin/utils/bind-test';
```

bin/commands/validate/port/bind/index.ts
```diff
@@ -19,7 +19,26 @@ const COMMAND_NAME = 'zwe-validate-port-available';
...
   common.requireZoweYaml();
-  let enabledComponents = componentName ? [componentName] : component.getEnabledComponents();
+
+  const enabledComponents = component.getEnabledComponents();
+  let checkedComponents;
+  if (componentName && componentName.trim().length > 0) {
+    if (enabledComponents.includes(componentName)) {
+      checkedComponents = [componentName];
+    } else {
+      const errMsg = `Component '${componentName}' is not enabled. Skipping port validation.`;
```
Martin-Zeithaml: If you made a typo in component name, this will return better explanation:
```javascript
            let errMessageReason = 'enabled';
            if (ZOWE_CONFIG.components[componentName] == undefined) {
                errMessageReason = 'defined';
            }
            const errMsg = `Component '${componentName}' is not ${errMessageReason}. Skipping port validation.`;
```
`ZOWE_CONFIG` must be before this.

Martin-Zeithaml (approved):
It is returning a number, because it is expected to use it during prepare phase. If used in command line, it is always returning 0, but it is acceptable.

Re-run my test, the result was ok, different warning message for disabled or invalid component and no strange characters in the output.

```
[1/4][expected rc=  0, result rc=  0  ]: /zowe/4581/bin/zwe validate port bind -c 'PARMLIB(ZOWE.PR#4581.YAML(TEST1))' -vv -o test
[   ][Description                     ]: Component disabled

[2/4][expected rc=  0, result rc=  0  ]: /zowe/4581/bin/zwe validate port bind -c 'PARMLIB(ZOWE.PR#4581.YAML(TEST2))' -vv -o test
[   ][Description                     ]: Component enabled

[3/4][expected rc=  0, result rc=  0  ]: /zowe/4581/bin/zwe validate port bind -c 'PARMLIB(ZOWE.PR#4581.YAML(TEST1))' -vv -o invalid-component
[   ][Description                     ]: Component invalid

[4/4][expected rc=  0, result rc=  0  ]: /zowe/4581/bin/zwe validate port bind -c /zowe/instances/3.3.0/zowe.yaml -vv
[   ][Description                     ]: Real instance
```

PR #4582, opened 2025-11-14
Refactor zwe components install for dry run

1000TurquoisePogs:
This PR splits `zwe components install` into parts which each have dry run options.

There will be a few parts:
1) zwe components install extract (or symlink)
2) zwe components install component-script (run install script of component if exists) (rename of zwe components install process-hook)
3) zwe components install register appfw zss pcbit (check for PC bit of zss plugins)
4) zwe components install register zis loadlib (handle registration of zis plugin)
5) zwe components install register zis parmlib (handle parmlib update for zis plugin)

This is backwards compatible, existing commands still do the same tasks.
But, there are now more fine-grained commands which those older commands run.

Each operation has a new dry-run mode, and each operation has more checks and more/better messages about progress or warnings.

You can test this by running the commands in the order listed, or just `zwe components install`, with or without dry-run.

With dry-run, you should not see any changes except that temporary files may be created, but they will not impact zowe. temporary files are either to help do later dry-run tasks, or serve as example output that a user can continue with manually.

I have split the code such that the majority of the logic for "register" commands continues to live within bin/libs/component.ts, so that it could be re-used for other purposes.

TODO: message review - many error messages do not have IDs assigned yet.


bin/libs/configmgr.ts
```diff
@@ -54,6 +54,21 @@ const ZOWE_SCHEMA_SET=`${ZOWE_SCHEMA}:${COMMON_SCHEMA}`;
 let ZOWE_CONFIG;
 let HA_CONFIGS = {};
 
+export function getFirstConfigFile() {
+  let configPath = std.getenv('ZWE_PRIVATE_CONFIG_ORIG');
+  if (!configPath) {
+    std.getenv('ZWE_PRIVATE_CONFIG');
```
Gautham-coder: The value returned by std.getenv() should be assigned to configPath?

bin/commands/components/enable/index.ts
```diff
@@ -23,9 +23,41 @@ export function execute(componentId: string, haInstance?: string) {
...
+      common.printMessage(`Global property is currently ${zoweConfig.components[componentId].enabled}`);
+    } else {
+      common.printMessage(`Global property is currently undefined`);
+    }
+    common.printMessage(`Setting property ${componentConfigPath}.enabled to true`);
+  }
+
+  const firstConfigFile = configmgr.getFirstConfigFile();
+  common.printMessage(`Setting ${componentConfigPath}.enabled: true`);
+  common.printMessage(`Updating ${firstConfigFile}`);
```
Gautham-coder: The Setting and Updating messages can be inside the condition where the update actually happens and maybe we can  add message in the else condition for indicating its dry run.

struga0258 (approved):

PR #4583, opened 2025-11-17
Updates zosmf check with attls and bypass options

1000TurquoisePogs:
This PR adds preflight check options similar to what's seen in the new bind test (PR 4447) but also uses zowe yaml config to check if the zosmf check should or shouldnt expect ATTLS+gateway settings.


bin/libs/zosmf.ts
```diff
@@ -13,7 +13,7 @@ import * as std from 'cm_std';
 import * as common from './common';
 import * as shell from './shell';
 
-export function validateZosmfHostAndPort(zosmfHost: string, zosmfPort: number): boolean {
+export function validateZosmfHostAndPort(zosmfHost: string, zosmfPort: number, scheme:string='https', jobname?: string, warnOnly?: boolean): boolean {
```
Martin-Zeithaml: `scheme` not used in all messages:
```javascript
console.log(zosmf.validateZosmfHostAndPort('example.com', 443, 'http'));
```
```
ERROR: Warning: Could not validate if z/OSMF is available on 'https://example.com:443/zosmf/info'. No response code from z/OSMF server.
```

bin/libs/zosmf.ts
```diff
@@ -38,5 +45,5 @@ export function validateZosmfHostAndPort(zosmfHost: string, zosmfPort: number):
   if (zosmfCheckPassed) {
     common.printMessage(`Successfully checked z/OSMF is available on 'https://${zosmfHost}:${zosmfPort}/zosmf/info'`)
   }
-  return zosmfCheckPassed;
+  return zosmfCheckPassed || warnOnly;
```
Martin-Zeithaml, line 48: Returns `undefined` for `console.log(zosmf.validateZosmfHostAndPort('example.com', 443, 'http'));`

```
ERROR: Warning: Could not validate if z/OSMF is available on 'https://example.com:443/zosmf/info'. No response code from z/OSMF server.
undefined
```

MarkAckert: Additionally, `|| warnOnly` is at odds with `internal start prepare` calling the function with `exit==true`, so the current behavior is it warns for `exit` and exits for `warn`. Either the caller or implementation need an update to fix that.

bin/libs/zosmf.ts
```diff
@@ -24,19 +24,26 @@ export function validateZosmfHostAndPort(zosmfHost: string, zosmfPort: number):
   }
   let zosmfCheckPassed=true;
 
-
-  const execReturn = shell.execOutSync(`${std.getenv('ZWE_zowe_runtimeDirectory')}/bin/utils/curl`, `https://${zosmfHost}:${zosmfPort}/zosmf/info`, `-k`, `-H`, `X-CSRF-ZOSMF-HEADER: true`, `-w`, `%{http_code}`, `-s`, `-o`, `/dev/null`);
+  let backupJobname = std.getenv('_BPX_JOBNAME');
+  std.setenv('_BPX_JOBNAME', jobname);
+  const execReturn = shell.execOutSync(`${std.getenv('ZWE_zowe_runtimeDirectory')}/bin/utils/curl`, `${scheme}://${zosmfHost}:${zosmfPort}/zosmf/info`, `-k`, `-H`, `X-CSRF-ZOSMF-HEADER: true`, `-w`, `%{http_code}`, `-s`, `-o`, `/dev/null`);
```
Martin-Zeithaml, line 29: I got stuck here with my test, it happen with `zosmf.validateZosmfHostAndPort('example.com', 4443, 'https')`.
We might consider `--max-time` or similar option as limit. 
It might be better to continue with z/OSMF not validated (and there will be another messages later) rather than to be stuck in the beginning with just couple messages.

MarkAckert: So by default, set the startup check to `warn` ?

MarkAckert:
My review is complete and I _think_ we're good, but I would like a refresher on how `zowe.setup.network.tls.attls` values find their way to the gateway and apiml components, and their order of precedence ( `components.gateway.zowe.setup` >`components.apiml.zowe.setup`  > `zowe.setup` ? ) . 

Other notes: 
- I've changed the default behavior of the zosmf startup check to `warn`.
- The z/osmf startup check will respect protocol from environment variables `ZOSMF_SCHEME` and `ZWE_zOSMF_scheme`
- `zwe support` was updated to invoke the z/OSMF check the same way `internal start prepare` does it.

bin/libs/zosmf.ts
Martin-Zeithaml: This problem is not so serious. I run the `curl` in verbose mode and unfortunately the `example.com` host is resolved into several IP addresses, so it took like 5 minutes to try them all. In case the user make a typo in zOSMF host or port, it will eventually print the correct error message.

bin/commands/internal/start/prepare/index.ts
```diff
@@ -197,9 +197,28 @@ function globalValidate(enabledComponents:string[]): void {
...
-  if (zosmfHost && zosmfPort) {
+  let zosmfCheckAction = ZOWE_CONFIG.zowe.launchScript?.startupChecks?.zosmf || ZOWE_CONFIG.zowe.launchScript?.startupChecks?.default || 'warn';
+  if (zosmfHost && zosmfPort && (zosmfCheckAction != 'disabled')) {
     if (enabledComponents.includes('discovery') || enabledComponents.includes('apiml')) {
-      let zosmfOk = zosmf.validateZosmfHostAndPort(zosmfHost, zosmfPort);
+      const gatewayJobname = (ZOWE_CONFIG.zowe.job?.prefix || 'ZWE1') + 'AG';
+      let checkScheme = 'https';
+      const apimlTlsPolicy = ZOWE_CONFIG.components.gateway?.zowe?.network?.server?.tls?.attls ?? 
+                              ZOWE_CONFIG.components.apiml?.zowe?.network?.server?.tls?.attls ?? 
+                              ZOWE_CONFIG.zowe.network?.server?.tls?.attls ?? false;
```
pablocarle: I think z/OSMF scheme validation should rely on the `client` setting, rather than the server one.
At least from Gateway/ZAAS/API ML (modulith) point of view, we use the client setting to decide whether to use `http` scheme (if client, i.e. outboud rule is set)

bin/commands/support/index.ts
```diff
@@ -94,7 +94,27 @@ export function execute(): void {
...
+            checkScheme = 'http';
+        }
+    } else if (ZOWE_CONFIG.components?.gateway?.enabled === true && gatewayTlsPolicy === true) {
+        checkScheme = 'http';
+    }
+    // envs should overrule attls settings
+    const finalScheme =  std.getenv('ZOSMF_SCHEME') || std.getenv('ZWE_zOSMF_scheme') || checkScheme;
+    const zosmfCheckAction = ZOWE_CONFIG.zowe.launchScript?.startupChecks?.zosmf || ZOWE_CONFIG.zowe.launchScript?.startupChecks?.default || 'warn';
+    //this checks with default tls and jobname settings, since this command is unlikely to be run under the STC account where ATTLS might be used
+    environment["zosmf_check"] = `'${checkScheme}://${zosmfHost}:${zosmfPort}/zosmf/info' => ${zosmf.validateZosmfHostAndPort(zosmfHost, zosmfPort,  finalScheme, gatewayJobname, (zosmfCheckAction == 'warn'))}`;
```
pablocarle, line 111: does this run with the `_BPX_JOBNAME` set to act as ZAAS or Gateway? If not, how does it manage to use the defined outbound rule for the connection to z/OSMF?

Martin-Zeithaml: It is using gateway job name (L97).

bin/commands/internal/start/prepare/index.ts
Martin-Zeithaml: Is [this](https://github.com/zowe/zowe-install-packaging/blob/93c94a8c83a100607d3c7e32f1d248d26877d19b/bin/libs/component.ts#L500) better solution?

bin/commands/support/index.ts
pablocarle: multi-service deployment is still supported, if multi service is enabled it should use ZAAS' jobname

bin/commands/internal/start/prepare/index.ts
pablocarle: I think that's the one we are currently using to determine the scheme to use with z/OSMF,. yes

MarkAckert (approved):

PR #4586, opened 2025-11-18
Sort ZWE_DISCOVERY_SERVICES_LIST to prefer local discovery url

### LINKED ISSUE #4568: Discovery service URLs are not in correct order
**Describe the bug**
Config manager creates an environment variable ZWE_DISCOVERY_SERVICES_LIST that contains list of URLs for all Discovery servies in HA setup. This list is always in the same order, which makes every Zowe component to register with the same Discovery service. 

**To Reproduce**
1. Go to .env folder in the zowe directory

**Expected behavior**
Each Zowe instance will obtain ordered list of Discovery service URLs with it's own as the first one.

**Screenshots**

**Additional context**

1000TurquoisePogs:
ZWE_DISCOVERY_SERVICES_LIST env var is given to each server as a way to tell them which places they can go to do eureka registration. It's expected there's only 1, except when using HA, in which case this is a comma separated list.

Currently, this list is in the same order for every HA instance.

That's not good, because it means every HA instance tries to register to the same destination instead of spreading out the load, but it also leads to higher occurrence of issues overall.

So, this PR changes the list so that the local discovery server, if it exists, is going to be first in the list for each HA instance.

Fixes https://github.com/zowe/zowe-install-packaging/issues/4568

How to test: Start up HA instances and check the value of the ZWE_DISCOVERY_SERVICES_LIST environment variable. You can expect it to be in the same order prior, but with this PR the local URL will be first in the list for each HA instance.


bin/libs/configmgr.ts
```diff
@@ -153,19 +153,43 @@ function getDiscoveryServiceUrlHa(config) {
...
+  const definedHaInstanceNames = haInstanceKeys.map(name=>stringlib.sanitizeAlphanum(name.toLowerCase()));
+  //TODO check if this exists
+  const currentHaInstanceName = std.getenv('ZWE_CLI_PARAMETER_HA_INSTANCE');
+  let currentInstance;
+  for (let i = 0; i < haInstanceKeys.length; i++) {
+    let mappedName = stringlib.sanitizeAlphanum(haInstanceKeys[i].toLowerCase());
+    if (mappedName == currentHaInstanceName) {
+      currentInstance = config.haInstances[haInstanceKeys[i]];
+    }
+  }
```
Martin-Zeithaml, line 172:
```yaml
haInstances:
  lpar1:
    hostname: lpr1.example.com
    sysname: lpr1
  elpardva:
    hostname: lpr2.example.com
    sysname: lpr2
  LPAR3:
    hostname: lpr3.example.com
    sysname: lpr3
  LPAR-4:
    hostname: lpr4.example.com
    sysname: lpr4
  lpar_4:
    hostname: lpr5.example.com
    sysname: lpr5
  lpar#4:
    hostname: lpr6.example.com
    sysname: lpr6
```

`/S ZWESLSTC,HAINST=ELPARDVA`

```
ZWE_DISCOVERY_SERVICES_LIST=
https://lpr2.example.com:7553/eureka/,
https://lpr1.example.com:7553/eureka/,
https://lpr3.example.com:7553/eureka/,
https://lpr4.example.com:7553/eureka/,
https://lpr5.example.com:7553/eureka/,
https://lpr6.example.com:7553/eureka/
```

1) 2nd LPAR started, 2nd hostname on the first place, which is OK (newlines added for readability)
2) IMHO `lpr5` and `lpr6` should not be there, `LPAR-4`, `lpar_4` and `lpar#4` sanitized to `lpar_4`, but the first HA look up match is for `LPAR-4`. So `lpar_4` and `lpar#4` are "unreachable/dead" instances.

bin/libs/configmgr.ts
```diff
@@ -153,19 +153,43 @@ function getDiscoveryServiceUrlHa(config) {
...
+  
+
+  const currentPort = currentInstance?.components?.discovery?.port ? currentInstance.components.discovery.port : config.components.discovery?.port;
```
achmelo, line 175: Discovery port is now possible to configure also under "apiml" component for modulith setup. Does it make sense to include it here? 
property name: `components.apiml.internal.discovery.port`

Issue #4569, opened 2025-11-05
Remove duplicate zOSMF registration file


achmelo:
**Describe the bug**
Each instance of Zowe creates new yaml file for zOSMF registration. These are not used and may lead to a confusion.

**To Reproduce**
1. Go to api-defs folder in workspace
2. See yaml definitions for each insatnace 

**Expected behavior**
Only instances that are used are located in the api-defs folder

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

MarkAckert:
Is this addressed by #4526 ? (Testable in 3.4.0)

PR #4587, opened 2025-11-18
Add 'once' static definition manifest option to ensure only 1 static def per HA instance

1000TurquoisePogs:
This PR tweaks static registration by adding a 'once' parameter in the manifest which tells zwe to keep only 1 static registration file per workspace.

This could be useful for situations where having multiple (when using HA) would result in misleading logs or erroneous behavior

Related to https://github.com/zowe/zowe-install-packaging/issues/4569 but requires discovery's registration file to use this new feature to have the intended effect.

This PR depends upon https://github.com/zowe/api-layer/pull/4403
It uses the ability added here.

You can test this by checking how many copies of zosmf reg file exists within the workspace as compared to zss. You really only want at most 1 zosmf file, but you want as many zss files as there are zss components in the HA environment.

Example of how this should look:

```
~/zowe/inst/once/workspace/api-mediation/api-defs>
total 48
-rwxrwx---   1 me    group     1434 Nov 26 09:43 apiml.zosmf_static_definition_yaml_template.sysa.yml
-rwxrwx---   1 me    group     1077 Nov 26 09:43 zss.apiml_static_reg_yaml_template.sysa.yml
-rwxrwx---   1 me    group     1077 Nov 26 09:42 zss.apiml_static_reg_yaml_template.sysb.yml
```

the name of the template shouldnt matter. whether sysa or sysb 'wins', the contents are going to be the same. that's important of the 'once' feature, it only makes sense if the contents are going to be useful with only 1 copy of static reg file.


achmelo (approved):

1000TurquoisePogs:
The PR https://github.com/zowe/api-layer/pull/4403 was merged and therefore this must also be merged. It is breaking the build of other PRs until then.
The test failure that is holding back cleanly merging this is unrelated to this PR and so I need to force merge this one.

Issue #4594, opened 2025-11-21
Extra files in `files`


Martin-Zeithaml:
`curl-8.13.0.20250516_092121.zos.pax.Z` and `zowe-utility-tools-2.0.3.zip` should be removed when creating the build.
```
cd /zowe/3.4.0-RC1/files
ls
SZWEEXEC                               SZWESAMP                               defaults.yaml                          workflows
SZWELOAD                               curl-8.13.0.20250516_092121.zos.pax.Z  sca                                    zowe-utility-tools-2.0.3.zip
```

PR #4595, opened 2025-11-21
restore njq, remove curl pax and zowe-utility-tools zip from `files/`…

MarkAckert:
address #4594 .

Restores `njq` since it's still used in `migrate for kubernetes` and container scripts, even though those functions aren't actively supported.

To be included in 3.4.0


MarkAckert:
Remote test failures are due to a tricky bug somewhere in the framework, unrelated to this change.

Martin-Zeithaml (approved):

PR #4601, opened 2025-11-25
Simplify example-zowe.yaml based on system symbols and internal references to avoid duplication of data

### LINKED ISSUE #4600: Tidy up the example-zowe.yaml based on feedback from customer support cases
*Customer content*

**Is your feature request related to a problem? Please describe.**
To help customers more easily zowe.yaml introduce &SYSNAME as a system symbol for the IP address and host name where appropriate.
Also make the gateway port be referenced for the external port (to avoid them getting out of sync)
Make the default for the caching service false to help with the initial Zowe setup where it's not being configured for HA.

Joe-Winchester:

#### PR type
- [ ] Bugfix 
- [X] Feature 

#### Relevant issues
Fixes #4600 

- [ ] Yes
- [X] No

- [ ] Yes
- [X] No

### DIFF
--- a/example-zowe.yaml
+++ b/example-zowe.yaml
@@ -275,9 +275,9 @@ zowe:
   # runtime z/OS job name
   job:
     # Zowe JES job name
-    name: ZWE1SV
+    name: "${{ std.getenv('ZWE_components_gateway_port') }}SV"
     # Prefix of component address space
-    prefix: ZWE1
+    prefix: ZWE3
 
   # >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
   # This is an ID you use to separate multiple Zowe installs when determining
@@ -296,14 +296,14 @@ zowe:
   # address bar.
   externalDomains:
     # this should be the domain name to access Zowe APIML Gateway
-    - example.com
+    - "${{ zos.resolveSymbol('&SYSNAME') }}.example.com"
 
   # >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
   # This is the port you use to access Zowe Gateway from your web browser.
   #
   # In many use cases, this should be same as `components.gateway.port`. But in
   # some use cases, like containerization, this port could be different.
-  externalPort: 7554
+  externalPort:  "${{ std.getenv('ZWE_components_gateway_port') }}"
 
   # >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
   #
@@ -318,13 +318,14 @@ zowe:
     server:
       tls:
         attls: false
-        # TLS settings only apply when attls=false
-        # Else you must use AT-TLS configuration for TLS customization.
+        # If you have changed 'attls: true' above, the TLS and SSL levels are defined in the AT-TLS policy, 
+        # see https://docs.zowe.org/stable/user-guide/configuring-at-tls-for-zowe-server/
+        # Only include the two lines below that control the min and max TLS if 'attls: false' above
         minTls: "TLSv1.2"
         maxTls: "TLSv1.3"
     client:
       tls:
-        attls: false
+        attls: "${{ std.getenv('ZWE_network_server_tls_attls')}}"
 
   # >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
   # You can un-comment and define any extra environment variables as key/value
@@ -454,7 +455,7 @@ node:
 #-------------------------------------------------------------------------------
 zOSMF:
   # Host name of your z/OSMF instance
-  host: dvipa.example.com
+  host:  "${{ zos.resolveSymbol('&SYSNAME') }}.example.com"
   port: 443
   applId: IZUDFLT
 
@@ -526,7 +527,7 @@ components:
 
   # >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
   caching-service:
-    enabled: true
+    enabled: false
     port: 7555
     debug: false
 


example-zowe.yaml
```diff
@@ -275,9 +275,9 @@ zowe:
   # runtime z/OS job name
   job:
     # Zowe JES job name
-    name: ZWE1SV
+    name: "${{ std.getenv('ZWE_components_gateway_port') }}SV"
```
Andrew-J-Metzger, line 278: @Joe-Winchester I think this references the wrong value, e.g., "7554SV".

Let's use the zowe job prefix:
```yaml
name: "${{ std.getenv('ZWE_zowe_job_prefix') }}SV"
```

MarkAckert:
In general we've kept templated variables and template logic in the `defaults.yaml` file so the `example-zowe.yaml` can remain as straightforward and readable as possible. While I don't have any issues with templates like `getenv(&SYSNAME)`, it could add some confusion. I don't think there's a hard rule on keeping templates out of `example-zowe.yaml`, but with that in mind, should some of these changes be moved to the `defaults.yaml` file, with the paired values in `example-zowe.yaml` commented out? Thinking `zowe.job.name`, `externalPort`, `client.tls.attls` fit in this category.

For the values which users must supply and there's no advantage to defaults (`externalHost`, `zosmf.host`), I'm a little mixed. Out of the box the values as-written are invalid and have to be replaced anyway, but they point the user towards &SYSNAME. Maybe we can call out the SYSNAME support in line-comments and setup doc?

Last item, configmgr supports symbol resolution via `zos.resolveSymbol()` and imo we should prefer that to `getenv` for clarity reasons...e.g. `${{ zos.resolveSymbol('&SYSNAME')`.

Interested in others' opinions cc @1000TurquoisePogs @Martin-Zeithaml

Joe-Winchester:
I thought I'd tested these changes, but just re-started the STC I was testing with and are getting some errors on the network port and ttls values, that seem to enforce integers and bools respectively, so they don't like the yaml lookups.

I'm thinking of maybe setting the attls client to false by default, but having the external port match the gateway port seems like a good idea so I'll add to the PR to relax the integer enforcement.  

I hadn't thought about the `defaults.yaml`,  that's a good idea I'll noodle on.

```
    type 'object' not permitted at /zowe/externalPort; expecting type 'integer'
    Validity Exceptions(s) with object at /zowe/network
      Validity Exceptions(s) with object at /zowe/network/client
        Schema at '/zowe/network/client/tls' invalid
          not anyOf schemas at '/zowe/network/client/tls' are valid
            Validity Exceptions(s) with object at /zowe/network/client/tls
              unequal constant value at /zowe/network/client/tls/attls
            Validity Exceptions(s) with object at /zowe/network/client/tls
              Schema at '/zowe/network/client/tls/attls' invalid
                not anyOf schemas at '/zowe/network/client/tls/attls' are valid
                  no matching enum value at /zowe/network/client/tls/attls; expecting one of values '[true, false]' of type 'string'
                  unequal constant value at /zowe/network/client/tls/attls

```

Martin-Zeithaml:
## Bulletproof definition
Any embedded value must be written in bulletproof form, like what if nothing is defined, is returned value valid etc.
This is wrong, beacuse
* (typo) `ZWE_network_server_tls_attls` should be `ZWE_zowe_network_server_tls_attls`
* What will be set, if `zowe.network.server.tls.attls` is not defined at all? It will return nothing, which is not valid `boolean` value.
```
attls: "${{ std.getenv('ZWE_network_server_tls_attls')}}"
```
## Readability
Let's consider the previous example should be like this:
```
attls: "${{ zowe.network?.server?.tls?.attls ? false : true }}"
```
Is this easy to read for every customer?

## Accessibility
Is every `zwe` command `configmgr` based? This requires small research, there were some PRs recently to remove "old" code, but I am not sure about this.

1000TurquoisePogs:
This is incompatible with containers and thus some consideration for containers is needed, such as to take an old example and shove it into `files/examples/containers/zowe.yaml`

example-zowe.yaml
```diff
@@ -318,13 +318,14 @@ zowe:
...
       tls:
         attls: false
-        # TLS settings only apply when attls=false
-        # Else you must use AT-TLS configuration for TLS customization.
+        # If you have changed 'attls: true' above, the TLS and SSL levels are defined in the AT-TLS policy, 
+        # see https://docs.zowe.org/stable/user-guide/configuring-at-tls-for-zowe-server/
+        # Only include the two lines below that control the min and max TLS if 'attls: false' above
         minTls: "TLSv1.2"
         maxTls: "TLSv1.3"
     client:
```
1000TurquoisePogs, line 326: the easiest solution for 'client' is just to get rid of it from the example.
it's an advanced option, and, allowing attls to be toggled on just client was a mistake so i think our code doesnt even respect it now.

example-zowe.yaml
```diff
@@ -526,7 +527,7 @@ components:
 
   # >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
   caching-service:
-    enabled: true
+    enabled: false
```
1000TurquoisePogs, line 530: this has little effect in staging since it's forced on within apiml modulith.
a better simplification would be to remove attributes of apiml components which arent used in modulith, such as port. i would leave that to APIML devs to determine how best to refine.

1000TurquoisePogs:
use of templates, and certain templates, are not backward compatible *for certain zwe phases*. That is, you can't use this file on an older version of zowe when doing an install, but you probably could at runtime.

some people may have v2 and v3, or various v3 versions, or want to do a downgrade, and encounter errors.

i would like to improve things, but we must be careful of what should wait for v4 and what can be done now. we need to know what impact changes will have.

1000TurquoisePogs:
Also see https://github.com/zowe/zowe-install-packaging/pull/4605

Joe-Winchester:
Lots of great feedback - thank you !
I'm going to convert this to draft,  as I think it needs more thought especially if this will break containers and also the zwe commands that don't respect symbols.
I think I'll work instead of the troubleshooting section to add some more docs.
The only thing I was thinking about was whether the 
```
externalPort: "${{ std.getenv('ZWE_components_gateway_port') }}"
```
is worth keeping, as I don't see why anyone would not want the two values to be in sync ?

PR #4602, opened 2025-11-25
3.4.0-RC2 component install was not updating zowe.yaml

MarkAckert:
This fixes an issue found in RC2 where component installations were not updating the zowe.yaml file properly.

Before the Node.JS removal PR, all updates through `json.ts#updateZoweYaml` would write to both the supplied file parameter **and** an originating `zowe.yaml` file. This `zowe.yaml` file is uncovered through `ZOWE_CONFIG_PATH` in `configmgr.ts`, and _not_ via the env `ZWE_CLI_PARAMETER_CONFIG`, as this value is sometimes modified and updated to a temporary merged yaml file. If the file parameter and the original `zowe.yaml` are one and the same, then it's written twice.

After the Node.JS removal PR, updates were only written to the supplied file parameter. This behavior has been separated into a new function and is called by `ModifyZoweYaml.ts`, while all other calls will continue with the older behavior.

### DIFF
--- a/bin/libs/json.ts
+++ b/bin/libs/json.ts
@@ -108,16 +108,17 @@ function buildUpdateObjWithArrays(zoweConfig: any, key: string): any {
 }
 
 /**
- * Updates the YAML key in the zowe.yaml passed by the caller with val. Always overwrites on-disk.
+ * Updates the provided zowe.yaml file ONLY with the YAML key and value passed by the caller. Always overwrites on-disk. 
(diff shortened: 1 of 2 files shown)


Martin-Zeithaml (approved):

1000TurquoisePogs (approved):
This looks good to me with comments on why these 2 functions exist

struga0258 (approved):

PR #4609, opened 2025-11-27
HA instance lookup in `zwe` commands

### LINKED ISSUE #4604: HA instances lookup in `zwe internal config set/get`
#4370 allows to use mixed case for HA instance name.

`zwe internal config set/get` is internally using [sanitizeHaInstanceId](https://github.com/zowe/zowe-install-packaging/blob/d63ca00bd831b403f5c4e8b1dca32d97111ce719/bin/libs/config.ts#L252) to transform the name.

Then for the following example

```yaml
zowe:
  useConfigmgr: true
haInstances:
  LPAR1:
    hostname: lpar1.example.com
    sysname: LPR1
```

It is impossible to get `LPAR1` values:
```
zwe config get --config ./zowe.yaml --ha-instance LPAR1 --path hostname
```

The `LPAR1` value is sanitized to `lpar1` and such object due to the case sensitivity is not found.

Martin-Zeithaml:
Fixes #4604 
## Changes
* HA lookup same as done in #4370 
* `internal config set` did not check the value properly
* `internal config set` ends with return code (it was always 0 before)

### Tests

```yaml
# ./yaml/zowe.1ha.yaml
zowe:
  useConfigmgr: true
haInstances:
  LPAR1:
    hostname: LPAR1.example.com
    sysname: LPAR1
  lpar1:
    hostname: lpar1.example.com
    sysname: lpar1
  LPAR-2:
    hostname: LPAR-2.example.com
    sysname: LPAR-2
  LPAR_2:
    hostname: LPAR_2.example.com
    sysname: LPAR_2
```

#### First lowercased match

Returns the same value, because the lookup stops at the first lowercased sanitized match.

```
/zowe/4609/bin/zwe config get --config ./yaml/zowe.1ha.yaml --ha-instance lpar1 --path hostname
LPAR1.example.com

/zowe/4609/bin/zwe config get --config ./yaml/zowe.1ha.yaml --ha-instance LPAR1 --path hostname
LPAR1.example.com
```

#### Sanitized match

`lpar-2` and `LPAR#2` are sanitized to `lpar_2`. Because this HA exists, all the following commands return the same value.

```
/zowe/4609/bin/zwe config get --config ./yaml/zowe.1ha.yaml --ha-instance lpar-2 --path hostname
LPAR_2.example.com

/zowe/4609/bin/zwe config get --config ./yaml/zowe.1ha.yaml --ha-instance LPAR#2 --path hostname
LPAR_2.example.com

/zowe/4609/bin/zwe config get --config ./yaml/zowe.1ha.yaml --ha-instance lpAr_2 --path hostname
LPAR_2.example.com
```

#### Set
* An update, which ends with error, ends with non-zero return code.
* Update number, but as a string
* Update HA instance, again the uppercased instance has the first match.
```
/zowe/4609/bin/zwe internal config set --config ./yaml/zowe.no_ha.yaml --ha-instance LPAR2 --path hostname --value host.example.com
- update zowe config ./yaml/zowe.no_ha.yaml, key: "haInstances.lpar2.hostname" with value: host.example.com, and validate: true
  * jqset success
Error: Validation of update to zowe-update-yaml found invalid JSON Schema data
  Validity Exceptions(s) with object at /haInstances
    Validity Exceptions(s) with object at /haInstances/lpar2
      missing required property 'sysname' at '/haInstances/lpar2'
# result rc=1


/zowe/4609/bin/zwe internal config set --config ./yaml/zowe.no_ha.yaml --path zowe.setup.jcl.header --value 123456 --string
- update zowe config ./yaml/zowe.no_ha.yaml, key: ".zowe.setup.jcl.header" with value: 123456, and validate: true
  * jqset success
cat ./yaml/zowe.no_ha.yaml
---
zowe:
  useConfigmgr: true
  workspaceDirectory: "/zowe/zwet/config.look.up.4609/ws"
  setup:
    jcl:
      header: "123456"
...

/zowe/4609/bin/zwe internal config set --config ./yaml/zowe.update.yaml --ha-instance lpar1 --path hostname --value new-lpar.example.com
- update zowe config ./yaml/zowe.update.yaml, key: "haInstances.LPAR1.hostname" with value: new-lpar.example.com, and validate: true
  * jqset success
cat ./yaml/zowe.update.yaml
---
zowe:
  workspaceDirectory: "/zowe/zwet/config.look.up.4609/ws"
haInstances:
  LPAR1:
    hostname: "new-lpar.example.com"
    sysname: "LPAR1"
  lpar1:
    hostname: "lpar1.example.com"
    sysname: "lpar1"
...
```


MarkAckert:
Re: `With the current code, we can't create for example new HA instance:`

All the updates written through `json.ts` by default require validation of the updated `zowe.yaml` against the schema, but this can be disabled pretty easily with an extra param to `updateZoweYaml` and passing that to `config.updateZoweConfig`. I think if we want to support this, by default validation should always run, and users can disable it with some parameter like `--skip-schema-validate`.  IIRC the old shell scripts never ran schema validation after config update which is convenient, but I think validation will likely help us more often than not.

Martin-Zeithaml:
> Re: `With the current code, we can't create for example new HA instance:`
> 
> All the updates written through `json.ts` by default require validation of the updated `zowe.yaml` against the schema, but this can be disabled pretty easily with an extra param to `updateZoweYaml` and passing that to `config.updateZoweConfig`. I think if we want to support this, by default validation should always run, and users can disable it with some parameter like `--skip-schema-validate`. IIRC the old shell scripts never ran schema validation after config update which is convenient, but I think validation will likely help us more often than not.

I tried that yesterday and add new parameter `--skip-validation`. It was updating the configuration without validation. But it failed in the next step, when we load the configuration in the command itself.

```
zwe internal config set --c zowe.yaml --path hostname --value host.example.com --ha-instance newlpar --skip-validation
```
Config was updated successfully. But to keep it valid, we have to add required `sysname`:

```
zwe internal config set --c zowe.yaml --path sysname --value NEW --ha-instance newlpar
```
Ends with schema validation error from `internal/config/set/index.ts` where we load the configuration (at the beginning) via `config.getZoweConfig()`.

The `set` command is internal only, there is no "official" `zwe config set`. I would keep it as it is, with this limitation. Internally we have functionality to update the config with object. And if needed, we can extend the functionality to update multiple keys.

And related to old code, yes, it is/was wild. I am for strict validation, each modification of configuration must end validation.

1000TurquoisePogs (approved):
This looks correct to me about needing to do extra work on case sensitivity on get and set operations.

MarkAckert (approved):

MarkAckert:
added test cases from the pr description to the remote integration suite. lgtm, merging in

PR #4621, opened 2025-12-09
No additional items for `java` and `node`

Martin-Zeithaml:
Eliminates errors like this (wrong `node` indentation):

```yaml
zowe:
  launchScript:
    startupChecks:
      default: exit
      default: This does not matter
  externalPort: 7554
  externalPort: This does not matter
java:
  home: /path/to/java/
  node:
  home: /path/to/node/
```
```
zwe config get -c ./zowe.yaml --path .java
{
  "home": "/path/to/node/",
  "node": null
}
```

Note:
* `java.home` is defined twice and the second value is used.
* For other types (`enum` | `number`) it looks like the first value is used and the rest is ignored.

### DIFF
--- a/schemas/zowe-yaml-schema.json
+++ b/schemas/zowe-yaml-schema.json
@@ -786,6 +786,7 @@
     },
     "java": {
       "type": "object",
+      "additionalProperties": false,
       "properties": {
         "home": {
           "$ref": "/schemas/v2/server-common#zoweOptionalPath",
@@ -795,6 +796,7 @@
     },
     "node": {
       "type": "object",
+      "additionalProperties": false,
       "properties": {
         "home": {
           "$ref": "/schemas/v2/server-common#zoweOptionalPath",


1000TurquoisePogs:
This looks good but it prevents new properties from being used on old installs.

Such as, if we later enhance java like:

```
java:
  home: abc
  securityFile: java.security
```

If someone used that in v3.8, they cant go use it back in v3.4 if `"additionalProperties": false`
It would be better if they could, and unknown properties just get ignored.
It would probably be best if we had no `additionalProperties: false` anywhere in our schemas and instead relied upon `requires` instead for this reason.

It's annoying that we can't make simple changes without consequences, but what I'd prefer here is an enhancement to configmgr or its yaml parser which throws an error if it sees 2 of the same properties in an object somehow. That behavior isnt part of json-schema. `uniqueItems` only applies to arrays sadly.

PR #4625, opened 2025-12-10
Set defaults for individual component startup checks to `null`

### LINKED ISSUE #4620: To discuss: startup checks in defaults
>  Individual startup checks will override the default behavior

Let's say, user sets only `zowe.launchScript.startupChecks.default = disabled`:

```yaml
# zowe.main.yaml
zowe:
  launchScript:
    startupChecks:
      default: disabled
```
Merged config contains all checks:
```
zwe config get -c zowe.main.yaml --path zowe.launchScript.startupChecks
{
  "default": "disabled",
  "ports": "exit",
  "certificate": "exit",
  "zosmf": "exit"
}
```

Expectation is that all checks are disabled. But in fact, all individual checks are set to `exit`.

From the user point of view, this might be confusing
* The `defaults.yaml` is "hidden"
* Keys not in main configuration are set

MarkAckert:
Fixes #4620

By setting the individual startup checks to `null` in the `defaults.yaml`, the following behavior is followed in order:

1. By default, individual checks are set to "exit"
2. If `zowe.launchScript.startupChecks.default` is set, use that value instead. 
   * This is currently set in `defaults.yaml`, and will always take effect in practice.
4. If `zowe.launchScript.startupChecks.<component>` is set, use that value instead.

If future startupChecks _should not default to "exit" as defined in 1._, it will require an update to the `getStartupCheckMode` logic.

Tests to be written/updated.


Martin-Zeithaml:
Small unit test for `getStartupCheckMode`:

```javascript
const ZOWE_CONFIG=config.getZoweConfig();
console.log(JSON.stringify(ZOWE_CONFIG.zowe.launchScript.startupChecks, null, 2));

function getStartupCheckMode(property) {
/* Original code ... */
/* Just changed "return" to "console.log" */
  console.log(`${property} => DO(${doCheck}) WARN(${warnOnly})`);`  // Debug update
}
getStartupCheckMode('default');
getStartupCheckMode('zosmf');
getStartupCheckMode('port');
```
Different configs:
## No startupChecks
```
{
  "default": "exit"
}
default => DO(true) WARN(false)
zosmf => DO(true) WARN(false)
port => DO(true) WARN(false)
```
## Only default
```
{
  "default": "disabled"
}
default => DO(false) WARN(false)
zosmf => DO(false) WARN(false)
port => DO(false) WARN(false)
```
## Default & zosmf
```
{
  "default": "warn",
  "zosmf": "exit"
}
default => DO(true) WARN(true)
zosmf => DO(true) WARN(false)
port => DO(true) WARN(true)
```
## Default & zosmf & port
```
{
  "default": "exit",
  "zosmf": "disabled",
  "port": "warn"
}
default => DO(true) WARN(false)
zosmf => DO(false) WARN(false)
port => DO(true) WARN(true)
```

Looks good.

MarkAckert:
@Martin-Zeithaml I added the first unit test in `[c770a84](https://github.com/zowe/zowe-install-packaging/pull/4625/commits/c770a84cc3770d27be119ce47bb6b9c623ea9c1f)`. This is a proof of concept for unit tests we could take forward, and it tracks/runs any `*.cmgr.ts` file under the `__configmgr__` directory as a unit test, and ignores the rest. There's a bunch of ugly configuration and duct tape making it all work, including [this symlink](https://github.com/zowe/zowe-install-packaging/blob/user/markackert/switch-startupCheck-defaults/tests/zwe-remote-integration/src/__tests__/unit/__configmgr__/bin), and [exporting private functions carefully from modules in `./bin`](https://github.com/zowe/zowe-install-packaging/pull/4625/commits/c770a84cc3770d27be119ce47bb6b9c623ea9c1f#diff-9f1ec4b3f092b42fcd644522a3e4560364d2043704ad54979d697809c63f203bR555), but the approach does work. If you think it's better off deferred to another PR, lmk and I can separate it.

tests/zwe-remote-integration/src/__tests__/unit/__configmgr__/tests/launchScript.cmgr.ts
```diff
@@ -0,0 +1,47 @@
...
+ * Copyright Contributors to the Zowe Project.
+ */
+
+import * as config from '../bin/libs/config';
+import { _unit_test } from '../bin/commands/internal/start/prepare/index';
+import { assertEqualsStrict } from './common/assert';
+import * as common from '../bin/libs/common';
+
+const ZOWE_CONFIG = config.getZoweConfig();
+common.printMessage('Starting "getStartupCheckMode" test cases.');
```
Martin-Zeithaml, line 17: Not a problem now, but if we do more unit tests, there might be the functions of the same name or a function (like in this example) not defined in `libs`. We might start identifying like "/bin/command/internal/start/prepare -> getStartupCheckMode". Just for consideration.

struga0258 (approved):

PR #4645, opened 2026-01-05
Recommend split keyring for new configurations

pablocarle:

#### PR type

- [ ] Bugfix

Split keyrings into two:
1. Contains the private key and full chain of the CA used to sign Zowe's certificate.
2. Contains the trusted certificates (Zowe server chain, z/OSMF certificate, other trusted certificates such as southbound services')

The reason behind it lies in AT-TLS configurations. 
Currently, AT-TLS configuration uses keyring to set up handshakes between services; the problem lies in that some connections may be using the `PERSONAL` certificate from the keyring to authenticate without knowing it, since AT-TLS will pick up certificates marked as `DEFAULT`.
The chosen way to deal with this is to split the configuration in AT-TLS to have a keyring that contains the private key and another one that only has trusted certificates acting as a truststore.'

This PR aims at updating the recommendations for new installations, since it's not known at the beginning if the installation will use AT-TLS or not.

Documentation is being updated for the AT-TLS article, targeting the 3.4.0 release.

- [ ] No

- [ ] No

Doc issue/PR number:


files/SZWESAMP/ZWEKRING
```diff
@@ -55,51 +55,55 @@
 //* 4) Update the SET ZOWERING= statement to match the desired
 //*    name of the keyring owned by the &ZOWEUSER user ID.
```
arxioly: Shouldn't be this statement updated as well to specify the difference from ZOWETRST?

files/SZWESAMP/ZWEKRING
```diff
@@ -125,8 +129,10 @@
 //         SET ZOWEUSER=ZWESVUSR     * userid for Zowe started task
 //*                     12345678
 //*
-//*      * Keyring for the Zowe userid
+//*      * Keyring that holds private key for the Zowe userid
 //         SET ZOWERING='ZoweKeyring'
+//*      * Keyring that holds trusted certificates for the Zowe userid
+//.        SET ZOWETRST='ZoweTrstRing'
```
arxioly: Just a suggestion...
```suggestion
//.        SET ZOWETRST='ZoweTrustRing'
```

files/SZWESAMP/ZWEKRING
```diff
@@ -332,7 +354,7 @@ $$
 /* authority (CA) with the keyring ................................. */
    RACDCERT CONNECT(CERTAUTH +
             LABEL('&ROOTZFCA.') +
-            RING(&ZOWERING.) USAGE(CERTAUTH)) +
+            RING(&ZOWETRST.) USAGE(CERTAUTH)) +
             ID(&ZOWEUSER.)
```
arxioly, line 359: Strange that there is nothing about intemediate z/OSMF's certificate at all...

files/SZWESAMP/ZWEKRING
```diff
@@ -354,6 +376,7 @@ $$
 
 /* Define profiles that control certificate access ................. */
    RDEFINE RDATALIB &ZOWEUSER..&ZOWERING..LST UACC(NONE)
+   RDEFINE RDATALIB &ZOWEUSER..&ZOWETRST..LST UACC(NONE)
```
arxioly: Well, if you will define the profile w/o giving the access to it (`UACC(NONE)`), Zowe server user won't have access to it . I would remove this definition for ZOWETRST since this permission is make sence only with regards to access to private key or if someone else would like to have access to the keyring.

files/SZWESAMP/ZWEKRING
```diff
@@ -459,7 +485,7 @@ ACF
 * Connect a Zowe's certificate with the keyring ................... */
    SET PROFILE(USER) DIVISION(CERTDATA)
    CONNECT CERTDATA(&ZOWEUSER..ZOWECERT) -
-   KEYRING(&ZOWEUSER..ZOWERING) USAGE(PERSONAL) DEFAULT
```
arxioly, line 462: So we are applying to variants here - without default certificate and 2 keyrings? Does it make sence than to remove DEFAULT also in RACF commands?

pablocarle: I missed this one, a leftover from when I wanted to ensure defaults were not set. I'll restore it

files/SZWESAMP/ZWEKRING
pablocarle: I wasn't sure about character length but if it's ok I'm fine with it

arxioly: It should be fine, z/OSMF default keyring name is definetley longer.

files/SZWESAMP/ZWEKRING
pablocarle: Should we add something? any suggestion?

arxioly: If you think it make sence, then we could add something similar like for Zowe's:
```
   RACDCERT CONNECT(CERTAUTH +
            LABEL('&ITRMZFCA.') +
            RING(&ZOWETRST.) USAGE(CERTAUTH)) +
            ID(&ZOWEUSER.)
```
and add new parameter `ITRMZFCA` in SET section.

files/SZWESAMP/ZWEKRING
Martin-Zeithaml, line 20: This sample is kept due to the SMP/E installation. It is deprecated, but updates are of course possible.
Those changes must be reflected in split versions of this sample:
* Keyring ACF2
  * ZWEIKRA1
  * ZWEIKRA2
  * ZWEIKRA3
* Keyring RACF
  * ZWEIKRR1
  * ZWEIKRR2
  * ZWEIKRR3
* Keyring Top Secret
  * ZWEIKRT1
  * ZWEIKRT2
  * ZWEIKRT3

pablocarle: Thank you, I'll bring the updates there

pablocarle:
Based on discussion in z/OS squad call, I made changes to files/SZWESAMP/ZWEIKRA1 as a sample, to discuss if these are appropriate changes, once decided I can port to other files
@Martin-Zeithaml @MarkAckert 
cc @1000TurquoisePogs @arxioly

files/SZWESAMP/ZWEIKRA1
```diff
@@ -176,14 +181,70 @@ ACF
 $$
 //IFZWCAED ENDIF
 //*
+
+//* If truststore is selected, create keyring for it
+//* If truststore is selected, connect z/OSMF certificate to it,
+//* otherwise, connect to keyring
+//IFTRST IF (&IFTRSTST EQ 1) THEN
```
Martin-Zeithaml, line 188: I am not sure about this:
```
2 //       SET IFTRSTST=0              
3 //IF1    IF (&IFTRSTST = 1) THEN     
  IEFC653I SUBSTITUTION JCL - (0 = 1)  
4 //IEFBR15  EXEC PGM=IEFBR15          
5 //       ELSE                        
6 //IEFBR14  EXEC PGM=IEFBR14          
7 //       ENDIF                                           
```

This JCL ended with ABEND806-04, which means an attempt to call IEFBR15 (which does not exist).

Based on the doc, only predefined operands could be used (RC, ABEND, ABENDCC, RUN).

arxioly: The problem might be in empty lines w/o `*`.

pablocarle: How can I try out these JCLs?

MarkAckert: Running `zwe init generate` will give you the rendered JCL in your `zowe.setup.dataset.jcllib` dataset.

PR #4652, opened 2026-01-12
Code Review updates for default-to-bring-your-own-keyrings

MarkAckert:
Important changes from base branch:

* Changes certificate setup scenarios so they must be overlayed onto the main `zowe.yaml` file, either manually (copy/append) or with configmgr (`FILE(scenario.yaml):FILE(zowe.yaml)`). This is documented in comments within the scenario YAML files.
* Multiple updates to `ZWEAMLCF` and `ZWECONF` to sync them up with `zowe.yaml`. In the case of `ZWEAMLCF`, existing changes and opinionated config were retained.
  - `zowe.setup.certificate` sections removed from both workflows (they were commented out before). If users want to run `zwe init certificate`, they need to follow the doc. The config workflow won't help customers with `init certificate`, and it didn't before anyway.
  - changed defaults in both workflows from `PKCS12` to keyring + JCERACFKS.
     * removed unused fields related to certificate/pems in the workflows 
     * **TODO**: should we retain PKCS12 as a selectable option in ZWECONF?
* Added PSWI test case for `ZWEAMLCF` so we can track deltas with `example-zowe.yaml`. This was manual / best effort before.


1000TurquoisePogs:
> TODO: should we retain PKCS12 as a selectable option in ZWECONF?

When we talked about this whole keyring-vs-PKCS12 topic with the TSC, the general opinion I got was it's unlikely such a user exists and we can cross that bridge if we get to it.

1000TurquoisePogs (approved):

PR #4654, opened 2026-01-13
PSWI: do not execute 'zwe init generate' for "Create Configuration" step

### LINKED ISSUE #4653: PSWI Create Configuration step is failing
Looks like the Create Configuration step is running `zwe init generate`:
* Zowe 3.3.0+
* This fails due to no MVS datasets are created by Create Configuration step 

Image

There is a workaround to edit the JCL and remove `zwe init generate` statement.

Martin-Zeithaml:
Fixes #4653 

## When

* Zowe 3.3.0+
* PSWI
* JCL feature

## What

Create Configuration steps executes `zwe init generate` before the MVS datasets are created. It fails with 
```
ERROR: ZWEL0143E Cannot find data set member 'prefix.SZWESAMP(ZWEGENER)'. You may need to re-run zwe install.
```
and the entire step fails.

## Fix
The `if` block executing `zwe init generate` was removed.

## To Do ideas
* Copy the JOB card from deployment (should be possible)
* For JCL, add `JOBPARM SYSAFF=` drop down menu
* Review the enforced dataset prefix format


1000TurquoisePogs (approved):
Looks like the prerequisite JCL header parameter already exists in this code, so, yeah this looks like the correct change to do since new installs default to JCL-based install.

PR #4693, opened 2026-02-27
Preflight check - UID 0

Martin-Zeithaml:
Check if `ZWEVSUSR` (or whoever is running) is UID 0.

* new check with YAML option `zowe.launchScript.startupChecks.user`

## General issue
* **Very bad practice**

### Known issue with `ipcs`
* V3: Only if Zowe Desktop is used
* V2: Always
* `cleanup-ipc-mq.sh`: _node.js instance is not fully cleaned up when exits. As time going, the message queue will be full and any node.js command will generate errors; script suggested by IBM_
* `ipcs` list all interprocess elements for `ZWESVUSR` (message queues | shared memory | semaphores)
* `ipcrm` remove "dead" elements

If multiple instances are running under the same user, it will be removing elements for the other users:
* `ZWESVUSR` running Zowe 3.0.0 on `LPR1` - started as the first
* `ZWESVUSR` running Zowe 3.4.1 on `LPR1` - this instance will successfully remove "3.0.0" elements


1000TurquoisePogs:
In my opinion it should default to exiting, which means that yes it would need a YAML property to opt-out. I guess that could require an SMPE HOLD to explain, but for the sake of security it would be a good idea to prevent this.

But just my opinion. Lets see what @MarkAckert  thinks.

MarkAckert:
My first thought was to add this as a pre-flight check with `warn` in v3 and `exit` in v4, but after reflecting, we'll be better served by `exit` in v3 with matching HOLDDATA. Nothing good can happen running Zowe as UID 0, and it's a poor substitute for a correct security configuration. 

I do think users should still be able to opt-out by switching the preflight from `exit` to `warn`, and we should doc it in the HOLDDATA, but we should also recommend strongly against it in the same breath.

Probably worth bringing this up on the TSC call since it may create some support tickets.

bin/commands/internal/start/prepare/index.ts
```diff
@@ -464,6 +464,20 @@ function configureComponents(componentEnvironments?: any, enabledComponents?:str
 
 // Few early steps even before initialization
 
+// Check if user is UID 0, this is not recommended
+let userCheckAction = ZOWE_CONFIG.zowe.launchScript?.startupChecks?.user || ZOWE_CONFIG.zowe.launchScript?.startupChecks?.default || 'exit';
+if (userCheckAction != 'disabled') {
+  const userID = shell.execOutSync('sh', '-c', 'id -u');
+  if (userID.rc == 0 && userID.out) {
+    if (userID.out == "0") {
```
ChongZhou-Broadcom: This line and the previous line can be combined to one:

```ts
if (userID.rc == 0 && userID.out === "0") {
  //...
```

ChongZhou-Broadcom (approved):
I didn't see noticeable problems. Just a minor thing as mentioned below. I am not requesting changes because it is the same logic anyway.

By the way, is it OK to silently ignore this check if the "id -u" returns non-zero?

bin/commands/internal/start/prepare/index.ts
```diff
@@ -464,6 +464,20 @@ function configureComponents(componentEnvironments?: any, enabledComponents?:str
 
 // Few early steps even before initialization
 
+// Check if user is UID 0, this is not recommended
+let userCheckAction = ZOWE_CONFIG.zowe.launchScript?.startupChecks?.user || ZOWE_CONFIG.zowe.launchScript?.startupChecks?.default || 'exit';
+if (userCheckAction != 'disabled') {
+  const userID = shell.execOutSync('sh', '-c', 'id -u');
+  if (userID.rc == 0 && userID.out) {
```
ChongZhou-Broadcom: Should I understand the logic in this way? ---

If `userCheckAction != 'disabled'`, then we can silently pass this check only when `userID.rc == 0 && userID.out !== "0"`. If my understanding is correct, then line 469 to 479 cannot perfectly do the logic.

If the user wants the script to do strict security check, then he can pass the check only if we know he is not Mr. 0. But now the logic is "he can pass the check if we know he is not Mr. 0 or if we don't know who he is."

Martin-Zeithaml: The logic should be:
* If check is disabled, skip it completely
* Otherwise ("warn" or "exit"), do the UID0 check
* If UID = 0, then print error message
* If check is "exit", terminate the processing, otherwise continue

Martin-Zeithaml: > If the user wants the script to do strict security check, then he can pass the check only if we know he is not Mr. 0. But now the logic is "he can pass the check if we know he is not Mr. 0 or if we don't know who he is."

With the check != disabled, you will pass this if you are not UID 0.
If we don't know, who you are, it means, that you were able to start ZWESLSTC without valid ACID. Which means, there is a bigger problem in External Security Manager than in this code.

ChongZhou-Broadcom: OK

bin/commands/internal/start/prepare/index.ts
```diff
@@ -464,6 +464,20 @@ function configureComponents(componentEnvironments?: any, enabledComponents?:str
...
+    if (userID.out == "0") {
+      common.printFormattedError("ZWELS", "zwe-internal-start-prepare", `${user} running as UID 0. Such a setting is strongly discouraged."`);
```
1000TurquoisePogs: i dont expect to see ${user} be anything other than root or bpx* or omvs* so it may be a funny message. alternative could be "Zowe is running as UID 0" instead.

1000TurquoisePogs (approved):
It looks good. I have a suggestion for an alternative message but it is not necessary.

bin/commands/internal/start/prepare/index.ts
Martin-Zeithaml: The customer with UID0 was using ZWESVUSR name for Zowe User. In this case, the message make sense, "ZWESVUSR is UID0".

If you set BPXROOT as Zowe user, then the message "BPXROOT is UID0" is overkill, but the main information for support/debugging is there.

Anyway the user name is part of the message prefix and simplification is always good. I will change the message to "Running as UID0".

`Such a setting is strongly discouraged.` - it was suggested by AI, if you don't like it, let's change it.
_"Strongly discouraged" is a phrase used to express that an action is highly inadvisable, dangerous, or contrary to best practices, though not strictly prohibited._

PR #4700, opened 2026-03-06
zwe validate components

1000TurquoisePogs:
This PR adds the command `zwe validate components` which is just used to ensure that enabled components in the YAML actually exists. This prevents confusion when we have seen where people just guess that something exists, and then gets weird behavior such as verifying a port that isn't used.

This command can be invoked manually to check any component(s) desired, but is invoked automatically at startup just to make sure that every enabled component actually exists. Disabled components are skipped over except when explicitly asked to be checked by the zwe command line.

>>> This _could_ cause minor grief for users who upgraded from v2, where they are asking for non-existent components like `metrics-server`.  So, we default this one to **warn** instead of **error**

#### PR type
- [x] Feature 

- [x] Yes


bin/commands/validate/components/index.ts
```diff
@@ -0,0 +1,105 @@
...
+  This program and the accompanying materials are made available
+  under the terms of the Eclipse Public License v2.0 which
+  accompanies this distribution, and is available at
+  https://www.eclipse.org/legal/epl-v20.html
+
+  SPDX-License-Identifier: EPL-2.0
+
+  Copyright Contributors to the Zowe Project.
+*/
+
```
Martin-Zeithaml, line 11: Missing `zowe.runtimeDirectory` in config, the result is `zss` cannot be started, which is not wrong, but missing runtime should be reported.
```
/zowe/pr/4700/bin: ./zwe validate components -c q.yaml -o zss
2026-03-13 07:21:33 <ZWELS:67438141> MARTIN INFO (zwe-validate-components) Validating manifests for 1 component(s).
ERROR: 2026-03-13 07:21:33 <ZWELS:67438141> MARTIN ERROR (zwe-validate-components) ZWEL0330E: Component 'zss' directory was not found in the runtime or extension directories.
ERROR: 2026-03-13 07:21:33 <ZWELS:67438141> MARTIN ERROR (zwe-validate-components) ZWEL0328E: The following component(s) are missing a valid manifest file and cannot be started: zss
```

bin/commands/validate/components/index.ts
```diff
@@ -0,0 +1,105 @@
...
+const COMMAND_NAME = 'zwe-validate-components';
+
+/**
+ * Validates that components listed in the zowe.yaml components section
+ * have a valid manifest file (manifest.yaml, manifest.yml, or manifest.json).
+ *
+ * @param quitOnError      - When true, exits the process on failure. When false,
+ *                           logs a warning and returns a non-zero exit code.
+ * @param componentNames   - Optional comma-separated list of component IDs to check.
+ *                           When omitted, only enabled components are checked — disabled
```
Martin-Zeithaml: Em dash used instead of normal one.
```
When omitted, only enabled components are checked — disabled
vs
When omitted, only enabled components are checked - disabled
```

Image

1000TurquoisePogs: Thanks, I absolutely don't want em dashes or any non-ascii content.

bin/commands/validate/components/index.ts
```diff
@@ -0,0 +1,116 @@
...
+ * @param componentNames   - Optional comma-separated list of component IDs to check.
+ *                           When omitted, only enabled components are checked - disabled
+ *                           components that are missing a directory or manifest are not an error.
+ * @returns 0 on success, 1 if any component is invalid.
+ */
+export function execute(quitOnError?: boolean, componentNames?: string): number {
+  common.requireZoweYaml();
+  const ZOWE_CONFIG = config.getZoweConfig();
+
+  if (!ZOWE_CONFIG.runtimeDirectory) {
```
Martin-Zeithaml: ZOWE_CONFIG.zowe.zoweRuntimeDirectory

bin/commands/validate/components/.errors
```diff
@@ -0,0 +1,10 @@
+ZWEL0327E||Component '%s' is missing a valid manifest file (manifest.yaml, manifest.yml, or manifest.json) in %s.
+ZWEL0328E|1|The following component(s) are missing a valid manifest file and cannot be started: %s
```
Martin-Zeithaml: 1) New errors not defined in bin/commands/.error
2) Return value 1 does not corresponds to error message code

bin/commands/validate/components/.errors
```diff
@@ -0,0 +1,10 @@
+ZWEL0327E|327|Component '%s' is missing a valid manifest file (manifest.yaml, manifest.yml, or manifest.json) in %s.
```
Martin-Zeithaml: ZWEL0327E is already used.

MarkAckert: changed to 333

MarkAckert:
There are multiple unused messages. Is there more implementation work planned or should we remove them?

```
Unused message: ZWEL0328E [bin/commands/validate/components/.errors]
Unused message: ZWEL0328W [bin/commands/validate/components/.errors]
Unused message: ZWEL0331E [bin/commands/validate/components/.errors]
Unused message: ZWEL0331W [bin/commands/validate/components/.errors]
```

**Update**: these are used and not captured correctly by automation. Exceptions are added to automation to handle this.

MarkAckert (approved):

PR #4702, opened 2026-03-10
APIML: Remove reference to unused apiml.debug property for single-service mode

taban03:
Remove reference to ZWE_components_apiml_debug in favor of the  ZWE_components_gateway_debug which is the recommended way to enable debug for apiml modulith.

#### PR type
- [ ] Bugfix 
- [ ] Feature 

#### Relevant issues

Fixes 

- 

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

---

---

---

---
_multi-line description_

Doc issue/PR number:


pablocarle (approved):

example-zowe.yaml
```diff
@@ -477,7 +477,6 @@ components:
   # >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
   apiml:
     enabled: true
-    debug: false
```
Martin-Zeithaml, line 480: This property is also defined in the workflows:

https://github.com/zowe/zowe-install-packaging/blob/reboot/fix_apiml_debug/workflows/files/ZWEAMLCF.xml
https://github.com/zowe/zowe-install-packaging/blob/reboot/fix_apiml_debug/workflows/files/ZWECONF.xml

taban03: thanks for catching it

taban03: done!

MarkAckert:
Please check with the workflow team that this change is OK: https://github.com/zowe/zowe-install-packaging/pull/4702/changes#diff-8ab4b1d3d8e75b3427006f2323eeac53054100a29a6af6cbebbce748aafc201cL1196 . I don't think the code block is necessary since the only variable under it was removed, but it's not my expertise.

After they say OK, this PR is ready to merge. thx

1000TurquoisePogs:
Can you confirm if APIML has changed to NOT use this debug property?
Because on v3.4, it DOES. If you remove it, then how else will we debug?

taban03:
> Can you confirm if APIML has changed to NOT use this debug property? Because on v3.4, it DOES. If you remove it, then how else will we debug?

I apologize, indeed I didn't provide the whole context in the PR description.
In the current version `ZWE_components_apiml_debug` property wouldn't work at all, as it is not even propagated and therefore consumed by our `start.sh`.  However, It is possible to enable the API ML debug (modulith) by setting the `ZWE_components_gateway_debug ` or the general debug `ZWE_configs_debug`, and we agreed in the team to still keep it as the recommended way to enable debug in single-service mode.
The reason is that since all the modulith configuration is still living under the gateway section, it would be consistent to apply the same concept for the debug property, so I removed the `ZWE_components_apiml_debug` from the zowe.yaml, and I made sure to document it properly in this PR https://github.com/zowe/docs-site/pull/4947

1000TurquoisePogs:
The code _does_ currently respond to components.apiml.debug, I can reproduce this.

FYI ZWE_configs_debug = ZWE_components_apiml_debug  for the apiml component, so components.apiml.debug does currently enable debugging.
If you do not want the code to respond to it you need to change this line https://github.com/zowe/api-layer/blob/c005c2ee798e4b22e135c8c750df986d060c8ce4/apiml-package/src/main/resources/bin/start.sh#L143

muzikovam (approved):

PR #4709, opened 2026-03-12
V3: getManifestPath: return correct file extension

Martin-Zeithaml:
`getManifestPath` return correct file extension.


MarkAckert (approved):

1000TurquoisePogs (approved):

JoeNemo (approved):
Approved.

MarkAckert:
This change is causing the following error. I didn't look too closely into the call chain yet...for remote integration test suite it's failing on the root manifest.json, but it is also failing in our system tests (logs aren't properly captured). 

```
 ERROR: Validation of /test/dir/components/../manifest.json against schema https://zowe.org/schemas/v2/server-component-manifest found invalid JSON Schema data
      ERROR: Validity Exceptions(s) with object at
      ERROR:   missing required property 'id' at ''
      ERROR:   Validity Exceptions(s) with object at build
      ERROR:     type 'string' not permitted at build.timestamp; expecting type 'integer'
```

Martin-Zeithaml:
`findAllInstalledComponents2` returns `..,apiml-common-lib,common-java-lib,apiml-sample-extension,launcher,app-server`
The level up directory `..` will cause the `manifest.json` is treated as component manifest.

The `getSubdirectories` returns it as `.` and `..` are very often valid and existing directories:
https://github.com/zowe/zowe-install-packaging/blob/8ba48cc552c61a06fdc3b0af095bc5e35b7025b8/bin/libs/fs.ts#L190-L204

MarkAckert:
I updated `getSubdirectories()` so it no longer returns the cwd and parent directories. From what I can gather, it's only used in two files: 

`npm.ts` - `doUpgradeAll` and `doGetPath`
`component.ts` - `findAllInstalledComponents` and `findAllInstalledComponents2`

In each case, the code in no way reads like it expected the parent directory or cwd to be a valid return value. I'll run all the tests, but right now I'm confident this change will be free of side-effects (read: new bugs).

bin/libs/fs.ts
```diff
@@ -192,7 +192,7 @@ export function getSubdirectories(path: string): string[]|undefined {
   let subdirs:string[] = [];
   if (!returnArray[1]) { //no error
     returnArray[0].forEach((dir:string)=> {
-      if (directoryExists(pathoid.join(path,dir))) {
+      if (directoryExists(pathoid.join(path,dir)) && dir !== '.' && dir !== '..') {
```
Martin-Zeithaml: Not that important, but if we switch the conditions, it is little bit better (from CPU POV):
```javascript
if (dir !== '.' && dir !== '..' && directoryExists(pathoid.join(path,dir)))
```
It will not call `directoryExists` for `.` and `..` - string comparison is cheaper and faster than calling function with another function.

Otherwise LGTM, but I can't approve it as the author of this PR.

PR #4713, opened 2026-03-17
Add preflight check for dependencies

1000TurquoisePogs:
This PR changes the validateNodeHome and validateJavaHome functions to behave similarly to other preflight checks, and utilizes them in the new command `zwe validate dependencies` which will check if dependencies are met relative to which components are enabled.

The default check behavior is to error when a version is too low, but only to warn when it is too high.


bin/commands/validate/dependencies/node/index.ts
```diff
@@ -0,0 +1,50 @@
...
+  if (!result) {
+    const baseMsg = `Node.js version validation failed. Ensure Node.js >= ${node.NODE_MIN_VERSION} and <= ${node.NODE_MAX_VERSION} is used wth Zowe.`;
+    let msg: string;
+    if (appServerEnabled) {
+      msg = `${baseMsg} Node.js is required for app-server functionality.`;
+    } else {
+      msg = `${baseMsg} Note: app-server is not enabled, so Node.js may not be required. ` +
+        `If Node.js is not needed, set 'zowe.launchScript.startupChecks.nodeMin' and 'zowe.launchScript.startupChecks.nodeMax' to 'warn' or 'disabled' in your zowe.yaml to bypass this error.`;
+    }
+    common.printFormattedError('ZWELS', COMMAND_NAME, `ZWEL0361E: ${msg}`);
```
1000TurquoisePogs, line 41: need to document error code

bin/commands/validate/dependencies/java/index.ts
```diff
@@ -0,0 +1,59 @@
...
+  if (!result) {
+    const baseMsg = `Java version validation failed. Ensure Java >= ${javaCI.JAVA_MIN_VERSION} and <= ${javaCI.JAVA_MAX_VERSION} is used with Zowe.`;
+    let msg: string = baseMsg;
+    if (enabledJavaComponents.length > 0) {
+      msg = `${baseMsg} Java is required for the following enabled component(s): ${enabledJavaComponents.join(', ')}.`;
+    } else {
+      msg = `${baseMsg} Note: none of the Java-dependent components (${JAVA_DEPENDENT_COMPONENTS.join(', ')}) are currently enabled, so Java may not be required. ` +
+        `If Java is not needed, set 'zowe.launchScript.startupChecks.javaMin' and 'zowe.launchScript.startupChecks.javaMax' to 'warn' or 'disabled' in your zowe.yaml to bypass this error.`;
+    }
+    common.printFormattedError('ZWELS', COMMAND_NAME, `ZWEL0360E: ${msg}`);
```
1000TurquoisePogs, line 45: need to document error code

MarkAckert (approved):

files/defaults.yaml
```diff
@@ -143,6 +143,12 @@ zowe:
       # user: exit
       # # Certificate (zowe.certificate) validation check - defaults to zowe.launchScript.startupChecks.default, or "exit" if this is unset
       # certificate: exit
+      # # Java runtime version checks - max version is set to warn by default to allow for future Java versions to be supported.
+      # javaMin: exit
+      javaMax: warn
+      # # Node runtime version checks - max version is set to warn by default to allow for future Node versions to be supported.
+      # nodeMin: exit
+      nodeMax: warn
```
Martin-Zeithaml: I would switch this with `example-zowe.yaml` (commented out here, set in `example-zowe.yaml`). 

We already had this discussion, if user sets `zowe.launchScript.startupChecks.default=disabled`, the `javaMax` and `nodeMax` will be processed with possible warning. 

I understand the logic here, just print warning for java/node over the supported maximum, but because the default is "hidden" for users, some configurations might be confusing.

PR #4718, opened 2026-03-23
Fix invalid gateway warning when it is disabled

1000TurquoisePogs:
the preflight zosmf check has some over-eager check which says

"Using z/OSMF as 'components.gateway.apiml.security.auth.provider' is not possible: discovery is disabled."

But it says this even when the gateway is disabled. Which is a very likely situation for when the discovery is also disabled.

We shouldn't have checks about what a disabled server would or wouldn't do, so I just added a check to guarantee either 'apiml' or 'gateway' are enabled before this check runs.

### DIFF
--- a/bin/commands/internal/start/prepare/index.ts
+++ b/bin/commands/internal/start/prepare/index.ts
@@ -216,7 +216,7 @@ function globalValidate(enabledComponents:string[]): void {
         privateErrors++;
         common.printFormattedError('ZWELS', "zwe-internal-start-prepare,global_validate", "Zosmf validation failed");
       }
-    } else if (std.getenv('ZWE_components_gateway_apiml_security_auth_provider') == "zosmf") {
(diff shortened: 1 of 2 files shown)


bin/commands/internal/start/prepare/index.ts
```diff
@@ -216,7 +216,7 @@ function globalValidate(enabledComponents:string[]): void {
         privateErrors++;
         common.printFormattedError('ZWELS', "zwe-internal-start-prepare,global_validate", "Zosmf validation failed");
       }
-    } else if (std.getenv('ZWE_components_gateway_apiml_security_auth_provider') == "zosmf") {
+    } else if (enabledComponents.includes('gateway') && std.getenv('ZWE_components_gateway_apiml_security_auth_provider') == "zosmf") {
```
Martin-Zeithaml, line 219: This condition is not the same as changelog text. Or I am confused.

Martin-Zeithaml (approved):

PR #4726, opened 2026-03-31
Fix HashFiles to produce consistent hashes across Java 17 and 21 (#4716)

balhar-jakub:
Hash raw bytes instead of converting to String, eliminating charset dependency introduced by JEP 400 (UTF-8 default in Java 21).

Please check if your PR fulfills the following requirements. This is simply a reminder of what we are going to look for before merging your PR. If you don't know all of this information when you create this PR, don't worry. You can edit this template as you're working on it.

- [ ] Tests for the changes have been added (for bug fixes / features)
- [x] Necessary documentation (if appropriate) have been added / updated
- [x] DCO signoffs have been added to all commits, including this PR

#### PR type
What type of changes does your PR introduce to Zowe? _Put an `x` in the box that applies to this PR. If you're unsure about any of them, don't hesitate to ask._ 
- [x] Bugfix 
- [ ] Feature 
- [ ] Other... Please describe:

#### Relevant issues

Fixes #4716 

#### Changes proposed in this PR
- Update the HashFile class to use UTF-8 and work with bytes where applicable.

#### Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

#### Does this PR add or change a YAML parameter?

- [ ] Yes
- [x] No

#### Does this PR do something the person installing Zowe should know about?

No

#### Is there a related doc issue or Pull Request? 

No

Issue #4716, opened 2026-03-19
`zwe support verify-fingerprints` - JAVA 17 vs 21


Martin-Zeithaml:
There is a small java [code for hashes](https://github.com/zowe/zowe-install-packaging/blob/v3.x/staging/bin/utils/HashFiles.java).

Seems to be working fine for JAVA 17, but not for JAVA 21:

```
/zowe/night/bin: export JAVA_HOME=$JAVA17 && ./zwe support verify-fingerprints
===============================================================================
>> VERIFY ZOWE FILE FINGERPRINTS

2026-03-19 12:57:34 <ZWELS:33882987> MARTIN INFO (validateJavaHome) Java version: 17.0.17 2025-10-21
- Create Zowe directory file list
- Calculate hashes of Zowe files
- Find different files
   * Number of different files: 0
- Find extra files
   * Number of extra files: 0
- Find missing files
   * Number of missing files: 0
-------------------------------------------------------------------------------
>> Zowe file fingerprints verification passed.


/zowe/night/bin: export JAVA_HOME=$JAVA21 && ./zwe support verify-fingerprints
===============================================================================
>> VERIFY ZOWE FILE FINGERPRINTS

2026-03-19 12:57:58 <ZWELS:17105096> MARTIN INFO (validateJavaHome) Java version: 21.0.9 2025-10-21
- Create Zowe directory file list
- Calculate hashes of Zowe files
IOException reading file ./components/app-server/share/zlux-server-framework/node_modules/@rocketsoftware/eureka-js-client/img/eureka-js-client.jpg
EDC5122I Input/output error. (errno2=0xE4000406)
IOException reading file ./components/app-server/share/zlux-server-framework/node_modules/keyring_js/prebuilds/os390-s390x/node.napi.node
EDC5122I Input/output error. (errno2=0xE4000406)
IOException reading file ./components/app-server/share/zlux-server-framework/node_modules/node-forge/flash/swf/SocketPool.swf
EDC5122I Input/output error. (errno2=0xE4000406)
IOException reading file ./components/app-server/share/zlux-server-framework/node_modules/superagent/docs/images/bg.png
EDC5122I Input/output error. (errno2=0xE4000406)
- Find different files
```
Used for test:
* Zowe v3.5.0
* build and hash: v3.x/staging#9799 (2a0601795eaa8e025bb5d400c59b08a7dbbef78e)

Same in 3.4.1.

Same file, different Java and different hash:
```
JAVA 21
[...]

PR #4740, opened 2026-04-14
Fix ExportPrivateKey Java 21 encoding issue (#4730)

### LINKED ISSUE #4730: `ExportPrivateKeyZos` Java 21 encoding issue
```shell
#!/bin/sh

for JAVA in 17 21; do

    echo "< JAVA ${JAVA}"

    "/sys/java64bt/v${JAVA}r0m0/usr/lpp/java/current/bin/java" -cp "/zowe/bin/utils" ExportPrivateKeyZos \
    /zowe/keystore/localhost/localhost.keystore.p12 \
    PKCS12 password localhost password \
    "/zowe/keystore/localhost/localhost${JAVA}.key"

    cat "/zowe/keystore/localhost/localhost${JAVA}.key" | head -c 27
    rm "/zowe/keystore/localhost/localhost${JAVA}.key"

    echo ""
    echo "JAVA ${JAVA} >"
    echo ""

done
```
```
< JAVA 17
-----BEGIN PRIVATE KEY-----
JAVA 17 >

< JAVA 21
����+�&�����.��
JAVA 21 >
```

Seems to be only encoding issue, when using Java 21, after adding `ISO8859-1` tag, the content is readable and correct.

balhar-jakub:
## Summary

Fixes #4730.

`ExportPrivateKeyZos.java` and `ExportPrivateKeyLinux.java` both use `new FileWriter(exportedFile)` without specifying a charset. `FileWriter` relies on `Charset.defaultCharset()`, which changed between Java versions due to [JEP 400](https://openjdk.org/jeps/400) (Java 18+):

| JVM | z/OS default charset | Result |
|---|---|---|
| Java 17 | IBM-1047 (EBCDIC) | USS file is written as EBCDIC; `cat` reads it correctly |
| Java 21 | UTF-8 | USS file is written as UTF-8; `cat` interprets it as EBCDIC → **garbage output** |

This is the same root cause as #4716 / PR #4726 (fixed in `HashFiles.java`).

## Fix

Replace `FileWriter` with `OutputStreamWriter(new FileOutputStream(...), StandardCharsets.ISO_8859_1)` in both files.

- PEM content uses only the ASCII subset of Base64, so ISO-8859-1 bytes are identical to ASCII bytes.
- On z/OS, IBM Java tags the USS output file as **CCSID 819** (ISO-8859-1) when this charset is used, which is exactly what downstream certificate tools expect.
- The output is now deterministic across all Java versions and platforms.

## Test plan

- [x] `ExportPrivateKeyTest.java` — standalone Java test (no JUnit dependency) covering:
  - `FileWriter` with default charset (current buggy code)
  - `OutputStreamWriter` with UTF-16 (simulated wrong-encoding scenario — expected to produce non-PEM output)
  - `OutputStreamWriter` with ISO-8859-1 (the fix — expected to produce valid ASCII PEM)
  - All-ASCII byte check on the fixed output
- [x] `run-export-key-tests.sh` — shell runner that exercises `ExportPrivateKeyLinux` and `ExportPrivateKeyTest` with both Java 17 and Java 21 via sdkman, including `-Dfile.encoding=UTF-16` override to confirm encoding sensitivity of the unfixed code

Run locally:
```sh
cd bin/utils
sh run-export-key-tests.sh
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)


bin/utils/run-export-key-tests.sh
```diff
@@ -0,0 +1,220 @@
+#!/bin/sh
```
Martin-Zeithaml, line 1: Is this AI generated? Was it even executed?

* This test is part of the distribution - it should be in the test directory or deleted when building pax/SMPE/PSWI
* New `sdk` dependency
* Line 23:
```
mktemp: command not found
```

balhar-jakub: It is intentionally draft, I intend to remove this script in the final version. It was used for local testing but not for testing on z/OS. 

That is why I asked you to see whether the issue is resolved on z/OS. The change works on Java versions on local machine.  

I did forget to commit the executable bit, sorry about that.

Martin-Zeithaml:
The files are identical (diff), but shows "garbage":

```
$ /4740/bin/utils: ./run_zos_test.sh
< JAVA 17
����+�&�����.��
JAVA 17 >

< JAVA 21
����+�&�����.��
JAVA 21 >
```

MarkAckert:
> The files are identical (diff), but shows "garbage":
> 
> ```
> $ /4740/bin/utils: ./run_zos_test.sh
> < JAVA 17
> ����+�&�����.��
> JAVA 17 >
> 
> < JAVA 21
> ����+�&�����.��
> JAVA 21 >
> ```

Manually testing ExportPrivateKeyZos shows the private key is coming out as UTF-8, and we want either EBCDIC or ASCII.

balhar-jakub:
BTW as Elena found out, the IBM is changing the behavior in the latest patch version of Java 21, so I will close this one and we will need to first look with API ML on what the impact is and how to make it work with the change before doing anything else. 

Sorry for the used time.

PR #4742, opened 2026-04-15
Remove AFC2 resource - needed only for z/OS < 2.5

Martin-Zeithaml:
The `files/SZWESAMP/ZWEIACF` sample contains a statement for defining resource class - this is needed only for z/OS < 2.5.

## files/SZWESAMP/ZWEIACF
* Remove the resource statement from JCL
* Rename resource type from `ZWE` to `ZOW`

## files/SZWESAMP/ZWEIACFZ
* Rename resource type from `ZWE` to `ZOW`
* These files (`ZWEITSSZ` and `ZWEIRACZ`) are kept for SMP/E compatibility, but no longer needed - used only if unsupported z/OS is used

## workflows/templates/ZWESECUR.vtl
* Resource statement commented out for RACF, ACF2 and TSS

## Another possible solution
Currently the shell code (`zwe init security`) is submitting `ZWESECUR`. JCL based code is running individual ESM and also checking the z/OS version to run resource definition JCL:
https://github.com/zowe/zowe-install-packaging/blob/4c28e78414c5547d2d402b1e884e750b515005d1/bin/commands/init/security/index.ts#L62-L66

We can reuse the JCL code in shell.


files/SZWESAMP/ZWEIACFZ
```diff
@@ -42,7 +42,7 @@ ACF
 * define ZOWE resource type and class mapping
 * skip this section if the ZOWE resource class already exists
 SET CONTROL(GSO)
-INSERT CLASMAP.ZOWE RESOURCE(ZOWE) RSRCTYPE(ZWE)
+INSERT CLASMAP.ZOWE RESOURCE(ZOWE) RSRCTYPE(ZOW)
```
MarkAckert, line 45: Noticed the difference in `RSRCTYPE(ZWE)` and `RSRCTYPE(ZOW)` between `ZWEIACF` and `ZWESECUR`. Is `ZOW` the default shipped after z/OS 2.5?

Martin-Zeithaml: I am just a typewriter for this PR, we have a question from customer claiming, that the default is `ZOW`, but we have `ZWE` in our sample. I will check this with @codezfire or ideally ACF2 team.

MarkAckert (approved):

PR #4743, opened 2026-04-16
Ignore errors for `JCECCARACFKS` or `JCEHYBRIDRACFKS`

Martin-Zeithaml:
For `JCECCARACFKS`
* Output is: `no such provider: IBMJCE4758`

For `JCEHYBRIDRACFKS`:
* Output is `JCEHYBRIDRACFKS not found`

We are currently unable to use `certificate-analyzer` for `JCECCARACFKS` or `JCEHYBRIDRACFKS`.

**Rather than exiting, it would be better to print a warning and continue the process, as the keystore may still be correct.**


MarkAckert (approved):

MarkAckert (approved):
cicd test failures likely unrelated, will debug later

PR #4757, opened 2026-04-24
Fix authloadlib copy to use IEBCOPY instead of TSO REPRO

struga0258:
Description:

Fixes zwe init mvs failing with [ZWEL0162E: Failed to find job result](vscode-file://vscode-app/c:/Users/jstruga/AppData/Local/Programs/Microsoft%20VS%20Code/560a9dba96/resources/app/out/vs/code/electron-browser/workbench/workbench.html) when the authLoadlib dataset already exists, and subsequently causing S806 ABEND (IEW4007I DIRECTORY ENTRY IS NOT VALID FOR A LOAD MODULE) when launching Zowe.

Problem
PR #4695 introduced JCL templates for zwe init mvs that use TSO REPRO to copy load modules (ZWELNCH, ZWESAUX, ZWESISDL, ZWESIS01) from SZWEAUTH to the authLoadlib dataset. This has two issues:

REPRO corrupts load modules — REPRO copies raw byte content but does not preserve the PDS directory entry metadata (entry points, AMODE, RMODE, etc.) required for valid load modules. The copied members are marked "invalid for load", causing S806-04 ABENDs at runtime.

Delete step fails when members don't exist — When --allow-overwrite is used on a dataset that exists but doesn't yet contain the target members, TSO DELETE returns LASTCC=8 for each missing member. IKJEFT01 propagates the highest RC as the step CC, causing the entire job to fail with a non-zero completion code.


bin/commands/init/mvs/index.ts
```diff
@@ -92,7 +92,6 @@ export function execute(allowOverwrite?: boolean) {
         mvsJclAuthLoadlib += template.resolveFile(`${pathTemplatesMvs}/authloadlib.copy.tjcl`, ZOWE_CONFIG);
       } else {
         if (allowOverwrite) {
-          mvsJclAuthLoadlib += template.resolveFile(`${pathTemplatesMvs}/authloadlib.delete.tjcl`, ZOWE_CONFIG);
```
Martin-Zeithaml, line 95: It seems, we don't need `authloadlib.delete.tjcl` at all. `IEBCOPY` will overwrite the load modules.

MarkAckert (approved):

PR #4751, opened 2026-04-21
Print system limitations during preparation of Zowe

taban03:
Add `ulimit -Ha` and `ulimit -a` output information during Zowe preparation, to ease troubleshooting. 
The goal is to print hard system limits and soft system limits and identify possible system limitations that could prevent Zowe to correctly work (i.e. `Maximum number of open file descriptors`).

- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Necessary documentation (if appropriate) have been added / updated
- [ ] DCO signoffs have been added to all commits, including this PR

#### PR type
What type of changes does your PR introduce to Zowe? _Put an `x` in the box that applies to this PR. If you're unsure about any of them, don't hesitate to ask._ 
- [ ] Bugfix 
- [ ] Feature 
- [ ] Other... Please describe:

#### Relevant issues

Fixes 

#### Changes proposed in this PR

- 

#### Does this PR introduce a breaking change?

- [ ] Yes
- [ ] No

#### Does this PR add or change a YAML parameter?

- [ ] Yes
- [ ] No

#### Does this PR do something the person installing Zowe should know about?

---
* Affected function: _general area of interest_ *

---
* Description: _1 line description_ *

---
* Part: _name of customizable file involved_  *

---
_multi-line description_

#### Is there a related doc issue or Pull Request? 

Doc issue/PR number:

PR #4758, opened 2026-04-24
ignore .sh_history if present from fingerprint check

MarkAckert:
verify-fingerprints is failing in our CI environment due to a stray `.sh_history` file in the Zowe directory. We should be able to safely ignore this in all environments.

This `.sh_history` is created by #4751, thought I haven't determined yet if it's the shell path or configmgr path causing it (should be configmgr....). This still needs to be investigated

TODO:
* Why does ulimit create `.sh_history`?


Martin-Zeithaml:
This is very... mainframe.

I did quick test, when running simple script to exec `ulimit` command, no `.sh_history` is created and my `~/.sh_history` is not updated too.

When running Zowe, the `zowe.runtimeDirectory/.sh_history` is created by `ZWESVUSR`.

This PR is probably fine, but we should do a test on READ filesystem only, if an attempt to create `.sh_history` will not cause any troubles.

struga0258 (approved):

PR #4759, opened 2026-04-27
Create new file .zowe-<haInstance>-merged.yaml to the workspace .env

Gautham-coder:
When running in HA mode, `zwe internal start prepare` now writes a per-HA-instance merged YAML config file (`.zowe-<haInstance>-merged.yaml`) to the workspace `.env` directory. This file contains the fully-resolved configuration for that instance with all `haInstances` overrides applied at the root level. The `haInstances` block is preserved in the written file so that consumers such as ZSS can still inspect it (e.g. to count instances for cookie-name disambiguation). The path to this file is exposed via the environment variable `ZWE_HA_INSTANCE_CONFIG` which is propagated to all child processes started for the instance

#### PR type
- [x] Bugfix 
- [x] Feature 

#### Relevant issues
https://github.com/zowe/zss/issues/737

#### Fixes:
ZSS currently uses ".zowe-merged.yaml" file which leaves the haInst as-is and the cfgGet functions don't refer to the HA section as shown here `https://github.com/zowe/zss/blob/940ad1a5f111f4ade0061a9f810c908029e13801/c/zss.c#L1124`. But other components do respect this. 

There are other such HA-overridden temp files there in `.env` and `.json` format, but not for `.yaml`. 

- This PR will create a new `.zowe-<haInstance>-merged.yaml` yaml in the `/zowe/workspace/.env/` which ZSS and other components if needed can use.
- Also, ZSS will use this newly created merged yaml file when running in HA (https://github.com/zowe/zss/pull/826)

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

Testing:
command: ./zwe start --config ../zowe.yaml --ha-instance lpar1
zowe.yaml

```
components:
  zss:
    enabled: true
    port: 56709
    crossMemoryServerName: "ZWESISV3"
    agent:
      jwt:
        fallback: true
      64bit: true

haInstances:
  lpar1:
    hostname: "lpar1.company.com"
    sysname: ${{ zos.resolveSymbol('&SYSNAME') }}
    components:
      zss:
        port: 56729        # override for this instance
        crossMemoryServerName: "ZWESISV3"

```

Before fix:

ZWES1035I ZSS Server settings: Address='0.0.0.0', port='56709', protocol='https'

After fix:

ZWES1035I ZSS Server settings: Address='0.0.0.0', port='56729', protocol='https'

Users who use HA and didn't see ZSS respecting the HA config earlier, and had to change the global values for ZSS or hardcoded values like for example the `crossMemoryServerName` can now see ZSS picking up the HA configuration.

Doc issue/PR number:


bin/libs/configmgr.ts
```diff
@@ -317,27 +317,37 @@ export function cleanupTempDir() {
...
-      zwePrivateWorkspaceEnvDir=`${workspace}/.env`;
+      zwePrivateWorkspaceEnvDir = `${workspace}/.env`;
       std.setenv('ZWE_PRIVATE_WORKSPACE_ENV_DIR', zwePrivateWorkspaceEnvDir);
     }
     fs.mkdirp(workspace, 0o770);
     const mkdirrc = fs.mkdirp(zwePrivateWorkspaceEnvDir, 0o700);
     if (mkdirrc) { return mkdirrc; }
+    return zwePrivateWorkspaceEnvDir;
   } else {
     return dirResult;
```
1000TurquoisePogs: string | number just to return an rc is rather odd.
maybe throw with a try-catch or return undefined.

bin/libs/configmgr.ts
```diff
@@ -587,11 +597,86 @@ function makeHaConfig(haInstance: string): any {
...
+/**
+ * Writes a per-HA-instance merged YAML file to the workspace .env directory.
+ * The result is a fully-resolved config for the given HA instance, with:
+ *   - haInstance component/zowe overrides applied on top of the global config
+ *   - the `haInstances` block is preserved so consumers (e.g. ZSS) can still
+ *     inspect it (e.g. to count instances for cookie-name disambiguation).
+ *
+ * Output: <workspaceDirectory>/.env/.zowe-<haInstance>-merged.yaml
+ */
+export function writeHaMergedConfig(haInstance: string): number {
```
1000TurquoisePogs: this seems to only be called internally (and likely for good reason), so it shouldnt be exported.

bin/libs/configmgr.ts
```diff
@@ -587,11 +597,86 @@ function makeHaConfig(haInstance: string): any {
...
+  const config = getConfig(ZOWE_CONFIG_NAME, ZOWE_CONFIG_PATH, ZOWE_SCHEMA_SET);
+  if (!config.haInstances || !config.haInstances[haInstance]) {
+    return 0; // no HA overrides for this instance, nothing to do
+  }
+
+  // Resolve the output directory (shared with writeMergedConfig via resolveWorkspaceEnvDir)
+  const dirOrErr = resolveWorkspaceEnvDir(config.zowe.workspaceDirectory);
```
1000TurquoisePogs: i dont see why you need to do this again, since if this is only called internally, you can just pass that in as an arg from writeMergedConfig()

bin/libs/configmgr.ts
```diff
@@ -587,11 +597,86 @@ function makeHaConfig(haInstance: string): any {
...
+  if (typeof dirOrErr !== 'string') { return dirOrErr as number; }
+  const zwePrivateWorkspaceEnvDir = dirOrErr;
+
+  // Build the override object from the haInstance entry, skipping instance-only keys
+  // (hostname, sysname) since those don't belong at the root config level.
+  const haInstanceData = config.haInstances[haInstance];
+  const overrideObj: any = {};
+  Object.keys(haInstanceData).forEach((key) => {
```
1000TurquoisePogs, line 673: verify haInstanceData exists, else Object.keys(undefined) could have an issue.

bin/libs/configmgr.ts
```diff
@@ -679,6 +764,15 @@ export function getZoweConfigEnv(haInstance: string): any {
   }
 
   envs['ZWE_haInstance_id'] = haInstance;
-  
+
+  // Propagate the HA instance merged config path to child processes (e.g. ZSS)
+  // so they can locate the fully-resolved, instance-specific YAML without
+  // reconstructing the workspace path themselves.
+  const haInstanceConfig = std.getenv('ZWE_PRIVATE_HA_INSTANCE_CONFIG');
```
1000TurquoisePogs: i really dont like that zss would be reading a *PRIVATE* var. 
if you want others to read this, it should just be a ZWE_ one, not a ZWE_PRIVATE_ one.

1000TurquoisePogs (changes requested):

bin/libs/configmgr.ts
Gautham-coder: There is a check at line 610 doing the verification.

1000TurquoisePogs (approved):

PR #4768, opened 2026-04-28
Switch `ulimit` to non-interactive shell

MarkAckert:
In #4751, we used an interactive shell to invoke `ulimit`, which created a `.sh_history` file in the runtime directory. This file popped up in `zwe support verify-fingerprints`, causing the command to fail. Switching `ulimit` to a non-interactive shell `sh -c` now prevents creation of the `.sh_history` file.

For reviewers: in #4758 , I updated the verify fingerprints to ignore .sh_history. Should we keep or revert that change?

If we keep the change...

Pro:
* Verify-fingerprints is more tolerant / higher signal

Con:
* We don't find out if other commands use an interactive shell. Not necessarily an issue, but our established pattern is non-interactive shells.

### DIFF
--- a/bin/commands/internal/start/prepare/index.ts
+++ b/bin/commands/internal/start/prepare/index.ts
@@ -132,8 +132,8 @@ function globalValidate(enabledComponents:string[]): void {
   const logLevel = std.getenv("ZWE_PRIVATE_LOG_LEVEL_ZWELS");
   const isDebug = (logLevel == "DEBUG" || logLevel == "TRACE");
   if (isDebug) {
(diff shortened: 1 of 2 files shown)


Martin-Zeithaml (approved):
Tested, no `.sh_history` file created.

Martin-Zeithaml:
I would revert #4758:
* `zowe.runtimeDirectory` is expected to be read only
* Any additional file (= verify-fingerprints failure) is a good indicator for Zowe support
* Our CI should not fail with this PR, but it will fail with similar change in the future, which is IMO good

taban03 (approved):

PR #4801, opened 2026-06-05
feat: add allowedDomains configuration

pablocarle:

#### PR type

- [ ] Feature

Add setting zowe.network.allowedDomains to list the hostnames that will be considered valid for the zowe installation. Hosts not listed in this list will not be allowed to be onboarded to the API ML Discovery Service.

Validation in API ML happens for API ML URLs (externalUrl, documentationUrl, swaggerUrl)

- [ ] Yes

- [ ] Yes

---

---

---

---
_multi-line description_

Doc issue/PR number:


bin/libs/configmgr.ts
```diff
@@ -211,6 +211,45 @@ function getDiscoveryServiceUrl(config) {
   return getDiscoveryServiceUrlNonHa(config);
 }
 
+function getDefaultAllowedDomains(config) {
+  const list: string[] = [];
+
+  const haInstanceKeys = Object.keys(config.haInstances);
```
pavel-jares-bcm: I would move this inside the `if (config.haInstances) {` on line 222

bin/libs/configmgr.ts
```diff
@@ -211,6 +211,45 @@ function getDiscoveryServiceUrl(config) {
...
+
+function getAllowedDomains(config) {
+  const defaults = getDefaultAllowedDomains(config);
+
+  let allowedDomains = config.zowe.network.allowedDomains;
+  
+  return Array.from(new Set([
+    ...(defaults || []),
+    ...allowedDomains || []
+  ]));
```
pavel-jares-bcm:
```suggestion
  return Array.from(new Set([
    ...(defaults || []),
    ...allowedDomains || []
  ]))
  .map(host => host.toLowerCase())
  .filter((item, index) => arr.indexOf(item) === index);
```

There could be potentially duplicities, but I'm not sure it worth it...

schemas/zowe-yaml-schema.json
```diff
@@ -1251,6 +1251,20 @@
           "type": "string",
           "description": "This is a variable derived from gateway's ATTLS state to be used by servers proxied through it in eureka or static definitions"
         },
+        "allowedDomains": {
+          "type": "array",
+          "description": "List of domains (supporting wildcards) that are allowed to register with the API ML Discovery Service.",
```
pavel-jares-bcm: Isn't "wildcard" to general. We support only * in the beginning.
```suggestion
          "description": "List of domains (supporting wildcards for subdomains) that are allowed to register with the API ML Discovery Service.",
```

workflows/files/ZWEAMLCF.xml
```diff
@@ -289,6 +289,17 @@ some use cases, like containerization, this port could be different.</descriptio
...
         </string>
     </variable>
+    <variable name="zowe_network_allowedDomains" scope="instance" visibility="public">
+        <label>Zowe allowed domains</label>
+        <abstract>The list of allowed domains in services that register to the API ML Discovery Service</abstract>
+        <description>Defines a list of domains that can be used by the Zowe instance.</description>
+        <category>network</category>
+        <string valueMustBeChoice="false" multiLine="true">
+            <!-- Put validation here -->
+            <!-- Specify choices here -->
```
pavel-jares-bcm, line 299:
```suggestion
```

Or is it on purpose?

workflows/files/ZWECONF.properties
```diff
@@ -182,6 +182,14 @@ zowe_launchScript_logLevel=info
 # Choices: warn,exit
 zowe_launchScript_onComponentConfigureFail=warn
 
+# zowe_network_allowedDomains
+# Label: Allowed domains for services that register to the API ML Discovery Service
+# Abstract: Update based on domains used by the Zowe instance
+# Category: network
+# Description:
+#  Allowed domains for services that register to the API ML Discovery Service.
+zowe_network_allowedDomains=www.example.com
```
pavel-jares-bcm, line 191: Do we trust customer will delete it?

example-zowe.yaml
```diff
@@ -222,6 +222,10 @@ zowe:
   # This section can be overridden per-component by placing it
   # Under a "zowe" subsection of a component, as in "components.zss.zowe.network"
   network:
+    # Allowed domains is a list of domains (supporting wildcards) that can be used in services that register to the API ML Discovery Service.
+    # Defaults to the external domains + lpar hostnames if defined in haInstances
+    allowedDomains:
+      - www.example.com
```
pavel-jares-bcm, line 228:
```suggestion
    #allowedDomains:
    #  - www.example.com
```
I would comment it out.

workflows/files/ZWECONF.properties
pablocarle: it's the same default we have in externalDomains anyway

workflows/files/ZWEAMLCF.xml
pablocarle: It's copied from another block, I can delete them I don't know what they mean

bin/libs/configmgr.ts
pablocarle: the code already removes duplicates

example-zowe.yaml
pablocarle: Most installations will need to set this up.

schemas/zowe-yaml-schema.json
```diff
@@ -1251,6 +1251,20 @@
...
+            "type": "string"
+          }
+        },
+        "allowedDomainsDefaults": {
+          "type": "array",
+          "description": "List of domains (supporting wildcards) that are allowed to register with the API ML Discovery Service.",
+          "items": {
+            "type": "string"
+          }
+        },
```
Martin-Zeithaml, line 1270:
```
$ /zowe/4801/bin: cat zowe.yaml && ./zwe config validate -c zowe.yaml
zowe:
  network:
    allowedDomains:
      - asdf
      - asdf*
      - '*'
      - ~
      - null
      -
      - ''
      - example.com
      - example.com
      - example.com
      - example.com
2026-06-09 06:58:34 <ZWELS:33882649> MARTIN INFO (zwe-internal-start-prepare) Zowe version: v3.5.0
2026-06-09 06:58:34 <ZWELS:33882649> MARTIN INFO (zwe-internal-start-prepare) build and hash: PR-4801#10489 (6949e50f2c5be41e047ff6932fee2ec11d42943a)
Validating Zowe core configuration for file(s)=zowe.yaml
Zowe core configuration is valid
```

If you check, how property `sysMessages` is defined, we are limiting an item to be unique and a string with at least one character (null or empty string not allowed) - this simplifies the code when the `sysMessages` item is processed.

I think, this could be used here too.

pablocarle: Thanks, I'll try this out

example-zowe.yaml
1000TurquoisePogs: If most installations will need to set this up, then you're saying this PR is a breaking change that requires a HOLDDATA?

files/defaults.yaml
```diff
@@ -91,6 +91,10 @@ zowe:
   network:
     # This is a variable derived from gateway's ATTLS state to be used by servers proxied through it in eureka or static definitions.
     proxyType: "${{ components.gateway?.zowe?.network?.server?.tls?.attls == true ? 'http' : zowe.network?.server?.tls?.attls == true ? 'http' : 'https' }}"
+    allowedDomainsDefaults:
+      - "www.ibm.com"
+      - "zowe.github.io"
+      - "www.zowe.org"
```
1000TurquoisePogs: What is the purpose of this list?
The schema says

> List of domains (supporting wildcards) that are allowed to register with the API ML Discovery Service.

But, why would I want to onboard something from zowe.org? There's no mainframe at zowe.org right?

schemas/zowe-yaml-schema.json
```diff
@@ -1251,6 +1251,30 @@
...
+          "uniqueItems": true,
+          "items": {
+            "type": "string",
+            "minLength": 1,
+            "not": {
+              "type": "null"
+            }
+          }
+        },
+        "allowedDomainsDefaults": {
```
1000TurquoisePogs: why is there a "defaults" of a property, when configmgr handles defaults through defaults.yaml file?
Is there a limitation of configmgr that has resulted in this decision?

1000TurquoisePogs (dismissed):
Please add a PR description that tells us what this is.
I am guessing it affects me but I cannot properly test & review it without a description.

schemas/zowe-yaml-schema.json
pablocarle: We wanted to have this list somewhere where it can be visible, the only property meant to be updated from zowe.yaml is `allowedDomains`. I can move it to code though, or any other proposal.

files/defaults.yaml
pablocarle: There are a few URLs (in z/OSMF static definition and in the core services) that use URLs from these domains for things like `documentationUrl`.
The reason we are adding this allow list is to control which domains are accessible from the API ML.

pablocarle: The restriction on API ML side is defined in https://github.com/zowe/api-layer/pull/4656

schemas/zowe-yaml-schema.json
Martin-Zeithaml: `"uniqueItems": true` can cause the problems, when defining an item, because the resulting set must be unique: user defined values AND default values from `defaults.yaml`.

* `"uniqueItems": true` is fine `allowedDomains`
* `"uniqueItems": true` is problematic `allowedDomainsDefaults`

I "vote" for removing `allowedDomainsDefaults`.

pablocarle: I can move those defaults to API ML, they are all based on API ML current configuration after all

files/defaults.yaml
pablocarle: btw, the default is blocking but we are providing an environment variable `ZWE_ONLY_WARN_ON_URL_NOT_ALLOWED` to leave it as a warning only

example-zowe.yaml
pablocarle: Yes, it's planned as a breaking change cc. @balhar-jakub

files/defaults.yaml
1000TurquoisePogs: Are you saying these URLs don't just control the discovery registration, but also control the API catalog? I thought `documentationUrl` was for API catalog, but not used by discovery service?

example-zowe.yaml
1000TurquoisePogs: Breaking changes are not supposed to occur until v4.0.

1000TurquoisePogs: A way to introduce this change without breaking, as has been done for several changes within the v3 line, is to have your behavior of "ZWE_ONLY_WARN_ON_URL_NOT_ALLOWED" just be the default in v3, and change it in v4.

schemas/zowe-yaml-schema.json
pablocarle: moved to API ML

files/defaults.yaml
pablocarle: They control if a service can register to discovery service evaluating the URLs services present in the eureka metadata for apiml keys::
"swaggerUrl",
"graphqlUrl",
"documentationUrl",
"externalUrl",
"corsAllowedOrigins"

plus the eureka default fields:
Home Page URL, HealthCheck URL, Status Page URL and Secure Health Check URL

example-zowe.yaml
pablocarle: I have no problem with that

1000TurquoisePogs:
I no longer have any technical objections to this PR - reviewed it again and it looks like all issues are resolved.

Martin-Zeithaml:
Very edge QA case, but anything concatenated by comma is comma and space sensitive:

```yaml
zowe:
  network:
    allowedDomains:
      - example.com
      - ", "
      - github.com
      - ','
```
```
Caused by: java.lang.ClassNotFoundException: ,github.com,,
Error: Could not find or load main class ,github.com,,    
```

Not sure, if we can prevent this, but we might throw error sooner than java.

pablocarle:
> Very edge QA case, but anything concatenated by comma is comma and space sensitive:
> 
> ```yaml
> zowe:
>   network:
>     allowedDomains:
>       - example.com
>       - ", "
>       - github.com
>       - ','
> ```
> 
> ```
> Caused by: java.lang.ClassNotFoundException: ,github.com,,
> Error: Could not find or load main class ,github.com,,    
> ```
> 
> Not sure, if we can prevent this, but we might throw error sooner than java.

Do you think worth to add a regex to validate the entries? Is this something supported in the schema validations?

Martin-Zeithaml:
Yes, regex is possible to use, this is already used for data set validation:
https://github.com/zowe/zowe-install-packaging/blob/53db99bfaebaee483329ea68f63fbeb1a1eee9d0/schemas/server-common.json#L19-L25

Or we can "clean" the allowedDomains when building `ZWE_ALLOWED_DOMAINS` and reject or trim string with spaces and reject strings with leading or trailing commas.

MarkAckert:
With the latest change, I saw this output from `internal start prepare`. 

```
Debug: Invalid domain: example.com
Debug: Invalid domain: www.example.com
```

Think it was a minor bug in the domain verification, so changed that (please review) and updated test snapshots.

pablocarle:
> With the latest change, I saw this output from `internal start prepare`.
> 
> ```
> Debug: Invalid domain: example.com
> Debug: Invalid domain: www.example.com
> ```
> 
> Think it was a minor bug in the domain verification, so changed that (please review) and updated test snapshots.

Thank you!

pablocarle:
Verified the PAX in internal environment successfully

MarkAckert (approved):

Martin-Zeithaml (approved):

PR #4805, opened 2026-06-09
Shell commands sanitizing

skurnevich:
heredoc and eval shell commands sanitizing


bin/libs/component.sh
```diff
@@ -295,13 +295,23 @@ process_component_apiml_static_definitions() {
       sanitized_def_name=$(echo "${one_def_trimmed}" | sed 's/[^a-zA-Z0-9]/_/g')
       # FIXME: we may change the static definitions files to real template in the future.
       #        currently we support to use environment variables in the static definition template
-      parsed_def=$( ( echo "cat <<EOF" ; cat "${one_def}" ; echo ; echo EOF ) | sh 2>&1)
+      # Use envsubst to expand only $VAR / ${VAR} references; no subshell is spawned.
```
Martin-Zeithaml, line 298: I am not sure about this - at this stage of processing we don't have access to original content.
If `VAR=$(ls)` was originally set, we can see here `VAR=file1 file2...`.

skurnevich: This fix should cover a malicious manifest file values, if i get the comment correctly, for having VAR=$(ls) someone must have control over the execution environment, which is a different issue

Martin-Zeithaml (approved):

PR #4807, opened 2026-06-12
`prepare_log_file`: update log file creation to use umask for permissions

Martin-Zeithaml:
1) User can read and write
2) Group can read - **do we want to allow write?**
3) Others no read | write

`chmod` removed as the file is created with `umask & touch`.

Used for `zwe` command, when `--log-dir` is specified.

```
./zwe some_command --config zowe.yaml --log-dir /myLogs

ls -al /myLogs

-rw-rw-rw-   1 MARTIN   ZOWEGRP      157 Jun 12 11:05 zwe-some-command-20260612T110543.log  <- old
-rw-r-----   1 MARTIN   ZOWEGRP      157 Jun 12 11:24 zwe-some-command-20260612T112442.log  <- new
```

Note: `prepareLogFile` (typescript version) seems to be not used and there is a [bug](https://github.com/zowe/zowe-install-packaging/issues/4811).


MarkAckert (approved):
Agree `0640` makes sense for this change. Multiple users running `zwe` commands each get their own log file by design; there's no benefit in sharing write access to a `zwe` log file.

PR #4826, opened 2026-06-26
Shell commands sanitizing v2

skurnevich:
Same as https://github.com/zowe/zowe-install-packaging/pull/4805/ for v2


MarkAckert (approved):
Same changes as #4805, and Martin approved those.

PR #4845, opened 2026-07-29
pkcs12 password hidden from the log

abastyr:
- Hidden the `pkcs12` password from the logs (replaced by `...`)
- Made logs invisible to the *Other* users.
- Fixed calling the `keytool` with a single parameter


bin/libs/certificate.sh
```diff
@@ -66,10 +66,29 @@ JAVA_KEYTOOL_ENCODING=" -J-Dfile.encoding=COMPAT "
...
+  printitem=1
+
+  for item in $args; do
+    if [ $printitem -eq 1 ]; then
+      maskedargs="${maskedargs}$item "
+    else
+      maskedargs="${maskedargs}... "
+      printitem=1
+    fi
+    if [ "${item}" = "-storepass" ]; then
```
MarkAckert: should we also mask `-keypass`, `-destkeypass`, and `-srcstorepass`? We call these with `pkeytool` during some certificate setup scenarios.

abastyr: Sounds like a plan

abastyr: I've updated the script with a case for multiple possible parameters

MarkAckert: I added two more, tested the changes and they LGTM!

MarkAckert (approved):

bin/libs/logging.sh
```diff
@@ -25,7 +25,8 @@ prepare_log_file() {
   ZWE_PRIVATE_LOG_FILE="${log_dir}/${log_file_prefix}-$(date +%Y%m%dT%H%M%S).log"
   if [ ! -f "${ZWE_PRIVATE_LOG_FILE}" ]; then
     # create and echo message if log file doesn't exist
-    mkdir -p "${log_dir}"
+    umask 0007
```
1000TurquoisePogs, line 28: why did we lose -p on the mkdir here? i think its still needed, else mkdir could fail.

abastyr: If `mkdir` fails, it means the directory probably exists, hence `chmod` is executed to assure the permissions are set correctly.

`mkdir -p` would end with rc 0 and `chmod` wouldn't be executed.

skurnevich: I believe the -p option was intended to catch a case when one or more parent dirs do not exist, it is a valid case as log_dir could be supplied as a parameter (-l / --log-dir) to any zwe command with any path, so this would be a regression.
Couple more notes here. 
`umask 0007` is set and never restored, so potentially may change the behavior of further dirs and files creation in this shell session. 
`chmod 770 -R "${log_dir}"` -R option seems to be misplaced so chmod may treat it as a directory name

skurnevich (changes requested):

skurnevich:
Closing as this one is now covered in https://github.com/zowe/zowe-install-packaging/pull/4807 and https://github.com/zowe/zowe-install-packaging/pull/4883 PRs

PR #4860, opened 2026-08-14
Better vars assigning, removed duplicated code

skurnevich:
#### PR type
- [ ] Bugfix 
- [ ] Feature 
- [x] Other... Please describe:

Better var assignment, the value is substituted by eval's own parameter expansion instead of being spliced into eval's source text, shell metacharacters in the value (quotes, $(...), backticks, ;) are never re-parsed as code.

Also removed duplicated piece of code 
- \# read keyring-specific z/OSMF info - lines 151-157

as it is unconditionally executed a few lines below 
- \# read z/OSMF info - lines 167 - 172


1000TurquoisePogs (approved):
This makes sense

MarkAckert (approved):

PR #4871, opened 2026-08-27
new generate variable for zowe conf

### LINKED ISSUE #4653: PSWI Create Configuration step is failing

pavelzlatnik:

#### PR type
- [X] Bugfix 
- [ ] Feature 

#### Relevant issues

Fixes #4653 

Simplified trigger for generate, just new variable which can control injection via velocity (to allow not to have it in workflow).
- 

- [ ] Yes
- [X] No

- [X] Yes
- [ ] No

---

---

---

---
_multi-line description_

Doc issue/PR number: 

#### Other information

### DIFF
--- a/example-zowe.yaml
+++ b/example-zowe.yaml
@@ -40,7 +40,7 @@ zowe:
       # When enabled, "zwe install & init" will use job submission for system changes
       # instead of shell commands. A `--dry-run` option is also enabled
       # which allows you to see jobs instead of submitting.
-      enable: true      
+      enable: true
       # Regenerate JCL Samples for Zowe Setup
       # is not skipped (command zwe init generate is executed).
       generate: true
--- a/example-zowe.yaml
+++ b/example-zowe.yaml
@@ -40,7 +40,10 @@ zowe:
       # When enabled, "zwe install & init" will use job submission for system changes
       # instead of shell commands. A `--dry-run` option is also enabled
       # which allows you to see jobs instead of submitting.
-      enable: true
+      enable: true      
+      # Regenerate JCL Samples for Zowe Setup
+      # is not skipped (command zwe init generate is executed).
+      generate: true
       # Edit to set a header for job cards. Some systems need this, but not all.
       # See `zwe init generate --help` for examples.
       header: ""
--- a/schemas/zowe-yaml-schema.json
+++ b/schemas/zowe-yaml-schema.json
@@ -570,6 +570,10 @@
                   "type": "boolean",
                   "description": "Controls whether zwe install and zwe init commands and subcommands will utilize jobs for executing system changes instead of shell commands."                
                 },
+                "generate": {
+                  "type": "boolean",
+                  "description": "Controls whether zwe regenerate jcls."                
+                },
                 "header": {
                   "type": "string",
                   "not": {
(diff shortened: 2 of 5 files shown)


MarkAckert:
I'm not sure if this fixes the linked issue as described, and may introduce another issue? 

**The new problem:**
If someone has `zowe_setup_jcl_enable=true` and `zowe_setup_jcl_generate=false`, then they would also have a bricked workflow. We'd try to submit JCL which was never generated. The field `zowe_setup_jcl_generate=false` is only valuable on repeated runs of the workflow (e.g. when going back in the WF and updating / changing variables in the `zowe.yaml`)

**The old problem:**
Users had `zowe_setup_jcl_enable=true` but `zwe init generate` couldn't find the template JCL on the system, so no JCL was generated. This will cause an error in Step #2, or if you removed the `zwe init generate` block of code, it will cause an error later in Step #3. In step #3, when any `zwe init` command is run and `zowe_setup_jcl_enable=true`, the `zwe init` command will look for rendered JCL or run `init generate` under the hood if it's missing.

`zowe_setup_dataset_prefix` is used as the common origin for the template JCL and the rendered JCL. This being set to something other than the prefix to the template JCL caused the original problem in the linked description. It wasn't clear that this was the root cause, and it introduces a requirement which wasn't clear: the prefix for the PSWI had to be `<ZOWE.PSWI.PREFIX>.T` (template JCL were only in target zone iirc). There were a few options to fix this as I understood them:

* Fix documentation and clarify what `zowe.setup.dataset.prefix` must be when using JCL.
* Separate `init generate`'s dependency on `zowe.setup.dataset.prefix` for sample JCLs. Either: 
   * Use a new, explicit variable which points to the template JCL directly. Render as per-usual.
   * If `zowe.setup.dataset.prefix` is not valid, read the JCL templates from USS instead and drop the rendered output as per-usual.

pavelzlatnik:
> * Fix documentation and clarify what `zowe.setup.dataset.prefix` must be when using JCL.
> * Separate `init generate`'s dependency on `zowe.setup.dataset.prefix` for sample JCLs. Either:
>   
>   * Use a new, explicit variable which points to the template JCL directly. Render as per-usual.
>   * If `zowe.setup.dataset.prefix` is not valid, read the JCL templates from USS instead and drop the rendered output as per-usual.

Hi Mark,

That is a fair point, and I agree with your concern. While it's true that misconfiguring any variable can cause a step to fail, adding zowe_setup_jcl_generate does create a direct path to a bricked workflow if a user sets it to false while zowe_setup_jcl_enable is true. We'd just be shifting the confusion rather than completely solving it.

My original intent with this script update was specifically to prevent failures when a user makes changes in the workflow and tries to re-run the step.

To solve this purely from a workflow perspective without risking misconfiguration, I can close this PR (dropping the new control variable idea). Since the configuration workflow isn't fully automatic anyway, I will instead add a clear Note to the step's instructions.

The note would look like this:

If you are modifying the workflow and re-running this step, you might encounter an error (ZWEL0143E Cannot find a data set member…) where it fails due to no MVS datasets being created by the Create Configuration step. To resolve this, manually remove the zwe init generate command from the step: click Edit JCL, delete the zwe init generate line, and submit the step.

We can then have a separate discussion about your proposed solutions for the root cause (which fall outside the scope of a simple workflow modification). Just to recap your ideas for fixing the underlying zwe init behavior:

Fix documentation and clarify what zowe.setup.dataset.prefix must be when using JCL.

Separate init generate's dependency on zowe.setup.dataset.prefix for sample JCLs. Either:

Use a new, explicit variable which points to the template JCL directly, rendering as per-usual.

If zowe.setup.dataset.prefix is not valid, read the JCL templates from USS instead and drop the rendered output as per-usual.

pavelzlatnik:
for zowe init generate command issue, at least note for step instruction created (No new control variable). 
https://github.com/zowe/zowe-install-packaging/pull/4874

PR #4877, opened 2026-08-31
use allowed domains = external domains

1000TurquoisePogs:
To reduce the likelihood that the 'allowed domains' field is a disruptive change users, it would make sense to have it templated to  external domains, so that the two match by default and only one needs to be set.

### DIFF
--- a/example-zowe.yaml
+++ b/example-zowe.yaml
@@ -224,8 +224,7 @@ zowe:
   network:
     # Allowed domains is a list of domains (supporting wildcards) that can be used in services that register to the API ML Discovery Service.
     # Defaults to the external domains + lpar hostnames if defined in haInstances
-    allowedDomains:
-      - www.example.com
+    allowedDomains: "${{ zowe.externalDomains }}"
     server:
       tls:
         attls: false


example-zowe.yaml
```diff
@@ -224,8 +224,7 @@ zowe:
   network:
     # Allowed domains is a list of domains (supporting wildcards) that can be used in services that register to the API ML Discovery Service.
     # Defaults to the external domains + lpar hostnames if defined in haInstances
-    allowedDomains:
-      - www.example.com
+    allowedDomains: "${{ zowe.externalDomains }}"
```
Martin-Zeithaml, line 227: This needs to be tested - how is the array `zowe.externalDomains` expanded?

```yaml
array0:
  - valid_0
  - valid_1

# We have to avoid this
array1: - not_valid_0
  - not_valid_1
  
array2: [ valid_0, valid_1 ]

array3: [ 
  valid_0, 
  valid_1
]
```

Martin-Zeithaml:
```yaml
zowe:
  externalDomains:
    - one.example.com
    - two.example.com
  network:
    allowedDomains: "${{ zowe.externalDomains }}"
```

```yaml
zowe:
  externalDomains: [ "one.example.com", "two.example.com" ]
  network:
    allowedDomains: "${{ zowe.externalDomains }}"
```

Both examples correctly expanded to:

```yaml
zowe:
  network:
     proxyType: "https"
     allowedDomains:
     - "one.example.com"
     - "two.example.com"
```

pablocarle:
The code in https://github.com/zowe/zowe-install-packaging/blob/v3.x/staging/bin/libs/configmgr.ts#L214 already takes any externalDomains defined, isn't this change redundant with the current logic?

MarkAckert:
Pablo makes a good point.

Also, with `allowedDomains: "${{ zowe.externalDomains }}"`, we make adding more allowedDomains a little troublesome. You'd either add the new allowedDomains to `externalDomains` or break the template altogether and revert to individual fields. 

Maybe we instead clarify how `allowedDomains` works in the zowe.yaml description?
 
It currently states: `# Defaults to the external domains + lpar hostnames if defined in haInstances`
Alternative version: `# Entries here are in addition to hostnames defined in zowe.externalDomains and zowe.haInstances`

PR #4893, opened 2026-09-15
Bugfix: zwe hangs when a child process writes more than 64 KiB (zowe-common-c#672)

### LINKED ISSUE #672: Shell does not end
Directory with large number of files

Following command
```shell
ls -TREal /zowe/runtime
```
Returns the 20 000+ lines of Zowe runtime directory structure.

Same command executed via `configmgr`:
```javascript
// testShell.js
import * as shell from './libs/shell';
const result = shell.execOutSync('sh', '-c', `ls -TREal "/zowe/runtime"`);
console.log(result.out);
```
Execution is stuck, direct shell command is done with couple seconds, JavaScript did not finish within minutes.

```shell
./configmgr -script ./testShell.js
```

Smaller amount of files is processed fine:
```javascript
// testShell.js
import * as shell from './libs/shell';
const result = shell.execOutSync('sh', '-c', `ls -TREal "/zowe/runtime/bin/utils"`);
console.log(result.out);
```

Related to zowe/zowe-install-packaging#4843

JoeNemo:
Fixes zowe/zowe-common-c#672, which is filed on zowe-common-c but is a `zwe` bug.

## What happens

`execOutSync()` in `bin/libs/shell.ts` runs the child with `block: true`, so `os.exec` sits in `waitpid()` while nobody reads the pipe. A pipe holds about 64 KiB. A child that writes more blocks in `write()`, the parent blocks in `waitpid()`, and `zwe` never returns. The report's `ls -TREal` of a 20 000-line runtime directory is well over the limit. The threshold is exactly the pipe buffer: 60 000 bytes of child output completes, 70 000 hangs.

Seen live on the Marist CI system on 5 September: three `configmgr -script .../internal/start/component/cli.js` processes from the staging instance, alive for over nine hours with no CPU.

A second latent fault in the same helper: `readStreamFully()` stopped at the first read shorter than its 4 KiB buffer. That only worked because the child was already dead and the pipe fully buffered; with a live writer short reads are normal and output would be truncated.

## The change

- `execOutSync()`: run the child with `block: false`, drain stdout to end of file, then wait for it.
- `execOutErrSync()`: two pipes cannot be drained one after the other without the same deadlock in the other direction, so stderr goes to a temporary file (`std.tmpfile()`, no buffer limit) and is read back after the child exits. Stdout is drained as above.
- `readStreamFully()` reads to end of file.
- `build/zwe/types/@qjstypes/cm_std.d.ts` gains `tmpfile`, `SEEK_*` and the `File` methods `seek`, `fileno`, `readAsString`, which the engine has always provided but the type file did not declare. Without them the TypeScript build rejects the fix.

## Testing

On Linux with the configmgr binary from zowe-common-c `build/build_cmgr_clang.sh linux`, two scripts mirror the helper before and after (`dev-test-deploy/issue672/hang-repro.js`, `hang-fixed.js` in the maintainer's working tree):

| child output | before | after |
|---|---|---|
| 1 000 bytes | ok | ok |
| 60 000 | ok | ok |
| 70 000 | hangs, killed at 10 s | ok |
| 200 000 | hangs | ok |
| 5 000 000 | not tried | ok, 98 ms |

`npm ci && npm run prod` in `build/zwe` passes with the change. Not yet run on z/OS: Marist is down. Nothing in the change is platform-specific; `os.exec`, `os.pipe`, `std.tmpfile` behave the same on both.


Martin-Zeithaml:
## Test

```javascript
// test4893.js
import * as shell from './libs/shell';
const result = shell.execOutSync('sh', '-c', `ls -TREal "/zowe/pull_request/4893"`);
console.log(result.out);
console.log(result.out.length);
console.log(result.rc);
```

## Execution

```shell
./utils/configmgr -script ./test4893.js
```

## Result
* It is printed 64kB + 1B
* Length 1 445 507
* RC = 0

## Next step
* This is fixing the main issue, which was a frozen process
* It shows new issue with `console.log`
  * I will prepare a fix
  * Already done some research
  * It looks like a `quickjs-portable` need a fix

